PR #7, opened 2019-05-15
non-RFC: add info about github issues

dhh1128:


kdenhartog:
Looks good and accurately reflects discussions from todays call.

PR #20, opened 2019-05-24
message types

TelegramSam:


concepts/0020-message-types/README.md
```diff
@@ -0,0 +1,223 @@
...
+{
+    `@type`:<message type string>,
+    // other attributes
+}
+```
+
+Message types are URIs that resolve to developer documentation for the message type. The end portion of the URI MUST follow the following format:
+
+```
+/<message_family>/<major_family_version>.<minor_family_version>/<message_type>
```
dhh1128: I would be happier describing the first section as protocol, not message family. Also, note that this content is redundant with a discussion about versioning in https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/uris.md and https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/semver.md. We should probably link to those docs instead of repeating the content here.

TelegramSam: Did I sufficiently fix this? I left the example in play. Also, message families -> protocols.

PR #33, opened 2019-05-28
accept RFC: l10n

dhh1128:


swcurran:
I would say proposed.  No one has really done anything with this, and I don't see how the language libraries will be published/managed in an effective way.  I'd like to see more on that before it is accepted.

PR #38, opened 2019-05-29
touch up title ordinal, flesh out keys and authn/encr

sklump:
Align title ordinal to 0005. Explicitly specify authn vs encryption keys.


sklump:
FFS DCO.

PR #42, opened 2019-05-31
propose RFC: Lox

mikelodder7:


kdenhartog:
Do you plan to define more APIs and Architectures in this document? From the looks of it this will turn more into a Feature doc rather than a concept document. Would you mind moving it to that folder?

mikelodder7:
> Do you plan to define more APIs and Architectures in this document? From the looks of it this will turn more into a Feature doc rather than a concept document. Would you mind moving it to that folder?

I moved it to the `features` folder and updated the text to be clearer about the Architecture and APIs.

kdenhartog:
Awesome thanks looks good to me.

PR #46, opened 2019-05-31
accept RFC: mediators and relays

dhh1128:


concepts/0046-mediators-and-relays/README.md
```diff
@@ -0,0 +1,242 @@
+# 0036: Mediators and Relays
```
troyronda: The RFC number here does not match the number in the folder name.

dhh1128: Thank you for the catch, @troyronda ! I will update.

troyronda: Cool. I had already created https://github.com/hyperledger/aries-rfcs/pull/57 if you want to use it.

PR #65, opened 2019-06-05
sha256 and description

sklump:
Make pics and text internally consistent re: sha256 and description


sklump:
Amended to propagate current 0017-attachments keys to 0036-issue-credential text.

PR #67, opened 2019-06-05
Fix links still pointing to the indy hipe repo

llorllale:
Updated most of the references to HIPEs to point to their respective RFCs in this repo.

There's still no Aries equivalent of [general error reporting mechanism](https://github.com/hyperledger/indy-hipe/blob/6a5e4fe2d7e14953cd8e3aed07d886176332e696/text/error-handling/README.md) so I've left that untouched.


swcurran:
It's in aries-rfc already - https://github.com/hyperledger/aries-rfcs/tree/master/features/0035-report-problem

PR #70, opened 2019-06-07
Add `forwardMsgReq` to Service Decorator

llorllale:
**Motivation**
* Allow more sophisticated gatekeeping by mediators/routers as described in [HIPE 0022 - Cross Domain Messaging](https://github.com/hyperledger/indy-hipe/tree/master/text/0022-cross-domain-messaging).
* Provide a way to signal to the Sender that a message for the mediator/router is required for gatekeeping

**Changes**
* Adds a new attribute to the `~service` decorator
* Tweaked the language around the definition of `~service` to say that it *extends* the `service` attribute in the DID Spec, as opposed to being entirely defined by it. Reasons:
   * DID*Comms* ("routers", "routing keys") should not leak into the DID Spec
   * There already exist attributes not in the DID Spec (`routingKeys`, `recipientKeys`)

**Note:** an accompanying change is required on HIPE 0022 once it's ported over to the Aries RFCs.

### DIFF
--- a/features/0056-service-decorator/README.md
+++ b/features/0056-service-decorator/README.md
@@ -13,32 +13,38 @@ The `~service` decorator describes a DID service endpoint inline to a message.
 
 ## Motivation
 
-This allows messages to self contain endpoint and routing information normally in a DID Document. This comes in handy when DIDs or DID Documents have not been exchanged.
+This allows messages to self contain endpoint and routing information normally found in a DID Document. This comes in handy when DIDs or DID Documents have not been exchanged.
 
 Examples include the Connect Protocol and Challenge Protocols.
 
-The `~service` decorator on a message contains the service definition that you might expect to find in a DID Document. These values function the same way.
+The `~service` decorator on a message extends the service definition that you might expect to find in a DID Document.
 
 ## Tutorial
 
-Usage looks like this, with the contents defined the [Service Endpoint section of the DID Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints):
+Usage looks like this, extending the contents defined in the [Service Endpoint section of the DID Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints):
 
-```json=
+```json
 {
     "@type": "somemessagetype",
     "~service": {
         "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
-        "routingKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"]
-        "serviceEndpoint": "https://example.com/endpoint"
+        "routingKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
+        "serviceEndpoint": "https://example.com/endpoint",
+        "forwardMsgReq": "did:example:aYgjkk34JDJKLJh;service=schema/exampleschema/1.0/exampletype"   # optional
     }
 }
 ```
 
+`forwardMsgReq` is optional and its value is a URI for a schema:
 
+* The inclusion of `forwardMsgReq` signals to the Sender that `forwardMsg` is required on the *Forward* message in order for the message to reach the Recipient.
+* There are no restrictions on the language used to describe the schema so long as both parties (sender, router) understand it.
+
+The combination of `forwardMsgReq` and `forwardMsg` enable more sofisticated gatekeeping as described in [Indy HIPE 0022 - Cross Domain Messaging](https://github.com/hyperledger/indy-hipe/tree/master/text/0022-cross-domain-messaging).
 
 ## Reference
 
-The contents of the `~service` decorator are defined by the  [Service Endpoint section of the DID Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints).
+The base contents of the `~service` decorator are defined by the  [Service Endpoint section of the DID Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints).
 
 The decorator should not be used when the message recipient already has a service endpoint. 
 
--- a/features/0056-service-decorator/README.md
+++ b/features/0056-service-decorator/README.md
@@ -40,7 +40,7 @@ Usage looks like this, extending the contents defined in the [Service Endpoint s
 * The inclusion of `forwardMsgReq` signals to the Sender that `forwardMsg` is required on the *Forward* message in order for the message to reach the Recipient.
 * There are no restrictions on the language used to describe the schema so long as both parties (sender, router) understand it.
 
-The combination of `forwardMsgReq` and `forwardMsg` enable more sofisticated gatekeeping as described in [Indy HIPE 0022 - Cross Domain Messaging](https://github.com/hyperledger/indy-hipe/tree/master/text/0022-cross-domain-messaging).
+The combination of `forwardMsgReq` and `forwardMsg` enable more sophisticated gatekeeping as described in [Indy HIPE 0022 - Cross Domain Messaging](https://github.com/hyperledger/indy-hipe/tree/master/text/0022-cross-domain-messaging).
 
 ## Reference
 


dhh1128:
@llorllale , I don't understand why forwardMsgReq is necessary. It seems like you are trying to distinguish between a case where the routing keys belong to a relay (no forwardMsgReq) and a case where the routing keys belong to a mediator (forwardMsgReq is necessary). But by definition, relays don't have keys and don't need to be taken into account by a sender, so the routing keys are inherently keys for a mediator. Thus forwardMsgReq will always have to be true if we add it.

What am I missing?

swcurran:
@dhh1128 - to be a little picky about your previous note, the relays have keys, just not keys that the sender needs to know about.

dhh1128:
You are right, @swcurran -- I was being imprecise. Relays might have keys, or they might not. They don't have to be agents. But for sure, the sender doesn't care one way or the other.

llorllale:
@dhh1128 

The intent of `forwardMsgReq` is to communicate to the Sender that they need to enclose a message for the Recipient's Mediator. `HIPE 0022` has no references to mediators - it should be updated to include Mediators in order to align it with [Aries RFC 0046](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0046-mediators-and-relays/README.md). This would mean specifying the mediator's keys in the DID document shared with the Sender.

The way `forwardMsgReq` would work is:

* Its *presence* signals to the Sender that the Recipient's *Mediator* will do 'gatekeeping' based on a message enclosed specifically for the mediator,
* Its *value* (a URI) lets the Sender know the "shape"/"form" of that message. The semantics of this message would be derived purely from the context in which the connection/relationship is being made

dhh1128:
@llorllale: I still don't get it. If there are routing keys at all, then a mediator (the routing agent) is present, and it is doing gatekeeping, and you must use a `forward` message. I don't see why we need another attribute to declare this; the presence of the `routingKeys` attribute already declares it. Maybe this is because I don't understand what possibilities you are imagining for the shape or form of the message. There is only one forward message that I know about.

swcurran:
@dhh1128 - I think that @llorllale is offering a way for the Receiver to tell the Sender about what mediators to use because he doesn't see how the Sender knows that information.  If that is clarified, I think it will be clear.

I'm not certain of the current accepted state of the DIDDoc Conventions HIPE, but at one point early on it was "assume there is one Domain Endpoint and one Routing Agent", which is pretty lame.

Later, @tplooker added an update (I think to that document) to indicate that in the service endpoint part of the DIDDoc, there would be an array that explicitly provided the routing information to the Sender. Thr array might be empty (send directly using the public key associated with the endpoint), or contain one or more references to keys in the DIDDoc. Each entry is a mediator and requires an additional `forward` message with encryption.

Have I got that right?

dhh1128:
@swcurran I will be interested in confirming comments from @tplooker, but my understanding is that the `routingKeys` array is a list of mediators, and that therefore the serviceEndpoint is a mediator's endpoint, not the endpoint of the agent itself. As such, every key in the `routingKeys` array requires a `forward` message, and the URI that @llorllale proposes in `forwardMsgReq` is redundant with the serviceEndpoint itself.

swcurran:
That's what I'm thinking as well.  My guess (but could be wrong) is that @llorllale is not aware of that mechanism, hence his suggestion.  George, did I guess right, or is there another purpose for the proposed message?

llorllale:
@dhh1128 @swcurran 

Somehow I missed the fact that `routingKeys` is an array. This might explain the confusion. But first, I should explain what we're trying to achieve:

The goal is to enable sophisticated "gatekeeping" as described in [Mediators & Relays - Scenario 2](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0046-mediators-and-relays/README.md#scenario-2-a-gatekeeper). Imagine Bob is a CEO and he instructs Carl to only forward messages from members of the board, directors, etc. To do this, **a verifiable claim needs to be sent by Alice and addressed to Carl (the mediator)**.

One option is to spawn some version of the `0037 Present Proof` protocol *per request* and have Carl and Alice carry out this back-and-forth where he asks her for proof and she supplies it. Only after Alice sends proof and Carl verifies it will her message be forwarded to Bob. This chattiness is undesirable.

A less "chatty" option where Alice sends both the message for Bob *and* proof of board membership **in one request**. The proof would need to be readable by Carl, and would have to be included in the outer envelope meant for him (the `Forward` request). Like in the first option, Carl will forward the message once he verifies the proof.

**This PR establishes a way to inform Alice *a priori* that she's required to send this proof to Carl.** It does not specify how that message should be included in the `Forward` request for Carl - that would be an extension to the RFC that will supsersede HIPE 0022.

Now, coming back to my mistaken reading of `routingKeys` - the `forwardMsgReq` attribute will have to be paired with its respective routing key. `routingKeys` would no longer be an array of strings, it would be an array of objects, eg:
 
```json
"~service": {
    "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
    "routing": [{"key": "8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"}, {"key": "aklsjFW2JKJKj24kjkjkjkjsasl94j", "forwardMsgReq": "did:example:aYgjkk34JDJKLJh;service=schema/exampleschema/1.0/exampletype"}],
    "serviceEndpoint": "https://example.com/endpoint"
}
```
In this example, I've renamed `routingKeys` -> `routing` and I've shown two routers: one with just its key, and another with its key and the new attribute.

tplooker:
@dhh1128, your comments above around the intended usage and function of routing keys is correct. If you have routing keys specified for an endpoint then the last routing key in the array is owned by the service endpoint.

e.g for did:example:123456789abcdefghi that declares the following service in its did doc.

```
{
  "service": [{
    "id": "did:example:123456789abcdefghi;did-messaging",
    "type": "did-messaging",
    "priority" : 0,
    "recipientKeys" : [ "did:example:123456789abcdefghi#1" ],
    "routingKeys" : [ "did:example:123456789abcdefghi#2" ],
    "serviceEndpoint": "https://agent.example.com/"
  }]
}
```

The owner/controller of the endpoint `https://agent.example.com/`, also possesses the private key associated to the key reference `did:example:123456789abcdefghi#2`. This enables the endpoint controller to unpack the encrypted envelope message to reveal the inner forward message that is destined and encrypted to `did:example:123456789abcdefghi#1`.

As a related note, the other way this convention is destined to function is to omit routing keys from the service endpoint of the shared did doc. And use a did service reference in place of the service endpoint. This leads a would be sender to preparing the message to the intended recipients, resolving the service endpoint via the did url. And using the provided information in resolved service endpoint to complete the preparation of the message, including the elected mediators
 
e.g for did:example:123456789abcdefghi that declares the following service in its did doc.

```
{
  "service": [
    {
      "id": "did:example:123456789abcdefghi;did-messaging",
      "type": "did-messaging",
      "priority" : 0,
      "recipientKeys" : [ "did:example:1234abcd#1" ],
      "routingKeys" : [],
      "serviceEndpoint" : "did:example:xd45fr567794lrzti67;did-messaging"
    }
  ]
}
```

And did:example:xd45fr567794lrzti67 declares the following service in its did doc.

```

tplooker:
Based on the evolving interest in this particular topic, I have made it a high priority to modify and transfer the [pending HIPE](https://github.com/hyperledger/indy-hipe/blob/f5726a3cf1c8bdd601a7ea29ffb4a9ad33969f89/text/0023-diddoc-conventions/README.md) across to aries, I will update this thread once I have completed this.

tplooker:
My most pressing concern about this decorator currently is the ambiguity introduced around the routingKey and recipientKey arrays contents. In the original [HIPE](https://github.com/hyperledger/indy-hipe/blob/f5726a3cf1c8bdd601a7ea29ffb4a9ad33969f89/text/0023-diddoc-conventions/README.md) I proposed around the did doc conventions which are related. The routing keys and recipient keys are did key references. Where as in this RFC along with the [did exchange](https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange) protocol, the keys are defined inline. However in this inline definition, what is the formal format of these inline keys? Base64, Base58, PEM, JWK, or further more what type of key is it, RSA, Ed25519? We make assumptions about this key in the did exchange protocol. However this is limiting the usage of other cryptographic suites and could prevent other community players being inter-operable.

tplooker:
As a counter to my above point there is a strong desire to keep brevity as a priority when expressing this convention, as messages often featuring this decorator will be encoded and transferred via a QR code which has a [low limit] (https://stackoverflow.com/questions/11065415/how-much-data-information-can-we-save-store-in-a-qr-code) to the amount of data that can be reliably expressed.

I for one am not opposed to exploring other options around QR codes that would remove the constraint of brevity described above.

swcurran:
> @dhh1128 @swcurran
> 
> Somehow I missed the fact that `routingKeys` is an array. This might explain the confusion. But first, I should explain what we're trying to achieve:
> 
> The goal is to enable sophisticated "gatekeeping" as described in [Mediators & Relays - Scenario 2](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0046-mediators-and-relays/README.md#scenario-2-a-gatekeeper). Imagine Bob is a CEO and he instructs Carl to only forward messages from members of the board, directors, etc. To do this, **a verifiable claim needs to be sent by Alice and addressed to Carl (the mediator)**.
> 
> One option is to spawn some version of the `0037 Present Proof` protocol _per request_ and have Carl and Alice carry out this back-and-forth where he asks her for proof and she supplies it. Only after Alice sends proof and Carl verifies it will her message be forwarded to Bob. This chattiness is undesirable.
> 
> A less "chatty" option has Alice sending both the message for Bob _and_ proof of board membership **in one request**. The proof would need to be readable by Carl, and would have to be included in the outer envelope meant for him (the `Forward` request). Like in the first option, Carl will forward the message once he verifies the proof.
> 
> **This PR establishes a way to inform Alice _a priori_ that she's required to send this proof to Carl.** It does not specify how that message should be included in the `Forward` request for Carl - that would be an extension to the RFC that will supsersede HIPE 0022.
> 
> Now, coming back to my mistaken reading of `routingKeys` - the `forwardMsgReq` attribute will have to be paired with its respective routing key. `routingKeys` would no longer be an array of strings, it would be an array of objects, eg:
> 
> ```json
> "~service": {
>     "recipientKeys": ["8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"],
>     "routing": [{"key": "8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"}, {"key": "aklsjFW2JKJKj24kjkjkjkjsasl94j", "forwardMsgReq": "did:example:aYgjkk34JDJKLJh;service=schema/exampleschema/1.0/exampletype"}],
>     "serviceEndpoint": "https://example.com/endpoint"
> }
> ```
> 
> In this example, I've renamed `routingKeys` -> `routing` and I've shown two routers: one with just its key, and another with its key and the new attribute.

Interesting note - thanks for clarifying.  My first response was scaled to the use case - a small whitelist of DIDs to be allowed through (or blacklist of DIDs not to be allowed through). For a small scale list, a simple solution is just having a mechanism for Bob to tell Carl what target DIDs are to be sent or not sent on.

However, if you take that to the extreme - a retroactively set condition on the messages to be fowarded based on a dynamically changing group - a proof-based approach is likely the preferred choice. For example, Carl is only to forward messages from shareholders of the large public company Bob operates.

An immediate challenge I see is that in the general case, the sender is not known to the mediator, and in fact, the mediator and the sender do not have a connection.  To the sender, the mediator is simply a key in the routingKeys array, and to the mediator, the sender is unknown - all senders are equal.

I think the solution is that **Bob's agent handle this filtering**, not Carl. Bob's agent would have a rule that on receipt of a message, before bothering Bob about it, send back and get back a proof from the sender.  Since a connection is already in place, that connection can be used for presenting the proof.  Bob's Agent can use a sub-protocol to complete the proof process without Bob's involvement. Policies can be used to define the frequency of the proof - e.g. per message, per thread, per time period. Alice could likewise have a policy with her agent to respond to the proof request automagically as appropriate - e.g. always provide the proof when asked by Bob.

Fun to think about. In this case, I think the use case is wrong.  We probably want to remove that example from the mediators RFC.

llorllale:
@swcurran 

Your insights around the state/existence of a connection between Bob and the sender are good, but I disagree with the solution you propose.

**re: "Bob's agent handle the filtering"**
We should not tell Bob, the CEO, to do all this work himself - that's what he hired Carl for. This happens to be a simple scenario where maybe the CEO could do the job themselves if they had the time. But there are other use cases where the mediation process is a specialty and/or is required to be performed by a trusted 3rd party.

**re: policies for getting proof back from the sender per message/thread/time period**
Counterpoints:
* it is desirable to remove this extra back-and-forth for performance reasons
* DIDComm is simplex by default - the Sender may not include a return route in their message
* this extra filtering (some may call it *real gatekeeping*) is aligned with the mediator's conceptual role as a gatekeeper that has been accepted since the Indy HIPE

**re: lack of connection between Bob and the senders**
Indeed, in this scenario Bob's address would have to be public - aka a public DID. This means this information would have to be conveyed there. This also means using the `~service` decorator only works where there is a connection. This, along with @tplooker 's valuable insights [here](https://github.com/hyperledger/aries-rfcs/pull/70#issuecomment-500638192) and [here](https://github.com/hyperledger/aries-rfcs/pull/70#issuecomment-500643676) lead me to believe this information should probably be conveyed in the DID document.

swcurran:
@llorllale - there are existing techniques to handle the use case you've presented, without needing to add anything specific to the use case.  For example, to handle the general case:

* While Bob is establishing a connection to Alice, Bob introduces Carl to Alice so they have a connection.
    * Carl uses that to determine if Alice has the credentials to talk to Bob before Bob completes the connection.
* Bob tells Carl the pairwise-DID/verkey that Alice will use and Carl associates that with his connection to Alice
* For future messages from Alice, Carl requests a proof from Alice
   * Carl may use a cache to limit the number of proofs sent to Alice
* On receipt of the proof, Carl forwards the message from Bob

My point is more that there exists a variety of existing techniques to handle what (I think) amounts to an edge case.  I've given a couple of techniques that could be used.  I don't think as yet we need to have a specific messaging feature to handle this case. There are lots of other more general things we need to be doing.  Mind you, I may be missing the general/common technical use case enabled by this.

I don't understand your "Public DID" comment.  In general, we don't expect there to be very many public DIDs - basically, the only ones would be the public DIDs of issuers.

kdenhartog:
Does anyone know the current state of this work? @tplooker @llorllale @swcurran

I'd like to get this PR moved through if a resolution has been found.

llorllale:
@kdenhartog [selection criteria](https://github.com/hyperledger/aries-rfcs/issues/174) is a bit more pressing for us right now, but we'll definitely end up at this point again. None of the options presented by @swcurran make sense for our use case.

swcurran:
Glad I could help :-).

llorllale:
@swcurran apologies - your input is always welcome of course :)

kdenhartog:
@dhh1128 @llorllale with the new forward protocol capabilities, is this still needed?

llorllale:
Closing this for now - revisiting later.

PR #71, opened 2019-06-07
Describe potential payment flow for Credential Issuance.

jovfer:


jovfer:
@dhh1128 thanks for review. Agreed on "more generic" for payment sentence. Will fix it soon.

btw link to the Sam's PR is already in current proposal, right after a description. My point was to reference Indy's PR and give an example with potential payment decorator structure (not full decorator description).
If migration (Indy->Aries) of Sam's PR will be approved - I also like the idea to reference neighbor RFC instead of some poor-defined example and remove sections `Payment request` and `Payment comfirmation`

dhh1128:
@jovfer Can you update this to hyperlink to RFC 0075 about payment decorators, as soon as that RFC is merged? Then we can merge this. See https://github.com/hyperledger/aries-rfcs/pull/75

kdenhartog:
> @jovfer Can you update this to hyperlink to RFC 0075 about payment decorators, as soon as that RFC is merged? Then we can merge this. See #75

I've just merged #75 so this shouldn't be a blocker anymore.

jovfer:
Kyle, thanks
Daniel, sure, will update this PR today

PR #75, opened 2019-06-11
propose RFC: payment decorators

dhh1128:
This supersedes indy-hipe PR #129.


features/0075-payment-decorators/README.md
```diff
@@ -0,0 +1,335 @@
...
+a credit card processing service. The payee emits a [PaymentRequest](
+https://www.w3.org/TR/payment-request/#paymentrequest-interface) JSON structure (step 1
+below); this causes the payee to be prompted (step 2, "Render"). The payer decides
+whether to pay, and if so, which payment method and options she prefers (step 3, "Configure").
+The payer's choices are embodied in a [PaymentResponse](
+https://www.w3.org/TR/payment-request/#paymentresponse-interface)
+JSON structure (step 4). This is then used to select the appropriate codepath and inputs to
+invoke the desired payment method (step 5).
+
+[![API flow](payment-request-api-flow.png)](payment-request-api-flow.puml)
```
llorllale: @dhh1128 the PaymentResponse is sent back to the payee after the user has made a payment

dhh1128: @llorllale That was my original understanding, too. However, when I went and studied the spec carefully, I could not find any mention of that. And the one good example that they give of a PaymentResponse, BasicCardResponse, contains card info that the payee shouldn't have if they're not processing a payment. So I came to doubt my first intuition. Can you point me to a section of the spec that clarifies?

kdenhartog: I'm going to approve this so we can get this into the repo. Let's move this conversation to the issues section.

PR #76, opened 2019-06-12
propose RFC: Did communication Did document conventions

tplooker:
Revises this [PR](https://github.com/hyperledger/indy-hipe/pull/92/files) and this previously accepted [proposal](https://github.com/hyperledger/indy-hipe/tree/master/text/0023-diddoc-conventions).


features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+    - The sender has created a [content level message](../../concepts/0021-didcomm-message-anatomy/README.md) that is now ready to be prepared for sending to the intended recipient(s). 
+
+1. The sender resolves the relevant `did-communication` service of the intended recipient(s) Did document.
+2. The sender resolves the recipient keys present in the `recipientKeys` array of the service declaration.
+3. Using the resolved keys, the sender takes the content level message and packs it inside an [encrypted envelope](../../concepts/0021-didcomm-message-anatomy/README.md) for the recipient keys. (*note-2*)
+4. The sender then inspects the `routingKeys` array, if it is found to be empty, then the process skips to step 5. Otherwise, the sender prepares a content level message of type `forward`. The resolved keys from the `recipientKeys` array is set as the contents of the `to` field in the forward message and the encrypted envelope from the previous step is set as the contents of the `msg` field in the forward message. Following this, for each element in the `routingKeys` array the following sub-process is repeated:
```
llorllale: @tplooker I think the value of the `to` field should just be a DID with no fragment (ie. just the `"did:" method-name ":" method-specific-id` portion of the key reference)

features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Motivation
+
+Standardization of these conventions is essential to promoting interoperability of Did communications.
+
+## Tutorial
+
+### Did documents
+
+A Did document is the associated data model to a Did, it contains important associated cryptographic information and a declaration of capabilities the Did supports.
```
llorllale: @tplooker I suggest we fully capitalize "DID" everywhere

features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+- recipientKeys : This is an array of [did key references](https://w3c-ccg.github.io/did-spec/#public-keys) used to denote the default recipients of an endpoint. (*note-1*)
+- routingKeys: This is an array of [did key references](https://w3c-ccg.github.io/did-spec/#public-keys), ordered from most destward to most srcward, used to denote the individual routing hops in between the sender and recipients. See [TBC]() for more information on how routing is intended to operate.
+- serviceEndpoint : Required by the [Service Endpoints Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints). This URL based endpoint is used to declare how the message should be sent. Did communication is transport agnostic, and therefore leverages existing application level transport protocols. However for each transport defined, which is identified by the URL scheme e.g `http`, a set of transport specific considerations are defined see [transports](../0025-didcomms-transports/README.md) for more details.
+
+>Notes
+>1. The keys featured in this array must resolve to keys of the same type, for example a mix `Ed25519VerificationKey2018` or `RsaVerificationKey2018` in the same array is invalid.
```
llorllale: @tplooker why must they be of the same type?

features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+    "serviceEndpoint": "https://agent.example.com/"
+  }]
+}
+```
+
+- id : Required by the [Service Endpoints Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints).
+- type : Required by the [Service Endpoints Spec](https://w3c-ccg.github.io/did-spec/#service-endpoints). It is mandatory that this field is set to `did-communication`
+- priority : This represents the priority of the service endpoint, used for distinction when multiple `did-communication` service endpoints are present in a single Did document. It is mandatory that this field is set to an unsigned integer with the default value of `0`.
+- recipientKeys : This is an array of [did key references](https://w3c-ccg.github.io/did-spec/#public-keys) used to denote the default recipients of an endpoint. (*note-1*)
+- routingKeys: This is an array of [did key references](https://w3c-ccg.github.io/did-spec/#public-keys), ordered from most destward to most srcward, used to denote the individual routing hops in between the sender and recipients. See [TBC]() for more information on how routing is intended to operate.
```
llorllale: @tplooker missing link for TBC

dhh1128:
@tplooker I will merge this as soon as you've addressed some of @llorllale 's comments. We can then continue the conversation to nuance it via github issues.

cywolf:
@tplooker I find it a bit strange personally for the routingKeys to be references to entries in the publicKeys block. We currently take all those public keys and put them into the reverse index to identify messages from that sender, but the routing keys would (should) never be used that way. I'm also not sure if the controller for those keys is technically the same DID?

features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+5. Resolve the service endpoint:
+    - If the endpoint is a valid Did URL, check that it resolves to another Did service definition. If the resolution is successful the process from step 2. is repeated using the message outputted from this process as the input message. 
+    - If the service endpoint is not a Did URL, send the message using the transport protocol declared by the URL's scheme.  
+
+> Notes
+>1. There are two main situations that an agent will be in prior to preparing a new message.
+>>-  The agent is responding to a message that has just been received and has the context of the sender key of the previous message.
+>>-  The agent is creating a new message to a connection and will use the default `did-communication` service convention for preparation of a message.
+>With the first case. A targeted lookup of the `did-communication` service definition could be done to find a service definition that features the sender key as a recipient key which would ensure that the response was delivered back to the sender.
```
llorllale: @tplooker can you please insert a blank line above this paragraph? Then it will render separate from the above bullet point.

Also, I don't understand this paragraph clearly - I don't see how the service definition can be identified with a key reference.

features/0067-didcomms-diddoc-conventions/README.md
```diff
@@ -0,0 +1,180 @@
...
+>With the second case. The default `did-communication` service description would be used by resolving the lowest priority service definition from the connections Did document
```
llorllale: @tplooker I don't understand the reason for this. Can you explain with a use case?

features/0067-didcomms-diddoc-conventions/README.md
tplooker: @llorllale this is due to the routing RFC not being proposed in Aries yet, my intention was to correct this once the RFC is defined?

features/0067-didcomms-diddoc-conventions/README.md
tplooker: @llorllale, basically fall back behavior or redundancy. You could define another endpoint for the same recipients with a higher priority that could be used in the event that the preferred endpoint fails.

features/0067-didcomms-diddoc-conventions/README.md
tplooker: @llorllale the current thinking is that the to field is actually the raw public key value(s). Because when the routing identifier is a did or a did key reference, this implies the intermediary (or mediator agent is what we have previously referred to) needs to know the did. When the did is pairwise, this involves authenticated exchange of this did between the owner and the mediator. Practically this adds quite alot of overhead to how routing is conducted too. Rather when routing is based on public keys, it is far lighter and proving to a mediator you own the public key and hence should be able to create a route is far more light weight. I realize that without the routing RFC much of this is unclear so I will endeavor to get that into a proposal phase, language in this RFC can then also be cleaned up.

features/0067-didcomms-diddoc-conventions/README.md
tplooker: @llorllale please have a look at https://github.com/hyperledger/aries-rfcs/tree/master/features/0019-encryption-envelope basically with our current encryption envelope when you encrypt to multiple recipients, the outputed JWE structure assumes the recipient keys are of all the same type. Supporting multiple recipients encrypted in a common JWE using different encryption primatives is out of scope for JWE's I believe too. @kdenhartog perhaps you can provide some additional commentary to this point?

tplooker: The point I am trying to make is that if you resolve a service decleration from a did doc and prepare a message for that service it should result in one encrypted envelope destined to be sent to one endpoint.

tplooker:
> @tplooker I find it a bit strange personally for the routingKeys to be references to entries in the publicKeys block. We currently take all those public keys and put them into the reverse index to identify messages from that sender, but the routing keys would (should) never be used that way. I'm also not sure if the controller for those keys is technically the same DID?

The problem with these being inline keys is we would have to have an inline public key object i.e 

The following is invalid IMO
```
"service": [
    {
      "id": "did:example:123456789abcdefghi;did-communication",
      "type": "did-communication",
      "priority" : 0,
      "recipientKeys" : [ "8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K" ],
      "routingKeys" : [ "did:example:1234abcd#3" ],
      "serviceEndpoint" : "did:example:xd45fr567794lrzti67;did-communication"
    }
  ]
```
Because what encryption primitive is `8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K` using and what format is that key in base64, base58, JWK, PEM?

So the alternative would be to permit
```
"service": [
    {
      "id": "did:example:123456789abcdefghi;did-communication",
      "type": "did-communication",
      "priority" : 0,
      "recipientKeys" : [ {
           "id": "did:example:123456789abcdefghi#keys-2",
           "type": "Ed25519VerificationKey2018",
           "controller": "did:example:pqrstuvwxyz0987654321",
           "publicKeyBase64": "8HH5gYEeNc3z7PYXmd54d4x6qAfCNrqQqEB3nS7Zfu7K"
       } ],
      "routingKeys" : [ "did:example:1234abcd#3" ],
      "serviceEndpoint" : "did:example:xd45fr567794lrzti67;did-communication"
    }
  ]
```
Which I do not have a problem with allowing. Does anyone else have thoughts on this?

To your point that the routing key might not be owned by the current did, correct but I dont see how a did key reference to the routing agents public did is a problem?

kdenhartog:
Let's continue the discussion in the github issues rather than here. Otherwise this PR will hang until problems get resolved. I'll open an issue over there to provide my opinion on it.

features/0067-didcomms-diddoc-conventions/README.md
llorllale: @tplooker +1 for updating the routing RFC

PR #24, opened 2019-05-27
propose RFC 0024: DIDCom over XMPP

Oskar-van-Deventer:
Signed-off-by: Oskar van Deventer oskar.vandeventer@tno.nl

PR #83, opened 2019-06-14
re-propose RFC: DIDComm over XMPP

dhh1128:
This replaces PR #24. I'm going to leave the RFC numbered as 0024 in the doc because the new PR number is just a cluttering detail.


Oskar-van-Deventer:
Daniel, 
Excellent, thank you for your help!
Oskar

PR #93, opened 2019-06-18
PROPOSE RFC: Return route

TelegramSam:


features/0092-transport-return-route/README.md
```diff
@@ -0,0 +1,88 @@
+# 0092: Transports Return Route
+- Author: Sam Curren <sam@sovrin.org>
+- Start Date: 2019-03-04
+
+## Status
+- Status: [PROPOSED](/README.md#rfc-lifecycle)
+- Status Date: 2019-06-18
+- Status Note: Supersedes [INDY HIPE PR 116](<https://github.com/hyperledger/indy-hipe/pull/116>)
```
dhh1128: @TelegramSam The <> pattern is a markdown hyperlink short-cut. The [...](...) form is the long form. Combining them is unnecessary and will break the the hyperlink unit test. Can you please remove the <>?

features/0092-transport-return-route/README.md
```diff
@@ -0,0 +1,88 @@
...
+- Application varies between transports, adding complexity to agent development.
+- Requiring outbound message queuing has agent architectural implications.
+
+## Rationale and alternatives
+
+- Using a transport level decorator makes this behavior transport agnostic.
+
+## Prior art
+
+The [Decorators RFC](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0011-decorators/README.md) describes scope of decorators. Transport isn't one of the scopes listed.
```
dhh1128: Please change the hyperlink to the decorators RFC to be relative (../../concepts/0011-decorators/README.md). We shouldn't be referring to the branch in our hyperlink.

features/0092-transport-return-route/README.md
dhh1128: (I will update the unit test to tolerate this, though.)

features/0092-transport-return-route/README.md
TelegramSam: good catch.

PR #95, opened 2019-06-19
[Proposed RFC] Basicmessage

TelegramSam:


features/basic-message/README.md
```diff
@@ -0,0 +1,77 @@
+# 0000: Basic Message
```
dhh1128: The number should be updated to 0095.

dhh1128: I suggest you add the word "Protocol" and the version "1.0" to the title at some point.

dhh1128: The folder name needs to be changed from "basic-message" to "0095-basic-message". Then the index.md needs to be updated again.

dhh1128:
@TelegramSam Now /index.md needs to be updated. Then we can merge.

PR #98, opened 2019-06-25
allow for encoding of values in credential previews

sklump:

For example, [https://jira.hyperledger.org/browse/IS-1281](JIRA IS-1281) calls for a binary attachment.


sklump:
Note that I've only hinted at 'base64' or the absence of the encoding. I haven't specified what else might be supported and it may get more complex the deeper we go into this question. Binhex? base64 variants? PEM/uuencode? base58? base32? base85? URLencode? Who can say what might emerge? I wanted to put encoding in as a possibility but leave the semantics underspecified for the moment.

Since I'm the first one in, I know I will stick to base64 (because I like powers of 2) but I don't want to support anything else for the moment until I get a proof of concept working.

Issue #101, opened 2019-06-26
DID Resolution work


kdenhartog:
As we all know DID Resolution is a component we have to grapple with and there's a lot of differing opinions on this subject. I'm earmarking this issue for discussing this work so that we can communicate about the topic in between calls related to the topic. @tplooker I know you had some work on this and @swcurran had some ideas as well. Would you mind adding your proposal here to kick off the discussion? Pinging @esplinr @jovfer @dhh1128 @ashcherbakov and @troyronda as well because I believe they would be interested in the discussion.

dhh1128:
I'm mentioning @peacekeeper and @talltree because I think both of them should track this conversation as well.

tplooker:
I believe we should be pursuing architecture similar to that pre-existing with the uniresolver @peacekeeper is the below an accurate design representation?

![Screen Shot 2019-07-01 at 8 31 11 AM](https://user-images.githubusercontent.com/15972525/60401801-a67e9480-9bda-11e9-880f-409787cfcff9.png)

Essentially, we should have a pluggable driver interface that allows registration of drivers to handle different did methods. This allows for both local and remote did resolution depending on the did method.

talltree:
Tobias, what you are describing in your diagram is exactly what I believe
Markus was describing to me when we talked last Friday.

The one difference between this and what Stephen has been describing is
that you are anticipating a DID resolver function that has plug-in drivers,
whereas Stephen is describing that all DID methods are just protocol
modules(either dedicated to a DID method or integrated in with a VC
exchange protocol). In that case, a remote DID resolver protocol would be
just another protocol module.

tplooker:
@talltree thanks, yes @swcurran I believe your approach would bring did resolution up into the protocol layer? I do see how this would benefit community velocity in the short term but I think it would become a problem in the longer term. 

To me DID comms should have DID resolution as a base function rather than did resolution being an artefact of a particular DID comm message protocol. Put differently, I see DID comms as an arbitrary messaging standard between DIDs, if thats the case then DID resolution is a primitive and the application layer should not introduce the notion of DID resolution.

tplooker:
Subsequent to the above diagram, this is how I would envisage peer dids being managed, simply as another driver that instead is designed to query what ever the pluggable storage mechanism is rather than a ledger or remote resolver.

[...]

PR #115, opened 2019-07-09
propose RFC: predefined identities

dhh1128:


TelegramSam:
you just barely beat me to it.

features/0114-predefined-identities/README.md
```diff
@@ -0,0 +1,345 @@
...
+The collective Aries developer community is represented by:
+
+```
+did:sov:BzCbsNYhMrjHiqZDTUASHg -- verkey = 6zJ9dboyug451A8dtLgsjmjyguQcmq823y7vHP6vT2Eu
+```
+
+This DID is currently allocated, but not actually registered on Sovrin's mainnet.
+You will see this DID in a number of RFCs, as the basis of a [PIURI](
+ ../../concepts/0003-protocols/uris.md#piuri) that identifies
+a community-defined protocol. You DO NOT have to actually resolve this DID or relate
```
troyronda: @dhh1128 @kdenhartog I am curious about the advantages of did:sov:BzCbsNYhMrjHiqZDTUASHg vs a hypothetical did:peer:<> (or even did::<> as in a special namespace)  for predefined identities in Aries. It might be worth spelling out a bit more.

Since:
-  did:sov:BzCbsNYhMrjHiqZDTUASHg  is "not actually registered on Sovrin's mainnet"; 
- "You DO NOT have to actually resolve this DID";
- did:peer method is being discussed as potentially core in https://github.com/hyperledger/aries-rfcs/issues/101

p.s., what is the meaning of "DID is currently allocated".

peacekeeper: ... and the most important question, who holds the private key for BzCbsNYhMrjHiqZDTUASHg? :)

swcurran: The "magic DID" is controlled by Daniel Hardman at this time - he created it and has the private key.

My/Our hope in agreeing to use DIDs for namespace separation is that the specifications will eventually become machine consumable equal with the human readable specs. I/we hope the DIDs will eventually resolve to physical addresses - e.g. a service endpoint that can be used for machine and human processing.  For example, the service endpoint lead to (for example) a folder in a GitHub repo containing the spec.  We're not currently doing that with the use of Aries RFCs, but I/we hope to get there.

dhh1128: I allocated the DID while a large group of us were sitting around a table at an "agent summit" meeting at Sovrin HQ a year ago. I still have the private key sitting in my password manager. This should probably change. I was thinking that I would share the private key with anyone who has official maintainer status over the Aries RFCs repo. That way if I get hit by a bus, we are not at risk of losing control of the DID. Does that seem like a good plan to you?

dhh1128: PS. To specifically address @troyronda 's question about the meaning of "allocated": I went through a normal process of creating a DID and its keys, but I never submitted the ledger transaction that would register it. This is what is meant by "allocated".

The reason I didn't register it is that Sovrin's DID doc support was simpler than I wanted; instead of allowing multiple keys, it only allowed one. And we didn't have a service endpoint ready to register -- an http(s) endpoint where people could browse to documentation on the protocols we were creating., using a URI pattern where you could find any protocol by appending it as part of the path or query string of the URI. We *did* have an https endpoint in github, but it didn't allow query strings and it assigned paths based on HIPE title rather than protocol name.

Maybe we should go ahead and register the DID anyway. I don't really have that strong of an opinion. We just didn't feel the need at the time. Mostly what we wanted was something URN-like, but DID-based.

Advice would be appreciated.

troyronda: @dhh1128 @swcurran Here is a hypothetical alternative for comparison:

If the "magic" DID were a peer DID, there could be an RFC that defines both the did:peer:id along with the associated DID document (to allow for resolving service endpoints). Aries implementations could include that magic "DID" and DID document, as defined by the RFC. Implementations could update this DID/DID document as the RFC evolves.

Note: control remains with the maintainers (via the RFC update process).

dhh1128: I see some advantages of that approach. One that I particularly like is that it removes the perception that there's some dependence on the Indy/Sovrin world in Aries. There really isn't--as I've been at pains to point out, the did:sov we're tallking about is effectively an opaque uuid-like thing for all intents and purposes. But I still don't like having to explain it. Managing the peer DID's DID doc would also be easier, and it would contribute to the momentum of peer DIDs, which is a good thing, generally.

The only drawback I see is that this particular value is embedded in many places, so we'd have a large search-and-replace, both in RFCs and in codebases that implement those RFCs. But if we get a few maintainers-of-codebases-that-implement-Aries-RFCs to agree, I would cheerfully proceed.

troyronda: Created issue #129

PR #116, opened 2019-07-10
Initial proposal for Evidence Exchange Protocol

vinomaster:
Aries RFC proposal submission for 0116-Evidence Exchange Protocol.


kdenhartog:
This is [documentation](https://github.com/hyperledger/indy-sdk/blob/5f9144feab5b9e969389e325da47d7218f87081e/docs/contributors/signing-commits.md) to fix the DCO fix

As for fixing the CI pipeline: It looks like the links for `footnote1` and `footnote2` in digital_notrary_usecase.md file. If you update that to properly link to something it should be fixed.

PR #123, opened 2019-07-12
Fix JSON-LD bugs

peacekeeper:


dhh1128:
@peacekeeper Thanks for this fix. It exposes a doubt that has nagged at me for the past 18 months, and I'm not surprised that you submitted a fix, since I knew when I wrote this that I wasn't following the pattern in the DID spec--but I've never understood those patterns. Can you please teach me more about them? I told @msporny long ago that I felt like DID Docs were confusing the distinct concepts of "ID" and "ID ref" from XML and even SGML days. Is this a JSON-LD-ism that I don't understand? How can you distinguish between an ID and a reference to an ID, when both appearances of it have the same form? Is it because only one of the appearances of the value is associated with a JSON key named "id"? To me, it makes sense that # would be used to navigate to a fragment from either within or outside a document, but it seems very weird that the "fragment navigation" token would be part of the ID value itself. Given the notation you are suggesting, if I wanted to refer to "key-1" inside the doc, I would expect to have to do "did:xyz##key-1", not "did:xyz#key-1".

peacekeeper:
> Is it because only one of the appearances of the value is associated with a JSON key named "id"?

Yes that's basically it. Or do be more precise, `id` is defined as a term in the JSON-LD context to be equal to `@id`. If you open the DID [JSON-LD context](https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld), you will see that it says in the beginning:

`"id": "@id",`

`@id` is a core, built-in keyword in JSON-LD which means that the value `#key-1` in your DID Document is the identifier (relative to the base IRI) of the node (an RDF resource) described by the current JSON-LD object.  If the same value `#key-1` appears as the value of a different term (e.g. `publicKey`), then it's a reference (an RDF property) pointing to the node. This assumes that the property `publicKey` is also defined in the JSON-LD context with `"@type": "@id"`, again you can also see that in the DID [JSON-LD context](https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld).

dhh1128:
Thank you. That helped me understand how JSON-LD overloads the meaning of "id" (although not why this was thought to be a good idea).

But half of my puzzlement remains.

Section 3.5 of RFC 3986 (https://tools.ietf.org/html/rfc3986#section-3.5):

```
   The semantics of a fragment identifier are defined by the set of
   representations that might result from a retrieval action on the
   primary resource.  The fragment's format and resolution is therefore
   dependent on the media type [RFC2046] of a potentially retrieved
   representation, even though such a retrieval is only performed if the
   URI is dereferenced.
```
So, a fragment identifier is what is passed to the handler of the media type (the JSON-LD library, I suppose). This identifier *cannot* include the # character, since the ABNF of `fragment` is:

```ABNF
fragment    = *( pchar / "/" / "?" )
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```
...where # is in the general reserved chars and not in `pchar`. The absence of # from the fragment is also made clear early in section 3:

```
URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

         foo://example.com:8042/over/there?name=ferret#nose
         \_/   \______________/\_________/ \_________/ \__/
          |           |            |            |        |
       scheme     authority       path        query   fragment
```

So, if "#" is not a legal character in a fragment, and that's what the JSON-LD library can count on receiving from an external piece of software, does the JSON-LD library prepend the "#" to the value of the fragment before it goes looking for it? If so, why?

msporny:
> That helped me understand how JSON-LD overloads the meaning of "id" (although not why this was thought to be a good idea).

JSON-LD was a mechanism that was designed to be "additive" to existing JSON documents. We had considered saying "The way you specify the identifier for something in the graph is by using the `id` keyword", but then, LOTS of developers complained that there are tons of JSON documents out there that use `id` to specify identifiers that are not URLs... and that's where that funky `@` syntax came from for all JSON-LD keywords. We also had a LOT of developers complaining simultaneously that if we kept the `@` syntax, that pure JSON developers would find that *weird* because no self respecting JSON developer uses microsyntaxes in their markup. :)

... and that's what typically happens in standards, you have half of the developer community wanting one thing and the other half wanting something diametrically opposed to the initial set. In order to cast as wide of a net as possible, without creating so much cognitive dissonance that neither side would pick up the technology, we said - "Ok, so we'll use the `@` character for JSON-LD keywords to ensure we don't conflict with existing JSON developers that might want to use JSON-LD, and then provide an *aliasing* mechanism that lets developers avoid the microsyntax stuff so that the JSON-LD looks and feels like standard JSON in the vast majority of the use cases".

There are a plethora of trade-offs that the decision above kicked off that would take much longer to explain than I have time right now, but hopefully the above answers the question you asked above, @dhh1128.

msporny:
> URI RFC says that the fragment should be identified inside the doc by "nose", not "#nose"

Unfortunately, I'm rushing to try and get the VC spec wrapped up, so I can't dive into the specific language you're referring to. My expectation is that you're mis-interpreting the specification (based on how I have experienced these things as implemented by various libraries).

Let me try to explain from a different direction:

Let's say you have a document at this URL: https://example.com/foo

... and let's say that the MIME Type for that document allows you to use fragment identifiers to identify "things". Typically, there will be two ways to "reference those things" ... implicitly, and explicitly. I will assert that there are good reasons for doing both (based on whether or not your document always stays at https://example.com/foo or if it might be moved to https://example.com/fnurt).

If I wanted to talk about https://example.com/foo#bar, **within the document identified by https://example.com/foo** then I could reference it implicitly by using:

```
#bar
``` 

or explicitly, by using:

```
https://example.com/foo#bar
``` 

If I did the following in https://example.com/foo: 

```
bar
``` 

Then, and here's where it gets really confusing, depending on the rules for that MIME Type on URI construction, and we haven't clarified this for DID Documents, and the use of BASE and the alignment of the moons, I may end up with https://example.com/foobar OR https://example.com/bar (the MIME Type is supposed to specify how fragments in documents are handled, IIRC... and HTML did something weird with &lt;a id="frag"> and confused everyone).

I think we should be explicit in DID Documents (did:example:123456789#foo), as that leaves no room for confusion, which is a good thing at the expense of the extra characters. We SHOULD support relative `#foo`, but tell people to please not do that. We will also need to define what using `foo` alone means... does it mean (did:example:123456789#foo, did:example:123456789foo, did:example:123456789/foo, or something else)?

... and remember, if you have a strong opinion on the above, you are just one developers and there is a high likelihood that some other developer feels just as strongly about the opposite opinion... and that's what the standards process is about (when at its best), picking one way and then creating a logical argument why that's the best way while also ensuring that many other developers will naturally come to that conclusion as well even without knowing the actual reason for why that makes sense.

Hope that helps.

features/0114-predefined-identities/README.md
```diff
@@ -92,7 +92,7 @@ In demos or tests, its genesis DID Doc looks like this:
         "serviceEndpoint": "https://localhost:23456"
     }],
     "publicKey": [{
-        "id": "key-1",
+        "id": "#key-1",
```
msporny: This is more correct, but I still think we should use explicit URLs. For example, if this document were to be lifted and placed on a website and served from this URL: https://example.com/foo, this fragment might change from did:example:1234#key-1 to https://example.com/foo#key-1 -- and that would change the meaning of the document quite dramatically in ways that we should explore.

peacekeeper:
For reference, links to related issues in the DID specs:

- https://github.com/w3c-ccg/did-spec/issues/97
- https://github.com/w3c-ccg/did-resolution/issues/9

dhh1128:
I am going to go ahead and merge this. Thank you, @peacekeeper, for raising the PR, and thank you, @peacekeeper and @msporny for patiently explaining. I am not feeling aligned with all of the thinking here, but I think I should work that out in DID spec tickets, not here. In the meantime, we should be aligned with the spec.

PR #128, opened 2019-07-17
tighter specification with presentation preview

sklump:
Defeat bait-and-switch for credential inclusion on distinct cred defs with overlapping attribute names; allow for specifiation therein of propositions and non-revocation timestamps.

For consideration at this point. It may be indy-centric; see *Drawbacks* text.


sklump:
I hate to bend the ideal to the pragma, but Marshmallow has a visceral dislike for what I've done here, and Marshmallow is the name of the game for aries-cloudagent-python. I'm going to turn this inside-out and submit another so that there are known constant slugs on the left hand side of the presentation preview. Since I missed a brace anyway, I don't feel quite as bad re-submitting.

Issue #134, opened 2019-07-19
RFC 0116: Should evidence be limited to physical documents?


vinomaster:
The initial motivation for this protocol was to address audit/regulatory/compliance concerns associated with KYC processing that *may* demand access to original source documents. 

Based on conversations with sponsor users. there may be another use of the protocol, specifically as a conduit for the sharing of non-physical evidence such as cryptographic security features that aid in the building of confidence around levels of identity proofing assurance.

vinomaster:
Resolved with PR #141

Issue #135, opened 2019-07-19
RFC 0116: Protocol should discuss to NIST identity assurance levels wrt evidence


vinomaster:
The protocol should take into consideration the guidance [NIST Special Publication 800-63-3 ](https://pages.nist.gov/800-63-3 )places on evidence.

vinomaster:
Resolved with PR #141

PR #141, opened 2019-07-28
Resolve Issues #134 and #135

vinomaster:
Revamped protocol and user stories to include NIST IAL2  and IAL3 assurances as well as support for Digital Assertions from Identity Proofing Service Providers.


dhh1128:
The changes in this PR are straightforward polishing, not something that needs substantive debate. Merging.

PR #161, opened 2019-08-06
Update references to Indy Agent across several RFCs

dbluhm:
Updated some of the commentary about Indy Agents in the Agent explainer RFC and corrected info about Indy Agent Python in the known implementation tables.


swcurran:
The references should be to Aries Cloud Agent - Python, not Indy.

dhh1128:
@swcurran , @dbluhm is just fixing a copy/paste error in my data. All of the places where he's changing a label to say "Indy Cloud Agent" or removing the mention of BC Gov already have a line about Aries Cloud Agent. They *also* have a line about the Indy reference cloud agent, but I inadvertently copied the description from the line about aca.py, which was confusing. So the result of the PR is to clarify that both aca.py and the old Indy Agent repo embody many of the RFCs here.

PR #164, opened 2019-08-06
DID Exchange demoted from accepted to proposed

TelegramSam:


swcurran:
Thanks for doing this and adding in the connection protocol.

+1 from me.  Should this remain for a couple of weeks for the change?

dhh1128:
I don't think it needs to remain. We've already discussed it quite a bit on rocketchat.

PR #167, opened 2019-08-07
WIP: 0167 - Consent Lifecycle

JanLin:
Signed-off-by: JanL <jan.lindquist@dativa.com>


concepts/0167-consent-lifecycle/README.md
```diff
@@ -0,0 +1,668 @@
+# Aries RFC 0167: Consent Lifecycle
+
+- Authors: \
+    Jan Lindquist, Dativa\
+    Mark Lizar, OpenConsent\
+    Harshvardhan J. Pandit, ADAPT Centre, Trinity College Dublin
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2019-08-07
+- Status Note: This RFC is based on [[old
```
swcurran: Hi Jan, the purpose of the "Supersedes" field is to include the link in that field so the Status Note doesn't have to reflect that.  Could you please update to move the link/text down to that field and reword the status note accordingly?

concepts/0167-consent-lifecycle/README.md
```diff
@@ -0,0 +1,668 @@
...
+
+![](media/consent_overview.png)
+
+Concepts
+--------
+
+These are some concepts that are important to understand when reviewing
+this RFC.
+
+**Secondary Data Controller**: The terms data subject and data
```
swcurran: I think it would be helpful to put this in terms of SSI as well.  How do issuer, holder and verifier align with the "** data controller" terminology?  Does it vary by use case?

llorllale: I strongly agree with this suggestion.

JanLin: Added a reference to SSI. Plan to use terms more frequently when explaining the use cases in more detail.

dhh1128:
@JanLin : Thanks for your work on this so far. I did an editorial pass and I have raised a PR against your PR. If you accept it, it will fix the DCO problems with some earlier commits (where the commit message lacked the `Signed-off-by` metadata). Only accept it if you agree to be bound by the DCO terms and conditions (basically, certifying that you're contributing IP you have the right to contribute, with no encumbrances).

The diffs look bigger than they actually are, because I did some reformatting to make the tables and bulleted lists look a bit nicer, and to make the doc as a whole look a bit more similar to other RFCs. That's minor details, and we could merge this without those updates if you prefer.

Here are the substantive changes I made:

1. I changed some of the uses of the phrase "consent receipt" -- including in the title of the doc and in its folder name -- to "data consent receipt". Consent is much broader than the issues raised in GDPR, and I expect someday that we'll have a protocol for giving consent for a DNA test, consent to receive advertising (opt-in to spam), etc. In that future day, I don't want people to think that this RFC covers *all* of that territory; it is only about consent to use personal data. I hope you're okay with this clarification.

2. I added a few hyperlinks.

3. I reordered two of your footnotes, because the references to them occurred in the opposite order. That is, the running text had notes in the order [1], [3], [2], and I changed the references to be [1], [2], [3] by reordering the footnotes.

dhh1128:
I have one other suggestion/request before I would vote to merge:

On your firsst diagram, I suggest that you place the holder, not the issuer, in the middle. Most other diagrams involving issuer, holder, and verifier do this, so that reading from left to right matches chronological flow. That order also simplifies the arrows, so the one from verifier to holder doesn't have to go behind the issuer. This makes it clearer. I also suggest that you rename the bottom rectangle to "Decentralized Registry (e.g., Blockchain)". This makes your diagram fit better with peer DIDs, which are not required but which ought to be compatible with what you are describing.

dhh1128:
Due to the gnarly problems with merge conflicts and DCO, this PR was replaced with a cleaner version of the same thing: #179 . Therefore, closing this one as redundant.

PR #184, opened 2019-08-13
revisit presentation preview to accommodate self-attested and non-rev…

sklump:
…ealed attributes.

The only (attribute) claim filter criteria for which I can imagine a use case in presentation preview are cred-def-id and schema-id: the holder proposes, and so the holder already knows the cred defs on which he has credentials to present. Explicitly inviting less stringent criteria seems to me to invite bait-and-switch style shenanigans. Note pedantically that in this context each attribute under proposal is a claim, not a credential.

I believe we should still keep this RFC status as PROPOSED as prototypical implementations may reveal further missing amenity.

### DIFF
--- a/features/0037-present-proof/README.md
+++ b/features/0037-present-proof/README.md
@@ -128,7 +128,8 @@ This is not a message but an inner object for other messages in this protocol. I
                 "cred_def_id": "<cred_def_id>",
                 ...
             },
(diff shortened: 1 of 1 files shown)


sklump:
Prototyping has shown my approach here for the presentation preview does not accommodate non-revocation status well. Further updates to come.

PR #185, opened 2019-08-13
Added reference code for Lox

mikelodder7:
I added a reference implementation for how lox can be implemented as a library and a command-line tool. The idea is that wallet implementations use Lox to open the wallet. The Lox library can be expanded to be compatible with other secure enclaves like TEEs and HSMs.


kdenhartog:
This is awesome work! Thanks for the reference code.

PR #187, opened 2019-08-15
Revisit presentation preview

sklump:
to allow for unrevealed attrs and self-attested attrs; remove non-revocation timestamps from presentation preview.

### DIFF
--- a/features/0037-present-proof/README.md
+++ b/features/0037-present-proof/README.md
@@ -128,7 +128,8 @@ This is not a message but an inner object for other messages in this protocol. I
                 "cred_def_id": "<cred_def_id>",
                 ...
             },
-            "value": "<value>"
+            "value": "<value>",
+            "non_revoked": "<iso_8601_datetime>"
         }
     ],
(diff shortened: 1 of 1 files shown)


sklump:
Hello,

I think I should get my protoype's unit tests to pass first. I hope I can say OK in about 24 hours (16:00 EDT), but golf off for the moment thanks.

Cheers
Stephen

Sent from my HTC

----- Reply message -----

kdenhartog:
Sounds good 👍

dhh1128:
I think these are good changes and should be merged. Just noting that whichever PR is merged first (this one or #189 ), the other will have some merge conflicts that need to be resolved.

sklump:
Hello, I've got all my unit tests to pass on my prototype. I believe you should merge this as is and then we can revisit what to take out and/or add from this position of strength.

sklump:
I will rework #0036 and #0037 with Daniel's implicit attribute specification encoding in their respective preview inner message objects, then submit another PR.

PR #192, opened 2019-08-19
make base64-encoding implicit by MIME type in Aries#0036, #0037proposals

sklump:


features/0036-issue-credential/README.md
```diff
@@ -183,37 +183,45 @@ Description of fields:
...
-The main element is `attributes`. It is an array of objects, each with the following fields:
+The main element is `attributes`. It is an array of (object) attribute specifications; the subsections below outline their semantics.
+
+##### Name
+
+The mandatory `"name"` key maps to the attribute name as a string.
+
+##### Value
+
+The mandatory `"value"` key maps to the attribute (string or integer) value. If the `"mime-type"` key is present and does not signify its default, the value is a base64-encoded string.
```
dhh1128: @sklump I feel like this description is still confusing, because "is present and does not signify its default" leaves open the possibility that `mime-type` could sometimes override the base64-encoded-stringness of `value`. What I want is to say the following:

* If `mime-type` is missing, its value is `null`, and `value` is a string or integer. In other words, `value` gets interpreted the same way as any other key+value pair in JSON.
* If `mime-type` is not null, then `value` is [ALWAYS a base64-encoded string that represents a binary BLOB](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0074-didcomm-best-practices#blobs), and `mime-type` tells how to interpret the BLOB after it's been base64-decoded. `mime-type` can never change the encoding.

I can't tell from your wording whether our intentions are aligned and we're just imagining different wording for the same idea, or whether you actually intend something different.

Also, I am tagging @devin-fisher here because the use of `value` to mean a string in some cases, and a number in other cases, doesn't fit well with JSON deserialization routines for statically typed languages like Rust. Devin had suggested that we never allow this kind of free variation on the data type of a value because it causes problems in serde--but in reviewing the DIDComm Best Practices RFC (0074), I don't see this rule called out anywhere. I'm wondering if it matters. What do you think, Devin?

If we want to follow the rule, what we'd do is create 3 different "value" fields: `str_value`, `num_value`, and `blob_value`. Only 1 of the 3 would ever appear in a given element under the `attributes` array. `str_value` would always be an ordinary string; `num_value` would always be a JSON number; `blob_value` would always be the base64-encoded string that contains a blob. In that case, `mime_type` would only be valid in conjunction with `blob_value`...

sklump: I agree, `value` should always be a string. I was leaving space for libVCX which sometimes uses an integer on some level. I believe this is wrong but I can't fight all the battles. Splitting up `value` into three is, I feel, stepping off the curb into something worse.

I will try to clean up the wording on my side and commit.

PR #193, opened 2019-08-19
propose RFC: Coin Flip Protocol

dhh1128:


pstuermlinger:
What about having the caller creating two strings like `me:<random_uuid1>` and `you:<random_uuid2>` and presenting both hashes to the recorder. The latter would choose one, the caller then reveals the uuids and both would see what the recorders choice was.
This is comparable to "choose left or right hand" and should save 1 round of communication.

dhh1128:
@pstuermlinger : sorry to be slow to respond; I've been on PTO.

Yes, that's a clever enhancement. Thank you for suggesting it. I haven't pondered the security implications deeply, but at first glance it seems more efficient and equally secure. Would you like to raise a PR to update the RFC and be named as a co-author?

pstuermlinger:
I do! :) Working on it. Thx @dhh1128

PR #195, opened 2019-08-21
Add details of how to sign commits to pass DCO

kdenhartog:
Signed-off-by: Kyle Den Hartog <kyle.denhartog@mattr.global>


swcurran:
@kdenhartog - is there a link you used to get this?  It seems that we should link to that vs. adding this text to many contrib documents.

PR #200, opened 2019-08-22
tweak: update verbiage slightly

dhh1128:


swcurran:
@dhh1128 - I've approved and merge this PR as it is a minor change to something that was previously merged.

What I don't understand is why the protocol exists as a standard/core protocol vs. as a vendor specific protocol.  Specifically, is it important?  I don't see anything in the text that alludes to its importance other than a vague reference to fairness that might be needed in some unstated use case.

I'm guessing there is more to it this than meets the eye, so I'm interested. What is it?

dhh1128:
@swcurran I wrote it because I was planning to use it for arbitrating delta reordering in the synchronization of peer DID docs. I suspect it will have use in many scenarios in the future; that is just the first one to trigger the need for the doc. AFAIK, no vendor (including Evernym) is privately implementing this, although Evernym did play around with a similar protocol when it was experimenting with how to model subprotocols for testing purposes. (We were using it to decide who goes first in a game played over a DIDComm protocol.)

BTW, I may not share exactly your same assumption that the Aries RFCs repo should be dedicated to things that are called "core". I do think that things that are clearly vendor-specific don't belong there--but I think anything that is generally useful and likely to solve problems for many people feels like a reasonable candidate for a community RFC, even if it's not immediately manifest in the use cases the community has currently prioritized. In other words, I think of the repo as being "community" but not crisply "core." Is that aligned enough with your thinking, or would you prefer to narrow the scope of the repo to a concept of "core" that we formally define? Should we discuss on the next Aries maintainers call?

swcurran:
I would suggest adding a more concrete use case to the RFC so that it's clear that it has a use beyond the superficial title.  I expected there was more to it than I was reading :-).

Maybe I'm recalling the protocol namespacing discussion as where we decided to separate out core from non-core. I don't think we ever discussed it in terms of (back then) HIPEs/RFCs.  I am leery of putting any more than necessary in the RFC repo - it's already pretty hard to navigate (although getting better with the recent index generator updates).  As I've always heard with fields in an issue tracking system - every one needs to earn the right to be there, or it becomes impossible to manage.

I don't think we need to go over this on an Aries call.  Things will evolve as they will evolve :-).

PR #209, opened 2019-09-02
propose RFC: credential fraud threat model

dhh1128:


dhh1128:
This PR had an error (accidentally included change to decorator RFC status). It is now superseded by #210

PR #211, opened 2019-09-04
NEW Proposed RFC: Routecoordination

TelegramSam:


features/0211-route-coordination/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Keylist Update
+Used to notify the _router_ of keys in use by the _recipient_.
+```jsonc
+{
+    "@id": "123456781",
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routecoordination/1.0/keylist_update",
+    "updates":[
+        {
+            "recipient_key": "",
```
llorllale: I do not understand why the Routing Agent needs to be aware of the recipient's keys. The RA should identify the recipient of the encrypted msg using the `to` field of the `Forward` msg: https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0094-cross-domain-messaging#required-end-to-end-encryption-of-the-agent-message. Furthermore, the `Forward` message does not include these keys.

I would presume the DID referenced in `to` would be shared by the recipient with the RA and the Sender...

llorllale:
Are we making a distinction between routers and [message queues](https://github.com/hyperledger/aries-rfcs/tree/master/features/0213-transfer-policy)?

swcurran:
> Are we making a distinction between routers and [message queues](https://github.com/hyperledger/aries-rfcs/tree/master/features/0213-transfer-policy)?

I think so.  A message queue is needed when when an edge agent does not have an addressable endpoint, and so must use return routing - e.g. a mobile agent. In that case it needs a routing agent to queue things for it.  However, routers can be used as a mediator for any agent, regardless of whether or not it can be addressed/sent a message or cannot be addressed and only uses return routing.

kdenhartog:
@TelegramSam can you resolve the conflicts and then I'll merge this? Then we can move discussion points raised by @llorllale in issues as well as @swcurran mentioned above.

PR #219, opened 2019-09-11
(#99) Add support for RFCs 0214 and 0217

### LINKED ISSUE #99: RFC 0028 - Enhance `request` message

llorllale:
Closes #99

### DIFF
--- a/features/0028-introduce/README.md
+++ b/features/0028-introduce/README.md
@@ -243,12 +243,22 @@ to Bob. No `proposal` message needs to be sent to CarolCorp;
 this is the `skip proposal` event shown in the introducer's
 state machine.
 
-#### Proposal initiated by introducee
+#### Introducee requests introduction
 
-[TODO: Alice is still doing the intro, but Bob now asks Alice to introduce
-him to Carol. Is this the same proposal message type, just with a field
-indicating that he's proposing to Alice that *she* do the intro? Or is it
-a different message?]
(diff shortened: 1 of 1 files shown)


llorllale:
@kdenhartog fixed

PR #220, opened 2019-09-11
make test suite more prominent and enforced

dhh1128:


dhh1128:
BTW, I am deliberately letting a test fail as a signal. If this is merged, we have several ACCEPTED RFCs that will need to be updated with links to test results. That is the test that's failing. Once we decide whether to merge it or not, I can go fix the test.

dhh1128:
We discussed this briefly during @dbluhm 's IIW session on the Aries protocol test suite. Everyone in the room said we should move ahead. Therefore, I am inclined to merge the PR as soon as I can grandfather in an exception for previously accepted stuff.

jljordan42:
I was not able to make that discussion at IIW. I want to better understand the test suites and specifically the test harness implementations, how they will be developed and maintained, etc. Especially if this means that passing test suites is a blocker to any agent implementation ... the test suite and implementation will need to be very current and rigorously maintained. Who is putting forward the resources to ensure that happens?

dhh1128:
@jljordan42 That's a good concern. Let me see if I can lay some of it to rest.

The test suite isn't required for any implementation work. Rather, it's required to document interoperability claims, and to gate the evolution of RFCs. An RFC that doesn't have any tests in the suite isn't qualified to advance to ACCEPTED status. An agent that can't display results from the test suite can't make any community-endorsed claim about interop.

This doesn't slow down agent development at all. Anybody can build an agent however they like, and choose to run or not run the agent test suite against it as much or as little as they like. They can release their agent without ever using the test suite, if that suits them.

However, what we hope would happen is that agent developers would run their agent against the suite from time to time. If they believe they're supporting a protocol and want to claim that publicly, they should make that claim by publishing the test suite results for their agent, for that protocol. If something doesn't pass, perhaps they have an interop gap to investigate. If something doesn't pass but should, perhaps the test suite or the RFC is flawed, and we hope the agent builders would raise a ticket and/or contribute something that fixes the problem. Once the agent passes the tests for a particular protocol/RFC to the extent that its builders intend, we hope the agent builders would submit a PR against the RFCs repo, putting their agent in the "Implementation" table for the RFC with a link to their test results.

All of this is optional, however. The agent can be released no matter what its profile is with respect to the test suite; people can update the Implementation tables in RFCs at their convenience or not at all; people can run the test suite against their agents at their convenience or not at all. The only mandatory thing is that community-endorsed claims of interop derive from test suite results.

dhh1128:
I have grandfathered in exceptions for stuff that doesn't meet this standard today. Let's discuss whether to merge this on the next Wed call.

features/0023-did-exchange/README.md
```diff
@@ -434,4 +434,4 @@ The following lists the implementations (if any) of this RFC. Please do a pull r
 
 Name / Link | Implementation Notes
 --- | ---
- |  |
+[Aries Protocol Test Suite](https://github.com/hyperledger/aries-protocol-test-suite) | [test_manual.py](https://github.com/hyperledger/aries-protocol-test-suite/blob/master/protocol_tests/connection/test_manual.py)
```
llorllale: The protocol test suite currently tests the `connection` protocol, not the `did-exchange` one. Also, it seems to be testing an older version that doesn't use the signature decorator.

dhh1128: Yes, I believe that has changed since I raised this PR; @dbluhm updated it for IIW. So should we take this impl line out?

llorllale:
@dhh1128 

> I have grandfathered in exceptions for stuff that doesn't meet this standard today. Let's discuss whether to merge this on the next Wed call.

We should set up a plan of action for these.

features/0023-did-exchange/README.md
llorllale: For sure it isn't testing for `did-exchange`.

But I was wrong about it not testing for the signature decorator - it, in fact, is: 

https://github.com/hyperledger/aries-protocol-test-suite/blob/15036428d41de75fa55e8b09b085bd25aa01a280/protocol_tests/connection/test_manual.py#L55-L63

llorllale: I suggest we remove this line from did-exchange and move it over to the `connection` RFC.

kdenhartog:
The way this reads, I think is good. As for @llorllale point about creating a plan to handle grandfathered exceptions would be good. For example, I wouldn't mind keeping an RFC as accepted and removing the currently added ones and having them go through the test suite to succeed.

llorllale:
To complicate things a bit more: I'm not sure if some of the ACCEPTED RFCs today are testable. For example, I believe the `issue-credential` protocol is currently un-testable in a network-agnostic way.

dhh1128:
Okay, I've updated this PR to conform to the discussion on today's Aries call:

* Hard-coded, grandfathered exceptions are removed.
* There is now a new tag, "test-anomaly", that allows an RFC to be uglified and marked as not conforming to our policy about tests being written.
* IFF this tag is present on an RFC, then in the Implementations table, you can turn an error into a warning in our integrity checks by appending '[MISSING test results](/tags.md#test-anomaly)` to the Notes column.

I also changed all the `Tags` fields on all RFCs to be hyperlinks instead of plain text, so it's easier to discover what a tag means.

features/0023-did-exchange/README.md
llorllale: @dhh1128 you missed this one

swcurran:
Reviewed this. At minimum, the DID-Exchange readme needs to be updated to indicate that there are no implementations - currently says the Test Suite has one.  @llorllale noted that above.

Are the tags added/managed manually or through the scripts in code?  For example, in the index.md, the "test-anomaly" tag is there when that tag in the corresponding RFC readme header, and only goes away when removed manually, presumably when all implementations remove that flag and have links to results.  Is that correct?  Or are tags generated into the README?

dhh1128:
Tags are added manually.

swcurran:
@TelegramSam, @dhh1128  - this was merged without the requested change to the DID-Exchange readme.  Does the Test Suite support the DID-Exchange protocol or do we need another PR to remove that?

dhh1128:
PR #266 fixes that detail, @swcurran .

PR #232, opened 2019-09-24
Include data.base64+data.json in subfield defs

jmandel:
These fields are documented above; seems like they should be included in the formal subfield definitions.


swcurran:
Looks correct to me, but requesting @dhh1128 to review since he is the author of the RFC.

PR #247, opened 2019-10-07
Fix broken link to Indy Cloud Agent - Python repo

gjgd:


gjgd:
@dhh1128 Good point, I've updated the link

PR #250, opened 2019-10-08
RFC 0250: Rich Schema Objects (conversion of rich-schema HIPE to an RFC) 

brentzundel:
Signed-off-by: Brent <brent.zundel@gmail.com>


brentzundel:
Perhaps the best thing to do is to move this RFC into the concepts folder, but keep the related, lower-level RFCs which are implementable (such as the one for contexts) in the features folder.

llorllale:
I agree with moving this to the concepts folder. Not too sure about the ACCEPTED status.

dhh1128:
When Indy HIPEs went through a rigorous merge process, getting a HIPE merged indicated significant buy-in. Moving such a HIPE to Aries suggests an ACCEPTED status would be appropriate. However, the community should weigh in on such statuses (as opposed to PROPOSED or DEMONSTRATED status, which are not subjective). So I think:

A) we should move this to the concepts folder
B) we should merge it immediately with the PROPOSED status
C) we should immediately raise a PR that changes the status to ACCEPTED, and then debate that PR in community circles

brentzundel:
@dhh1128 @llorllale I have moved this RFC to the concepts folder and changed the status to PROPOSED.

llorllale:
@dhh1128 I agree with the steps you propose.

concepts/0250-rich-schemas/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+The claims that make up a presentation are specified by the presentation
+definition. The credentials from which these claims originate are used to
+create new derived credentials that only contain the specified claims and
+the cryptographic material necessary for proofs. The type of claims in
+derived credentials is also specified by the presentation definition. These
+types include revealed and predicate proof claims. The presentation also
+contains the cryptographic material which supports a proof that source
+credentials are held by the same entity. This is accomplished by proving
+knowledge of a link secret.
```
troyronda: I am curious about the linking between https://w3c.github.io/vc-data-model/#zero-knowledge-proofs and this RFC?

I'm also wondering if the anonymous credential material (e.g., link secrets) should be coupled with the rich schema material?

PR #254, opened 2019-10-10
did-exchange: correlate requests to invitations

### LINKED ISSUE #208: Q: Why doesn't didexchange/1.0/request include ~thread ?
The did-exchange RFC does not include the `~thread` decorator on the [`Exchange Request`](https://github.com/hyperledger/aries-rfcs/tree/master/features/0023-did-exchange#1-exchange-request). I think it should if it is a response to an `invitation`. `~thread` is the standard way to correlate different messages exchanged in a given *context*. 

Having this odd quirk in the did-exchange protocol then requires this particular state machine to have extra reliance on cryptographic primitives in our architecture. (my assumption here is that the RFC implies the agent should use the new keys included in the invitation for correlation)

Or is there a good reason why `~thread` is not included?

### LINKED ISSUE #242: Correlate did-exchange request to invitation?
@dbluhm @TelegramSam @swcurran @dhh1128 

I wonder if implementations can benefit from having the `request` message include a reference to the `invitation` that it is responding to. This would simplify things a bit I think.

Correlating incoming `request`s to outgoing `invitation`s is useful for several things: logging, checking expiry, counting how many times the invitation was used, etc.

This correlation can currently be done by first having indexed the invitation with the `recipientKeys` included, of which there may be multiple. When receiving the `request` afterwards, the crypto layer would return the keys used to pack this message and the application layer would have to perform a reverse lookup to find the corresponding invitation.

I think an alternative to make this process straightforward is to simply include a reference to the invitation in the `request` message. The `invitation` already has the necessary characteristics to make this possible: it has an `@id` and the Sender's [requirements](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0008-message-id-and-threading/README.md#message-id-requirements) for generating this ID include that it be "sufficiently unique" (UUID recommended). The `invitation` message should be indexable by its `@id` by its Sender. In the case of [implicit invitations](https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/README.md#implicit-invitation) from public DIDs, the uniqueness of the invitation's ID is derived from the DID itself (see [example](https://github.com/hyperledger/aries-rfcs/tree/master/features/0067-didcomm-diddoc-conventions#service-conventions) in our diddoc conventions RFC).

Having the `request` message include an attribute like `invID` that would reference the invitation's ID would make the correlation more straightforward.

WDYT?

llorllale:
Closes #208 
Closes #242

I also took the liberty of updating the section that references the signature-decorator RFC.

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -107,7 +107,11 @@ The *invitee* sends the *inviter* an ack or any other message that confirms the 
 An invitation to exchange may be transferred using any method that can reliably transmit text. The result
 must be the essential data necessary to initiate an [Exchange Request](#1-exchange-request) message. A exchange
 invitation is an agent message with agent plaintext format, but is an **out-of-band communication** and therefore
-not communicated using wire level encoding or encryption. The necessary data that an invitation to exchange must result in is:
+not communicated using wire level encoding or encryption. It is also contained within its own
+[thread](../../concepts/0008-message-id-and-threading/README.md#threaded-messages) and all subsequent messages are
+exchanged in new threads spawned from this one.
+
+The necessary data that an invitation to exchange must result in is:
 
 * suggested label
 * publicly resolvable did
@@ -125,7 +129,7 @@ When considering routing and options for invitations, keep in mind that the more
 
 The _inviter_ will either use an existing invitation DID, or provision a new one according to the did method spec. They will then create the invitation message in one of the following forms.
 
-Invitation Message with Public Invitation DID:
+### Invitation Message with Public Invitation DID
 ```json
 {
     "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0/invitation",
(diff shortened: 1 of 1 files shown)


llorllale:
@dhh1128 maybe it would help to list good reasons why it should be a SHOULD?

Note that I left wiggle room on the *Inviter*'s role:

> The _inviter_ MAY look up the corresponding invitation identified in the request's `~thread.pthid` to determine whether it should accept this exchange request.

This PR further moves us in the direction of correlating based on the message IDs as opposed to keys. Not encrypting invitations allows the reuse of those keys for other invitations.

TelegramSam:
In the other thread I missed the placement of the DID URL as parent thread when I agreed. I'm not opposed but wanted to verify that this was within acceptable behavior for parent thread references.

llorllale:
@dhh1128 ^

dhh1128:
I merged this because the main changes were discussed on github issue threads, and it seemed that consensus was emerging there. Also, in the comment stream here, the only pushback came from me and Sam. My concerns were resolved, and I think Sam's were, too. If I merged in error, then we can undo.

PR #266, opened 2019-10-22
Remove Aries Test Suite as impl of 0023

dhh1128:


dhh1128:
Follow-up to PR #220 ; fixes RFC 0023's impl section

PR #267, opened 2019-10-23
Allow signed attachments (similar to old ~sig decorator)

dhh1128:


dhh1128:
We should not merge this until we resolve an important question from @sklump, which is how the verkey associated with the JWS is communicated. As he said:

>The crypto `verify()` interface requires a verkey and a wallet with the corresponding private key. JWS format does not have a verification key available to extract from the signature or header itself. It's apparently all baked into the glorious signature expansion.

The closest answer in JWS is the `kid` field in the header (see 4.1 in https://openid.net/specs/draft-jones-json-web-signature-04.html). This could be a DID key reference that gets you to a verkey. However, as I've pointed out elsewhere, that's problematic for a two reasons.

Reason 1 is temporal mismatch: the key as defined in the diddoc at the time of reception may not be the same key value that had that ID at the time of sending. This could be solved by making the DID key reference include a version and/or timestamp.

Reason 2 is that crypto shouldn't depend on DID resolution. I don't see any way around this. It's just plain wrong to make crypto (primitive math) depend on something as complex, network-dependent, time-consuming, and high-level as DID resolution.

Possible answer A: put the value of the verkey into `kid`. This would work, but I've been told by several JW* aficionados that this feels like a code smell, and it also eliminates another use of `kid`, which is binding the specific key value to a larger context. It's one thing to know/confirm that data was signed by key X, and another (related but distinct) thing to know that the key in question had a particular meaning to its owner. The latter is the intent of `kid`.

Possible answer B: add a new field into the JWS. This works and is compatible with the JWS standard in theory, but diverges from standard impls of JWS.

@kdenhartog , what do you think?

dhh1128:
^^ @troyronda @andrewwhitehead @OR13

dhh1128:
Does anybody know the github handle for Mike Jones or Kim Cameron? Is this is an example of the type of JW* misfit that should cause us to update the standards rather than trying to work around a problem?

troyronda:
@dhh1128 To answer the question posed about JWS, we need to understand the expectations for choosing the signing key being used.

The simpler situation would be that the signing key is always contained within a DID document related to the previously performed DID Exchange for these participants. In this case, is there a single key that will always be used or the signer can choose any of the keys in their DID document?

The more complex situation would be that the key can be unrelated to the DID Exchange. In this case, we would need to understand how the Agent decided to trust the other keys.

At first glance this leads to three possible outcomes:
1. No ambiguity (key can be determined from DID exchange context): I assume no kid field is needed.
2. Ambiguity due to allowance for choosing a key from the associated DID document: I assume kid field is _derived_ from a DID Key ID reference contained within the associated DID.
3. Key comes from any DID. In this case, I think we need more explanation.

Crypto and DID resolution:
In the first two outcomes, we could argue that an implementation doesn't have to perform DID resolution on every usage. Instead a mechanism could exist that allows the implementation to perform lookups against resolved keys based on a key identifier and the current context. This mechanism should be generic and decouple Crypto and DID resolution (i.e, "Key IDs" used by the crypto layer have no special "DIDness" to them).

Temporal Mismatch:
I'm not sure I see this issue as being particularly specific to JWS. The simpler situation is that signatures are treated as being dynamically generated as part of a messaging protocol. If you are unlucky that a key rollover happens during your message, this could mean usage of retries. The more complex situation is to identify the keys based on version or time using some defined rules.

dhh1128:
@troyronda :

> The simpler situation would be that the signing key is always contained within a DID document related to the previously performed DID Exchange for these participants.

We can't depend on this. We need to support ephemeral mode, where a party who has not exchanged DIDs in the past does the signing. We also need to support the case where Alice sends a message to Bob, but an attachment to the message was signed by Carol. And we need to give up the assumption that the party signing something is *already* trusted. A digital signature could be done with a key that is *not yet* trusted, but that will become trusted in the future. There is no logical reason (other than IMO bad software design) why the key must already be known and trusted by the time the message is sent.

> we could argue that an implementation doesn't have to perform DID resolution on every usage. Instead a mechanism could exist that allows the implementation to perform lookups against resolved keys based on a key identifier and the current context. This mechanism should be generic and decouple Crypto and DID resolution (i.e, "Key IDs" used by the crypto layer have no special "DIDness" to them).

That's not a good enough decoupling. Caching things eliminates an immediate dependency, but not the logical complexity. Crypto is to DIDComm what IP routing is to the internet. IP routing doesn't deal with domain names by saying that packets containing domain names as their destination can use cached DNS lookup; it says that the only thing that can be routed is a raw IP address. DNS continues to exist and be relevant, but at a layer above IP routing. DID resolution should exist and be relevant, but at a layer above crypto. Doing any form of DID lookup inside the crypto--even if it's cached--is still wrong. Math doesn't depend on resolution.

>Temporal Mismatch: I'm not sure I see this issue as being particularly specific to JWS.

You're right. It's a problem with the whole intersection between DIDs and JW*. The DID spec made a choice that I disagree with, which is that the ID of a key in a DID doc isn't immutable. Because of this, all references to keys need to be clarified by answering the __when?__ question. Huge pain.

>The simpler situation is that signatures are treated as being dynamically generated as part of a messaging protocol. If you are unlucky that a key rollover happens during your message, this could mean usage of retries.

I think you're conflating signatures that have 2 different purposes. When the purpose of a signature is to make a **message** non-repudiable, then the signature will be generated when the message is created. That is what signatures on the message envelope as a whole are used for. But this discussion is about signatures on data inside a message; these signatures make **data** non-repudiable. What distinguishes data from messages is that data has a different lifetime. It may come into being before the message, the same data item may be transmitted and retransmitted in multiple messages, and it may be persisted to disk and live a life of its own, quite independent of the message(s) that carry it. Think of a home inspection report that's part of a home-buying process. That is the sort of thing that needs a `~sig` decorator.

My point is that signatures on data are not so easily bound to messaging context, and simple retries are not going to work.

concepts/0017-attachments/README.md
```diff
@@ -154,6 +151,31 @@ following decorator:
 
 [![appended attachments](appended.png)](crime-scene.json)
 
+### Signing attachments
+
+In some cases it may be desirable to sign an attachment in addition to or instead of signing the message as a whole. Consider a home-buying protocol; the home inspection needs to be signed even when it is removed from a messaging flow.
+
+Embedded and appended attachments support this by replacing the `data.base64` field with a `sig` field that uses [JWS format](https://openid.net/specs/draft-jones-json-web-signature-04.html), where the payload of the JWS is the data that would otherwise have been base64-encoded:
```
llorllale: prefer link to the non-draft IETF RFC: https://tools.ietf.org/html/rfc7515

llorllale:
@dhh1128 the `jwk` header is there precisely for the purpose of inlining the signing key. From the [RFC](https://tools.ietf.org/html/rfc7515#section-4.1.3):

>    The "jwk" (JSON Web Key) Header Parameter is the public key that
>    corresponds to the key used to digitally sign the JWS.  This key is
>    represented as a JSON Web Key [JWK].  Use of this Header Parameter is
>    OPTIONAL.

JWK is preferred because it carries descriptive information about the key, unlike what's being done so far in Aries using the `kid` field.

dhh1128:
@llorllale Thank you for pointing it out. I had just found the same thing and was going to post about it when I saw your comment. You are right -- `jwk` exactly resolves my concern. If we use that instead of `kid`, I am happy.

dhh1128:
I have updated one paragraph of the running text to clarify that the JWS must use the `jwk` field, and that if `kid` is populated, this is just an optional hint. Do you agree with that, @troyronda and @llorllale ?

troyronda:
> We need to support ephemeral mode, where a party who has not exchanged DIDs in the past does the signing.

Ok.

> That's not a good enough decoupling. Caching things eliminates an immediate dependency, but not the logical complexity. Crypto is to DIDComm what IP routing is to the internet.

Perhaps I muddied the picture when I mentioned "caching". Agree - I also do not think a crypto layer knows about DIDs. 

So is a higher layer component determining/fetching/extracting the appropriate key(s) (based on a DID) for the given "context" before we get to the signature object validation?

> My point is that signatures on data are not so easily bound to messaging context, and simple retries are not going to work.

Ok. I guess I was thinking about the usage in a messaging protocol (e.g., within DID exchange) when I mentioned the "simpler situation".

concepts/0017-attachments/README.md
dhh1128: Updated.

dhh1128:
>So is a higher layer component determining/fetching/extracting the appropriate key(s) (based on a DID) for the given "context" before we get to the signature object validation?

I think the preceding discussion, and many other discussions I've heard in the community where DIDs and JW* intersect, may be conflating two forms of validation:

1. **Using math** to prove that the private key associated with a particular public key value truly did create a signature. (Or, relatedly, using the recipient's public key to look up the corresponding recipient private key, and then using the sender's public key value and recipient's private key value to decrypt something.)

2. **Guaranteeing** that the key used by the sender has particular **semantics** with respect to a DID (e.g., it's in the DID's DIDDoc with a certain set of privileges).

IMO it is an antipattern to think of these two as the same task, to name them with a single label, or to implement them in the same code. I'd call item 1 "cryptographic verification" (or, for decryption, "cryptographic transformation"). And I'd call item 2 "semantic verification".

Now, to get to your question, what I want is that there is no fetching of keys involved in item 1, anywhere. The crypto requires only the `jwk` field.

The semantic verfication of item 2 would involve a higher level of code that is DID-aware doing whatever tests it wants against the keys that were used by the lower-level crypto. For example, it might say, "Okay, the key used to sign checks out cryptographically; the signature's not corrupt. Now let me see if it has the correct relationship to the DID in question." And then it goes and fetches the DID Doc from a blockchain or from local cache, and tests it against its own policy about staleness, and makes a decision about whether semantic verification is adequate. What it does not do is hold up the crypto while it does this. The semantic layer is not input into the crypto layer; the crypto layer's output is input into the semantic layer.

kdenhartog:
@selfissued is Mike Jones' github handle. 

As for what's been proposed, JWKs are the format we should be using to handle both the cryptographic material as well as the key reference back to the DID. My only reservation with JWKs has been the size, but given everything is based on JSON, that's a tradeoff that's worth accepting. If in the future, we run into concerns of message size, COSE is a good option that will make for easy migration.

dhh1128:
(Part of why I want this strong separation is that I think the crypto verification is cut and dried, but the semantic verification is subjective. Example: you have a DID key reference in the form `did:foo:abc123#key1?timestamp=2019-12-03T18:06:21.123Z` (or whatever correct syntax we settle on for matrix parameters), and you want to know whether that key was a signing key for the DID at that time. Well, how precise is the timestamping mechanism? If the blockchain says the key in question was updated at 2019-12-03T18:06:21, I think it's the evaluator, not the library the evaluator is using, that has to decide if that's close enough.

dhh1128:
@kdenhartog : can you please comment more about "JWKs are the format we should be using to handle...the key reference"? I see them as a perfect way to inline a key value, but I was unaware that they had an application to key referencing. I want to know more.

llorllale:
@dhh1128 

The JWK also has a `kid` header - that would be the key reference.

I think I am OK with the optionality of the `kid` header but only for the ephemeral use cases.

dhh1128:
@llorllale Yes, I know about `kid`. So are you saying that `kid` is always populated with a DID key reference, when the DID is known to the recipient, and that's the key reference Kyle meant in the sentence I highlighted? If so, peace. But I thought Kyle was saying that any time we make a reference to a key, we should use a JWK to do it.

kdenhartog:
My thinking on this is that we would do something like this:

```
{
    "alg":"EdDSA",
    "jwk": {
    "kty" : "OKP",
    "crv" : "Ed25519",
    "x"   : "11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo",
    "d"   : "nWGxne_9WmC6hEr0kuwsxERJxWl7MmkZcDusAxyuf2A",
    "kid" : "FdFYFzERwC2uCBB46pZQi4GG85LujR8obt-KWRBICVQ"
  }                
}
```

dhh1128:
@kdenhartog , can you please clarify the context for your JSON snippet? You're embedding/inlining a JWK as the `jwk` field inside ... what?

kdenhartog:
This would be the `protected` field of the JWS. As I understand the JWS and JWK specification, this is how we would specify the cryptographic material with the kid. The protected header is the first base64url part of the compact JWS (see section 4 of the JWS spec).

sklump:
@kdenhartog
I believe "d" specifies the private (signing) key, and we want to specify the verification key, so only "x" is required here? Also, please confirm that our process for extracting a base64url (no-padding) specification for the public key from an indy-specified verkey should be to base58-decode it, then base64url-encode with no padding (as opposed to base64-encoding the base58 encoding)?

The sooner we can nail this down so nobody can mess it up, the sooner my aca-py implementation won't.

kdenhartog:
You're probably correct on this. I was trying to clear my backlog of messages yesterday and was lazy where I just copied the example from the did spec discussion on key formats. Let's follow what is specified in [IETF RFC 8037](https://tools.ietf.org/html/rfc8037#appendix-A.2) as described. I've learned the lesson the hard way that diverging from specs is hardly ever helpful and if it really is, an errata should be made.

To your second question. I think we should be base58-decode the key then base64url-encode the byte array with no padding.

sklump:
Good, thanks.

Note that [RFC 7515](https://tools.ietf.org/html/rfc7515#section-7) defines compact serialization and JSON serialization. Compact serialization does not allow for multiple signatures, but JSON serialization is not URL-safe.

concepts/0017-attachments/README.md
```diff
@@ -432,7 +456,9 @@ attachment. Contains the following subfields:
   * `links`: A list of zero or more locations at which the content may be fetched.
   
   * `base64`: Base64-encoded data, when representing arbitrary content inline instead
-  of via `links`. Optional. 
+  of via `links`. Optional.
+  
+  * `~sig`: Signed attachment. See [above](#signing-attachments). When present, `base64` is not used.
```
andrewwhitehead: Remove tilde on `sig`?

kdenhartog:
On the call today, it was discussed that we deprecate the old form of the sig decorator which is described in RFC 0264. To replace this functionality, we should be relying on RFC 0066 (using standard JWS), to handle the signing of a full message through the use of the pack() and unpack() API.

Additionally, to handle the signing of a specific piece of data, we'll move to add signing capabilities to the ~attach decorator which would use standard JWS. (This work)

Implementors of this new method should support both JWS Compact serialization and JWS JSON serialization. Compact serialization would be used for signing with a single signer. JSON serialization would be used if multiple signers are signing the message. As @sklump pointed out above, it's not very useful to support flattened JSON serialization.

Is there anyone who would like to help deprecate RFC 0264 (sig decorators) and update RFC 0066 (non-repudiable signatures)? I can take on handling of the deprecating 0264. Having help crispening the language in 0066 would be useful as well. We'll also need to find the instances where decorators are being used and update the functionality to be compatible with this work.

kdenhartog:
Of note, I'm trying to go back and complete all of my prior commitments. I read through the last comment and see that RFC 0264 should be deprecated, but it appears we're still using something similar in DID-exchange. @TelegramSam or @swcurran can you refresh my memory on the action that needs to be taken on this based on the updates in did exchange?

PR #269, opened 2019-10-24
Added RFC 0268

Brownie79:
RFC 0268 covers common practices that mobiles agents should follow when exchanging didcomm messages to better provide interoperability between mobile agents.


swcurran:
@Brownie79 - looks like you are missing an "implementations" table at the bottom of the RFC - that's causing the failure of the CI/CD processing.  It would be nice if the error message was more explicit about that.

Nice work on this - looking forward to getting input from the community and getting a solution for this implemented.

Brownie79:
@swcurran Thanks! I was trying to figure out what was missing. CI/CD should be good now.

swcurran:
Merging...despite some misspellings of "Aries" :-).  Those need to be fixed.

Looking forward to hearing from other Wallet/Mobile Agent makers on this one.  Does this give us what we need?

@tmarkovski @tplooker @smithbk @mitfik

PR #272, opened 2019-10-26
shifted language around end-users.

Identitywoman:


dhh1128:
@Identitywoman : Thanks for this improvement. We should merge it quickly. However, it fails our [DCO check](https://developercertificate.org/) because it consisted of an unsigned commit. I'm sorry for that formalism getting in the way of simple stuff, but it is the nature of the beast when lawyers get careful. :-) Two options to fix:

1. You can follow the instructions [here](https://github.com/hyperledger/aries-rfcs/pull/272/checks?check_run_id=276065714)

2. You can authorize me or Kyle or someone, here in this comment stream, to update the commit on your behalf, adding your DCO attestation. Since this creates a written record, it provides the necessary documentation without forcing you to jump through quite so many hoops.

Just say the word, and we'll merge this under option 2. Or feel free to update under option 1.

dhh1128:
PR #301 is a cherry-pick of @Identitywoman 's commit from this PR (d7679f1), minus all the cruft that was bogging this down. I modified the commit message to include Kaliya's DCO signoff as listed in the comment stream, except that I had to use the same email address that is used for the author's identity (the one in the github account).

Since PR 301 has been merged, this PR is now redundant and can be closed.

PR #275, opened 2019-10-30
Add optional fields to the 'Propose Credential' message

smithbk:
Add the schema_name and schema_version fields as optional to allow a credential to be requested based on the schema name and/or version.


sklump:
This might as well include issuer_did while we're going there

kdenhartog:
+1 to checking how we might version this. What's the impact for implementers. @tmarkovski @johncallahan @troyronda I'm curious if this has an impact on you at all.

tmarkovski:
No impact on Aries dotnet. I do think this adds complexity in how the holder can propose credentials. Versioning won’t be necessary.

sklump:
I agree with Mr. Markovski, since it does not invalidate existing credential proposals.

troyronda:
We are just about to start implementation of issue-credential protocol, so no impact yet.

dhh1128:
Adding optional fields is a non-breaking change to a protocol; see [our protocol+semver doc](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/semver.md), which says:

>An individual message type can add new optional fields, or deprecate existing fields, with only a change to its protocol's minor version. Similarly, a protocol can add new message types (or adopted ones) with only a change to the minor version. It can announce deprecated fields. It can add additional semantics around optional decorators. These are all backwards-compatible changes, also requiring only a minor version update.

We SHOULD be bumping the minor version num with this change; adding fields without even bumping minor version would be a mistake.

swcurran:
Could you remind me as well how we handle a minor bump in the version in the handling of the RFC?  Do we create a new document (0xxx) or do we keep the same document.

Would this work for updates?

- We put in a minor versions changlog section in the RFC, and include in the changelog a link to the commit of the previous minor version.
- Major versions are entirely new documents.
-

dhh1128:
Yes, I agree with that formalism, @swcurran .

kdenhartog:
+1 to the suggestion about managing minor changes @swcurran

swcurran:
Questions for all on this, so we can provide guidance to @smithbk on how to revise this: 

- Should all of the messages be updated or just the ones that are updated?
   - My thought is just the changed ones.
- What is the next minor version?
   - My thought is that it is 1.01
- Where should the changelog go?
   - My thought is that it should be called "Version Change Log" and go before the "Summary" (right after the header section.

Thoughts?

dhh1128:
I agree with all 3 of @swcurran 's suggestions.

kdenhartog:
I also agree with all 3 suggestions @swcurran

sklump:
I hate to swan in late here, but we still ought to add schema_issuer_did to the propose-credential message criteria; i.e., I propose you grant me a credential with on schema 'abc' as 'this:did' originated it.

dhh1128:
@smithbk : Are you able to make the minor tweaks described above in the comment stream? This PR proposes a good change; I'd like it not to languish...

smithbk911:
@dhh1128 Those changes should be made in the Oct 30 commit.  Did I miss something?

dhh1128:
@smithbk : Sorry. I saw that you had added the fields but didn't realize you had bumped the version number. I might have just read a bit too fast. Now that I reviewed again, I see that the version change was also made, so I should have merged this long ago. Ah well--better late than never! :-)

PR #276, opened 2019-10-30
RFC 0160: Rename DID/DIDDoc properties to match implementations

andrewwhitehead:
- Also fixes minor JSON formatting issues (enables comments in JSON blocks)
- Use the term 'DID document' consistently in verbiage instead of 'DID Doc' or 'DID Document'


features/0160-connection-protocol/README.md
```diff
@@ -375,15 +375,15 @@ The signature data must be used to verify against the invitation's `recipientKey
...
+- The `connection` attribute contains the `DID` and `DIDDoc` attributes to enable simpler signing.
+- The `DID` attribute is a required string value and denotes DID in use by the _inviter_. Note that this may not be the same DID used in the invitation.
+- The `DIDDoc` attribute contains the associated DID Doc. If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `DIDDoc` attribute is optional.
 
 In addition to a new DID, the associated DID Doc might contain a new endpoint. This new DID and endpoint are to be used going forward in the connection.
 
 #### Response Transmission
 
-The message should be packaged in the wire level format, using the keys from the request, and the new keys presented in the internal did doc. 
+The message should be packaged in the wire level format, using the keys from the request, and the new keys presented in the internal DID Document.
```
peacekeeper: Hey, just a quick note that in the DID core spec we recently decided to change the capitalization of all key terms to lower case (e.g. "DID document" instead of "DID Document"), see https://github.com/w3c/did-core/pull/73/
```suggestion
The message should be packaged in the wire level format, using the keys from the request, and the new keys presented in the internal DID document.
```

andrewwhitehead: Updating those references might be a bit out of scope for this PR but I went ahead and changed them (including a bunch of references to 'DID Doc'). The same would likely need to be done for the DID-Exchange RFC, as this is mainly a maintenance PR until that becomes the standard.

kdenhartog:
On line 141, we're using `did` in the invitation schema. Should we make this `DID` as well? Also, it doesn't look like this code path is tested in the test suite. Can someone double check the test suite as well to make sure I'm not misreading anything? @dbluhm @TelegramSam

dhh1128:
@andrewwhitehead : I am ready to merge this, if you can confirm that in current impls, the invitation uses a field named "did". See @kdenhartog 's question above...

andrewwhitehead:
@dhh1128 Looking at aries-framework-dotnet it doesn't seem to support the `did` property, but aries-cloudagent-python uses the lowercase name.

dhh1128:
Okay, that means that merging this in its current form does no harm, so we will go ahead.

PR #298, opened 2019-11-07
Rename `signers` property to `signer` in signature decorator

andrewwhitehead:
Proposed because existing implementations in ACA-py and aries-framework-dotnet only support the `signer` property and the spec only defines support for a single signature (with a suggestion toward supporting multiple signatures in future). The signature decorator is also likely to be retired as the functionality is added to the `attach` decorator.  @kdenhartog @dhh1128 @swcurran


swcurran:
@esplinr @tmarkovski @tplooker @jovfer - I'd like to merge this change.  Any issues with this?  Since this decorator will be eliminated soon, only existing implementations should care about this.

tmarkovski:
No issues. Ship it.

dhh1128:
I agree with @tmarkovski . Go for it, @swcurran .

PR #302, opened 2019-11-09
Initial version of the Aries Stable RFC

swcurran:
An RFC that defines a process for defining and evolving Aries Stable versions, and to propose an initial Aries Stable v1.0.0 that is based on implementations that are in place today.

### DIFF
--- a/concepts/0302-aries-stable/README.md
+++ b/concepts/0302-aries-stable/README.md
@@ -1,0 +1,137 @@
+# 0302: Aries Stable
+
+* Authors: [Stephen Curran](mailto:swcurran@cloudcompass.ca), [John Jordan](mailto:john.jordan@gov.bc.ca)
+* Status: [PROPOSED](https://github.com/hyperledger/aries-rfcs/blob/master/README.md#proposed)
+* Since: 2019-11-09
+* Status Note: This RFC proposes an Aries Stable process and Aries Stable v1.0.0
+* Supersedes:
+* Start Date: 2018-11-06
+* Tags: concept
+
+## Summary
+
+This RFC defines the process for the community of Aries agent builders to 
+
+* enumerate a versioned set of Aries concept and feature RFCs which are collectively referred to as ‘Aries Stable Vx.y.z’
+* track Aries Stable versions.
+
+"Agent builders" are organizations or teams that are developing open source code upon which agents can be built (e.g. [aries-framework-dotnet](https://github.com/hyperledger/aries-framework-dotnet)), or deployable agents (e.g. [OSMA Mobile App](https://github.com/mattrglobal/osma)), or commercially available agents.
+
+An Aries Stable version provides a clearly defined set of RFCs for Aries agent builders to target their agent implementation when they wish it to be interoperable with other agents supporting the same Aries Stable version. The Aries Stable versioning process is intended to provide clarity and predictability for Aries agent builders and others in the broader Aries community. The process is not concerned with proposing new, or evolving existing, RFCs, nor with the development of Aries code bases.
+
+At all times, the [Reference](#reference) section of this RFC defines one or more current Aries Stable versions -- a number and set of links to specific commits of concept and features RFCs, along with a list of all previous Aries Stable versions. Several Aries Stable versions can coexist during periods when multiple major Aries Stable versions are in active use (e.g. 1.x.x and 2.x.x). Each entry in the previous versions list includes a link to the commit of this RFC associated with that Aries Stable version.
+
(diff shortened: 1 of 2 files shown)


TelegramSam:
This process does make a change to what PRs can be accepted by RFC repo maintainers. Not arguing here, just pointing that out.

llorllale:
@TelegramSam can you please clarify what the impact would be?

troyronda:
Not all the protocols in this RFC are in Accepted state.

swcurran:
That's ok. The wording is they SHOULD be, but don't have to be. It's a
separation of the protocol RFCs life cycle vs. the Aries Stable version
lifecycle.

kdenhartog:
I don't think that makes sense. Why would we consider things that are still in flux stable?

troyronda:
What is the process for continuing the lifecycle of RFCs with the comment made above?
> This process does make a change to what PRs can be accepted by RFC repo maintainers.

As mentioned in the chat, I'm uncertain where we continue working on the next iteration of DIDComm.
Do we branch aries-rfcs?
Do we create brand new RFCs?

The quickest / easiest example is updating message types to https://didcomm.org.

swcurran:
@troyronda - this RFC is not concerned with where the RFCs continue to live, just linking to commits of versions of RFCs wherever they wind up.  It would be good to get settled where the RFCs will continue to be refined and extended.

@kdenhartog - we've had a de facto Aries Stable for six months without Approved/Accepted RFCs. They should be (as stated in the RFC) but we (as one of the builders) don't require that to include an RFC into Aries Stable.

I will take a look at the ones that aren't approved yet and look to see if we can move them forward. BTW - if it is just a status change, but no change to the spec. in the RFC, there would be no need to change the Aries Stable version.

kdenhartog:
I feel it would be wise of us to set a few minimum requirements for us to make Aries Stable have the interoperability we’re after.

1. All RFCs must be of accepted status.
2. All RFCs must have an implementation within the test suite (I believe this is a requirement of accepted status)

Without this I feel like stable is going to be too hard to nail down and make sure implementations are working as expected.

swcurran:
I think that is fine when we have all RFCs accepted, and the test suite is implemented. Which is why both of those are SHOULD in the Aries Stable RFC.

We want our implementation to be interoperable with other implementations now, so we want to move forward with Aries Stable now. aries-cloudagent-python and aries-framework-dotnet have been informally interoperable for months - time to make it explicit and let others join in.  

If any builder isn't concerned with interop now - they can wait until more conditions are met. I don't think that should stop the rest of us.

swcurran:
> I'm generally cheerful about this idea, and I think this RFC is going to expose some goodness. However, I don't like the section where agents make a claim to support an Aries Stable version. The reason is that this information is redundant with the information in the Implementation Notes section of an RFC. Also, just because an agent supports _some_ of an Aries Stable version doesn't mean it has to support all of it--but as written, this RFC suggests that supporting Aries Stable at version X is an all-or-nothing proposition.

My thought is that this RFC makes the per RFC implementation tracking unnecessary, and that we should consider switching to this.

As per all or nothing - I agree that can/will happen (although hopefully not that often).  The claims of compatibility with an Aries Stable version are self attested, and the status note can add colour to the claim - mentioning some protocols not covered or other protocols also supported.

kdenhartog:
It seems counter-intuitive to me to call something "stable", but then ignore the necessary requirements we've set in place to make something stable. I completely understand your viewpoint that ACA-py and dotnet have had this stuff working for some time now. It's been valuable to have these two implementations figuring things out ahead of schedule. 

My viewpoint is focused on recognizing that Aries is larger than two implementations and we should be setting guidelines for all implementations not just the two that are pushing the boundaries. This is why I'm willing to account for a slightly slower initial stable release in favor of getting good consensus on which RFCs go in stable and what normative, important checks should be made. 

We're going to need a bit of patience to work through this, but in the longer term it's going to make it much easier for us all to get it right. I want to be crystal clear on this too. **I am not opposed to Aries Stable.** I'm actually in major favor of this suggest and believe that this is incredibly useful for implementations to be interoperable. 

The only thing I'm asking is that we rely on stable RFCs that we know meet our needs and that implementations have the ability to test against the test suite as a point of reference to know if things are working properly. Is that really too high of a barrier to set in the long term future of **ALL** Aries implementations, even if it means it takes a little longer to get our ducks in a row in the short term?

dhh1128:
I don't mind merging the RFC in this form--it's just a proposal, after all. However, I won't vote to move it to ACCEPTED status with the mechanism of declaring conformance as it currently stands. I like the declaring of conformance to individual RFCs, and dislike the idea of declaring conformance to Aries Stable, and will vote against it after it's proposed, on that basis.

dhh1128:
Is everyone okay with me merging it? We can then discuss the merits of the RFC via github issues...

kdenhartog:
Yeah I'm good to merge this now as well since it's proposed status and we can work through the details in the issues @dhh1128 @swcurran

swcurran:
Works for me.

PR #311, opened 2019-11-18
Update Readme to reflect indy.agent > aries.agent

DougAnderson444:
Update Readme to reflect indy.agent > aries.agent

### DIFF
--- a/concepts/0004-agents/README.md
+++ b/concepts/0004-agents/README.md
@@ -493,7 +493,7 @@ The following lists the implementations (if any) of this RFC. Please do a pull r
 
 Name / Link | Implementation Notes
 --- | ---
-[Indy Cloud Agent - Python](https://github.com/hyperledger/indy-agent/tree/master/python) | Reference agent implementation contributed by Sovrin Foundation and Community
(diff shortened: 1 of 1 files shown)


dhh1128:
Hey, @DougAnderson444 : thanks for this contribution. It generally looks fine, and I'd be glad to merge it. However, there are 2 minor details to take care of:

1. There's a failing unit test, caused by the fact that the name of acapy is given differently in different places. The error message is: `inconsistent variants on impl name: "Aries Cloud Agent Python", "Aries Cloud Agent - Python"` Please put back the hyphen so the name in this README is identical to the name in others.

2. There's a DCO problem. This comes from not signing your commit to attest that you agree to the repo's license applying to what you are committing (`git commit -s`). For instructions on how to fix, see https://github.com/hyperledger/aries-rfcs/pull/311/checks?check_run_id=307250091.

swcurran:
Actually, I think the right thing to do is just remove the changed line. It was actually intended to reference the old "Indy Agent" vs. Aries Cloud Agent - Python. Notice that ACA-Py is listed a couple of lines below in the file.

The correct thing is either leave the line as is, or remove it as Indy Agent should no longer be referenced.

swcurran:
Closing this without merging for the reasons given my comment above.  Removing the Indy Agent from the implementations is a call that the Sovrin Foundation should make, and if done should be across all RFCs.

Issue #309, opened 2019-11-13
RFC0028: why is the `request` message not part of any state machine?


llorllale:
@dhh1128 @TelegramSam @swcurran @tplooker 

This line regarding the [`request`](https://github.com/hyperledger/aries-rfcs/blob/master/features/0028-introduce/README.md#request) message was brought to my attention (emphasis mine):

> **This message is not part of any state machine**; it can be sent at any time, and when it is received, the recipient can choose whether or not to honor it in their own way, on their own schedule.

Why not? I believe that...

* When the Introducer receives a request, they may a) transition to "arranging", or b) transition to "abandoning".
* When the Introducee sends a request, they may a) transition to "deciding", or b) transition to a new state called "abandoning".

dhh1128:
Suppose a guy walks up to you on the street, flips open his overcoat to reveal a bunch of dangling watches, and says, "Hey, Mister, want to buy a watch?" He looks shady, so you ignore him and don't even respond. Have you abandoned the watch buying protocol, or have you refused to engage in the protocol in the first place? Do you need a state machine to ignore something, or only once you pay attention?

When I wrote the RFC, I was defining "abandon" to mean that an introduction is underway, but a party decides to disengage without completing the workflow. Under that view, it doesn't make sense to equate ignoring the request with abandoning an introduction.

However, we could decide to answer my watch-buying question by saying, "At least one party--the watch salesman--thinks a protocol has begun, so interpreting the 'ignore' decision as abandonment is reasonable from his perspective." This would allow us to accept the suggestion in your first bullet point.

I think your second bullet point still doesn't work, however. The reason is that "deciding" means to decide about a proposal, and no proposal has been sent. The introducee who sends a request therefore can't transition to "deciding."

llorllale:
> I think your second bullet point still doesn't work, however. The reason is that "deciding" means to decide about a proposal, and no proposal has been sent. The introducee who sends a request therefore can't transition to "deciding."

Right, let me amend that:

When the introducee sends a `request`, they enter a `requesting` state. When they receive a proposal, they enter the existing `deciding` state:

[...]

PR #313, opened 2019-11-18
RFC0028: add 'request' to state matrix

llorllale:
@dhh1128 

As per #309, added `request` to state matrix, with new state `requested` for the introducee.


llorllale:
@dhh1128 

Please clarify what you mean by "optional". Is it in the sense that there are two entrypoints to the introducee's state machine, or is it in the sense that implementations are not required to support it?

dhh1128:
I mean it in the sense that there are two entry points. Everybody has to support both, but a given instance of the protocol flow doesn't have to start with a request.

llorllale:
agreed

Issue #306, opened 2019-11-12
Simplify the "Ack" feature protocol and move to Accepted


swcurran:
I think the simple "Ack" protocol should be moved to Accepted as that message is in some use. However in looking at the RFC, there are some more complex features in the RFC that I don't think would be considered "Accepted".

I recommend that we remove the "~please-ack" decorator and the other advanced features from the RFC, leaving just the ACK message.  The "~please-ack" decorator could go into a separate RFC if there is interest, possibly with the other related advanced features. There is not really a reason that the Ack message and the ~please-ack decorator need to be in the same RFC vs. being in separate ones with the ~please-ack decorator referencing the Ack message.

Does anyone object to that approach?

Has anyone implemented the "Ack" message?  The ~please-ack decorator?

llorllale:
aries-framework-go implements the did-exchange protocol which itself adopts the `ack` message to notify the inviter upon successful completion of the did-exchange.

RFC0015 is not a protocol since it lacks core [ingredients](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/README.md#ingredients) of a protocol (roles and states). Both the messages it defines are ones up for adoption by other protocols - I'm temporarily calling these "orphaned messages".

I agree with removing `~please-ack` into a separate RFC because, as a driver of a protocol's state, the use of `ack` should be strictly defined by a protocol. I agree that `~please-ack` introduces needless complexity and ambiguity.

I also agree with moving this RFC (with just `ack`) into ACCEPTED status. This RFC by its own can't be tested with the protocol suite.

swcurran:
This issue has been addressed with the merging of #316. The 0015-acks RFC now covers only the ACK message, and a new RFC has been added for ~please_ack.

Closing.

PR #316, opened 2019-11-20
Per Issue #306, simplifies this RFC and moves the ~please_ack decorator to a new RFC

swcurran:


swcurran:
@dhh1128 - as you'll see, I removed ~please_ack and did not create a decorator RFC for it. At this time, I'm not aware of any implementations that have that.  Even if we add it back here, or in a separate RFC, I don't think the advanced features are necessary at this time. As noted in the second commit, I removed links from other RFCs to the ~please_ack section of this RFC and added a TODO to link to the new RFC when it is added.

dhh1128:
@swcurran : I'm not strongly opposed to this change, but I'm not particularly in favor of it, either. Is there a reason why you feel like it is worth doing?

dhh1128:
Okay, I went and read issue #306.

I don't mind removing this content from the main RFC, but I don't want to just throw away the thinking about ~please_ack. That is substantial thinking, and just deleting it to start over doesn't seem appropriate.

I'd say, if you want to make the simplifying change, please create an RFC for ~please_ack and move the other content there.

swcurran:
Fortunately, we've used a version control system, so we can still get to
the content. :-)

I'm happy to create an RFC for ~please_ack, but it will be a very simple
version, and will not have anything about the "advanced features" from the
original PR. I think while the simple version is needed, the advanced
features not compelling.  I can do a new RFC, but thought others (like you)
might want to add more.  My thought is that we do the simple ~please_ack
decorator and then a separate RFC that defines a subscription decorator
that covers the advanced features.

dhh1128:
That answer ("I'll put ~please_ack in a new RFC if I have to, but I'm not
moving the advanced thinking") doesn't feel reasonable to me. You are
forcing two issues to be combined that ought to be separated:

1. Limiting the scope of the current RFC so we can promote its core to
ACCEPTED status.
2. Deleting functionality that's been described that you apparently don't
believe in.

We can accomplish #1 without taking any position on #2, and that's what
should be done for the current issue. This requires us to shift content out
of the RFC--not wholesale delete it.

swcurran:
As noted, I deleted content from the RFC for the goal of getting the RFC into an accepted state.  That content was removed from that does not prevent anyone from putting it into a separate RFC.

On the separate issue of what to do with the content removed from the RFC, I stated above my recommendation - add an RFC for the simple ~please_ack, and take the advanced thinking and apply it to a more general subscription mechanism vs. tied directly to the ack/please_ack as I think the tie between them is not particularly strong.

swcurran:
An alternative is that we merge this version of the RFC, move it to accepted and then someone (perhaps you :-) ) can just submit a new proposed ACK RFC that is the same as the previous version.  That gives us something to point to with for Aries Stable, and nothing is deleted.

As noted, I'm fine with the approach, but would raise a question about whether the advanced features should be part of the ACK protocol.  That would be aligned with the process.

Alternatively, what I have done here goes in as a different ACK RFC vs. reusing the same 0015 numbering.

Thoughts from a process perspective?  My goal is that we have an accepted RFC that reflects what has been built so that we can reference it in Aries Stable.

swcurran:
Updated the PR to add a new RFC for `~please_ack` that moves both the decorator and the advanced concepts from the RFC 0015-ACKs.

@dhh1128 - what do you think? I did eliminate the png's :-)

dhh1128:
I am just getting off of a plane, but this sounds great, and I will try to
merge it tomorrow morning when I get up.

Issue #317, opened 2019-11-20
RFCs 0160, 0094, 0067: Differences between implementations and spec


dbluhm:
While writing tests for the connection protocol in the test suite, I made a few observations regarding DID Documents as described in the DIDDoc conventions RFC and Cross domain messaging RFC and how those differ from some of the current implementations.

First off, the connection protocol doesn't actually make specific recommendations or requirements regarding what is expected in DID Documents even though the information contained therein is crucial to the successful parsing and processing of connection requests and responses. I wonder whether we should make some sort of recommendation by linking to the DIDDoc connection or another RFC where expected values (especially services) are outlined.

My second observation is closely related to the first. The DIDDoc convention and cross domain messaging RFCs describe or show examples of DID Documents where the service denoting the agent service uses a service type of `did-communication` and a service identifier suffixed with `;did-communication`. However, I believe the Aries Cloud Agent Python and Aries Framework DotNet projects are successfully interoperating using `IndyAgent` as the service type and a service identifier suffix of `;indy`.

So my question is where should we encourage change? I think usage of `did-communication` makes more sense given that we've now migrated from Indy Agent but I didn't want to "enforce" one practice or the other without first calling attention to the discrepancies.

@swcurran @tmarkovski @TelegramSam @dhh1128

troyronda:
FYI: We have now implemented DIDDoc convention in aries-framework-go using `did-communication`.

swcurran:
My opinion is that 0160 is the connection protocol, so that it should be the one takes priority amongst the three mentioned.  That it does not align to 067 should be noted, and  we must make sure that a future version of the connection protocol (which will be DID Exchange) is aligned with 067.  I believe that is what I think Troy is saying the Go team has done.  If we do have another version of the connection protocol before going to DID Exchange (which I doubt we will do), the 067 alignment should be included in the update.

0094 is a concept RFC and so should be updated to align with 067, but concept RFCs should be of less importance from an implementation perspective than features.

dbluhm:
@swcurran the connection protocol doesn't make any specific recommendations on DIDDoc format, deliberately omitting those details, so there's nothing to align with 67.

swcurran:
Got it. Sorry.  I suggest we get whatever is being used in the implementations we have into 0160. We should do that without changing the version of the RFC. I'll get a PR in today.

dbluhm:
[...]

PR #319, opened 2019-11-21
Addresses #317 - adds example DIDDoc to current RFC

swcurran:


dhh1128:
Normally we would not merge changes into an ACCEPTED RFC without debate--but this change just provides an example to clarify what's already implemented and accepted, so I figure it doesn't need special comment. Merging.

PR #322, opened 2019-11-21
RFC 0005 DIDComm --> ACCEPTED; Deadline 2019.12.05

swcurran:


dhh1128:
Comment period has elapsed with no dissent. Conversations at connectathon confirmed that community is aligned with this.

PR #325, opened 2019-11-22
RFC 0092 Return Route --> ACCEPTED; Deadline: 2019.12.16

swcurran:


llorllale:
I have issues with the way the semantics of this decorator introduce coupling between layers of an agent's internals.

From reading through these RFCs I get the sense that all these "features" should be composed into mutually independent layers in an agent's implementation.
The [General Patterns](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0004-agents/README.md#general-patterns) section of RFC 0004 seems to imply so.
It also aligns very well with the efforts to enable plugin functionality for a lot of these components.

In particular, a protocol handler should be decoupled from the underlying transport handler. If I were to draw the layers, it would look like this:


```
|--------------------|
|     Controller     |
|--------------------|
|  Protocol Handers  |
|--------------------|
| Packers/Unpackers  |
|--------------------|
|     Transports     |
|--------------------|
```

What `~transport` effectively does is couple the `Protocol Handlers` and `Transports` layers together (or `Packers/Unpackers` <-> `Transports`). A higher layer needs to inform the transport layer to release/close the connection on which the incoming message was received if no `~transport` decorator was received.

To address this concern, I think an Agent should mark a `serviceEndpoint` as either `full-duplex` or `simplex` in its DID document. Other Agents may connect to it on whichever endpoint they wish. `full-duplex` and `simplex` would simply become different `Transport` implementations and no coupling would occur with higher layers.
Either of these two modes is enabled from the context (ie the serviceEndpoint) as opposed to being evaluated on a per-message basis.

andrewwhitehead:
@llorllale I agree that the `~transport` decorator does violate the layered approach. So does the `~timing` decorator IMO, in that it should be handled independently of the handler that produces the message.

Encoding that behaviour in the DID Document might be a bit cleaner in terms of separating those layers, but it seems like the complexity of handling duplex connections is going to be about the same. Maybe it would be useful in order to indicate such support.

Personally, I think this information would be better represented at the envelope layer, in the ciphertext but not in the message itself. Message signatures could go there as well. That would require an update to the envelope format which is definitely out of scope for the 1.0 interop profile I think, although I have some notes on that. For now, there are multiple implementations supporting this method and it's been critical to us supporting mobile clients.

kdenhartog:
I'd be curious to learn more about your thinking @andrewwhitehead I had similar ideas very early on about using the ciphertext for routing, but not transport layer. We eventually decided against it for privacy reasons, but I have little reason to believe transport data would introduce similar concerns. The devil always seems to be in the details though.

llorllale:
Adding more to my previous thoughts: this decorator seems a weak attempt to reverse the natural dynamics between a client and a server.

At the request of the client agent, this decorator turns a simplex connection into a duplex one if the server agent honors it. It follows that the server agent will honor it only *if* they are interested in serving such clients. As per the RFC (emphasis mine):

> Compliance with this indicator is optional for agents generally, but **required for agents wishing to connect with client based agents**.

It is the *server agent* that somehow needs to broadcast support for this feature, not the client.

Furthermore, implementations of synchronous and asynchronous architectures are sufficiently distinct to the point where it doesn't make much sense to serve both under the same endpoint (DIDComm messages are routed by their type as opposed to a path). A server agent allocates resources one way or the other depending on which they intend to serve.

For these reasons, it should be the server agent that should signal support for this feature, not the other way around.

andrewwhitehead:
It's useful to have the server advertise support for duplex communication, as the alternative is having a client with no public endpoint sending a message and simply getting no response (or possibly an error code if the transport supports it, and if the server explicitly rejects messages requesting return routing).

Return routing is not necessarily all-or-nothing either - it is still useful for the client to request only specific responses to be returned (by thread ID at the moment) which this decorator supports. It seems natural to assume that a server which supports duplex connections can also handle routing messages to an endpoint, if only for delivery of responses to a client which lost its connection, so we don't necessarily want all responses to be returned on the same socket just because the transport (and server) supports it.

dhh1128:
I've always felt like this RFC has the kernel of an important idea in it, but that I was always uneasy that we weren't exposing it or describing its semantics right. I haven't pushed on it because I figured experience would tell us what we needed to learn.

If I apply the semantics of ACCEPTED meaning "incubating on a standards track", then I am okay with the status change. I would not vote for ADOPTED, however.

swcurran:
@tplooker @andrewwhitehead - I think both of you have opinions on why this is not the way we should go.  I think we do need something in place, so what do we do?

@mickrau heads up on this issue as there may be a move to remove this from the RFCs list in favour of another approach.  I know that your team is using this capability.

TelegramSam:
@swcurran can you update this to be mergeable?

swcurran:
@TelegramSam I think it is now ready.  Let me know if I need to do something index.md.  Right now it's just waiting on approval.

dhh1128:
Discussed on today's community call; consensus was to merge.

PR #327, opened 2019-11-25
Add crypto service protocol RFC

mitfik:
Introduce crypto service as a common way to expose internal cryptographic
functions to the world outside of the Agent

Signed-off-by: Robert Mitwicki <robert.mitwicki@difacturo.com>


mitfik:
Following discussion during our weekly call submitting proposal for crypto service protocol which would allow to expose crypto mechanism outside of the agent. This RFC is work in progress and required community review but as we agreed such a service would be very helpful for 3rd party application.

PR #333, opened 2019-11-27
In example, Bob acks only having stored payload as per pic in RFC 36

sklump:
This tiny change is bigger on the inside: it syncs up the example to the pic, which changes processing order in aca-py. On the other hand, one might argue that the prior wording implied credential storage as per the diagram, in which case one ought to reject this change for simplicity.


dhh1128:
This change feels very reasonable to me. It captures the likely intent of the ACK. Note, however, that there is no way to enforce the requirement that the ACK be returned once storage has occurred; all we can do is document desired behavior.

PR #338, opened 2019-12-03
RFC0335: HTTP over DIDComm

Moopli:
Support transport of HTTP messages over a DIDComm channel.

Defines concepts, flow, and message format for a communication protocol between agents that carries HTTP messages, to enable non-agent clients/servers to use DIDComm for their communication, by interposing aries agents in between.


Moopli:
All feedback is welcome.

RFC Name? Formatting? More links? Flow diagrams? Diagrammed examples? More specificity with states/flow?

Baha-sk:
RFC looks good... probably some diagrams showing request/response flows between Aries agents would provider a better (visual) understanding.

troyronda:
I think this could be split into two RFCS:

- Generic JSON Relay Message
  - Enable external applications to exchange JSON messages using Agent as proxy (via DIDComm connection).
  - Allows new applications to be built without changing the Agent. Particularly useful when the Agent does not need to be aware of the external application's state (nor state machine).
- HTTP Semantics on top of the JSON Relay Message (the current PR).

troyronda:
A couple quick thoughts on a "multiplexed" JSON relay message:

- Some agents and frameworks enable an external application/service to provide the business logic (the "controller" model). 
- The JSON Relay Message extends this concept for interactions where the agent can remain stateless (beyond transient relay state). In this case, the agent simply relays for the registered application for that message.
- This approach potentially allows invocation of interoperable APIs that have been defined outside DIDComm and Aries (where the recipient agent simply relays to the external application/service). Note: the agent code does not need to be changed to support new APIs exposed by the external application/service.

troyronda:
@dhh1128 

@Moopli has created a generic version of the JSON relay at https://github.com/hyperledger/aries-rfcs/pull/351

features/0335-http-over-didcomm/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+```jsonc
+{
+  "@type": "https://didcomm.org/http-over-didcomm/1.0/request",
+  "@id": "2a0ec6db-471d-42ed-84ee-f9544db9da4b",
+  "request": {
+    "method": "",
+    "target": "",
+    "version": "",
+    "headers": {},
```
llorllale: Suggest this be made into an array. [Message headers](https://tools.ietf.org/html/rfc2616#section-4.2) are allowed to appear more than once, with a little bit of parsing logic to collapse them into a single unit. It seems wiser to just let the existing, battle-hardened http components handle this.

features/0335-http-over-didcomm/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+##### `request`
+
+```jsonc
+{
+  "@type": "https://didcomm.org/http-over-didcomm/1.0/request",
+  "@id": "2a0ec6db-471d-42ed-84ee-f9544db9da4b",
+  "request": {
+    "method": "",
+    "target": "",
```
llorllale: Suggest rename this to "resource-uri" in order to be more aligned with the [RFC](https://tools.ietf.org/html/rfc2616#section-5.1.2)

features/0335-http-over-didcomm/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+Allows HTTP traffic to be routed over a DIDComm channel, so applications built to communicate over HTTP can make use of DID-based communication.
+
+## Motivation
+
+This protocol allows a client-server system that doesn't use DIDs or DIDComm to piggyback on top of a DID-based infrastructure, gaining the benefits of DIDs by using agents as HTTP proxies.
+
+Example use case: 
+Carl wants to apply for a car loan from his friendly neighborhood used car dealer. The dealer wants a proof of his financial stability from his bank, but he doesn't want to expose the identity of his bank, and his bank doesn't want to develop a custom in-house system (using DIDs) for anonymity. HTTP over DIDComm allows Carl to introduce his car dealer to his bank, using Aries agents and protocols, while all they need to do is install a standard agent to carry arbitrary HTTP messages.
+
```
Moopli: TODO: Further use cases, and further explanation of the use cases, were requested in the 18/12/19 morning call.

swcurran:
@Moopli - are you ready for this to be merged?  There are still To Dos, as of your last comment.  If ready, right before you respond, please update the index.

swcurran:
After reading the comments on the generic JSON RFC #351 it looks like this one should be paused. I've added the label to it. Please remove and let me know if I have misread and this is ready to merge.

troyronda:
@Moopli Looks like there are a couple updates to this RFC with the changes to #351 .

features/0335-http-over-didcomm/README.md
```diff
@@ -0,0 +1,198 @@
...
+DIDComm messages for this protocol look as follows:
+
+##### `request`
+
+This message has the `data-type` value `http/request`, and uses the inline `data` attachment.
+
+```
+{
+  "@type": "https://didcomm.org/https://didcomm.org/json/1.0/message",
```
troyronda: type has a duplicate prefix.

features/0335-http-over-didcomm/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Tutorial
+
+#### Name and Version
+
+This is the HTTP over DIDComm protocol. It is uniquely identified by the URI:
+
+    "https://didcomm.org/http-over-didcomm/1.0"
+    
+TODO: as a protocol defined on top of generic json messaging, does it get its own URI?
```
troyronda: I had thought the idea was to have a single protocol, with differing data type definitions?

Moopli:
@swcurran we can un-pause this - all it needs is to verify links when RFC 0351 is merged.

swcurran:
Sure - that's totally up to you. The intent is that you unpause it when you are ready.

Done.

swcurran:
@dhh1128 - looks like this is ready for another review.

dhh1128:
I still feel like this RFC will create confusion because of the way it special-cases HTTP 1.1 as a transport. What if I only want HTTP on *one* side of the interaction instead of both? What if I want to use the same technique with a transport like HTTP 2? We're not generalizing enough in the RFC, IMO.

However, that's not a valid reason to hold up a proposal, so I will go ahead and merge and we can work the RFC in its PROPOSED state.

Issue #198, opened 2019-08-22
RFC 0029: signature verification and MTC


dbluhm:
After writing an implementation of Message Trust Contexts, I get the feeling that it may be helpful to have a context for signatures being successfully verified (SIG_VERIFY_OK or something). I think it plausible, for instance, to have a decorator pre-processor verify and unpack signatures such as `connection~sig` in the connection protocol, adding the context to the message. However, having more than one signature decorator inside of a message where one verifies and another does not muddies the waters. We would have to be able to specify which field verified and which one did not.

This might be scope creep for the concept of MTCs; but, I think we already face the same issue with the `nonrepudiable` context. If a message has a signature decorator inside of it that verifies, do we apply the `nonrepudiable` context to the whole message? Are MTCs, perhaps, recursive meaning that sections of a message can have their own MTC? Or do we need another data structure that adds more context to a specific message trust context?

cc @dhh1128

dhh1128:
The intent was that the NONREPUDIABLE attribute of an MTC would equate to what you are calling SIG_VERIFY_OK.

I agree that having a signature over a specific subset of a message's content, and having a signature over the entire message, muddies the water. But I wonder if incomplete signature verification is really useful; could we just say that either all signatures validate, or else the whole message is unworthy of the NONREPUDIABLE flag?

I'm feeling like we should shy away from extra complexity if we can get away with it, because right now code written by @dbluhm and code written by me are the only two impls I know of for MTCs, and I want to make it easy to pick up support; I think pretty good MTCs that are widely adopted would be better than highly precise, granular/recursive MTCs that feel too intimidating to back into impls.

That's just a quick gut reaction, though. I could be talked out of it.

dbluhm:
I understand the reasoning to keep MTCs simple and I think I agree.

>I agree that having a signature over a specific subset of a message's content, and having a signature over the entire message, muddies the water. But I wonder if incomplete signature verification is really useful; could we just say that either all signatures validate, or else the whole message is unworthy of the NONREPUDIABLE flag?

I'm not sure I follow here. If we did have a message with one or multiple signed subsets of content and the signatures verify, applying NONREPUDIABLE to the whole message would seem to imply that any contents outside of those subsets are also nonrepudiable which would not be true.

[...]

PR #346, opened 2019-12-11
RFC 0029: Add notes on attachments and MTCs

dbluhm:
Here's an attempt to capture the outcome of #198 
cc @dhh1128 @kdenhartog


dhh1128:
Discussed and agreed to on community call, 2019-12-11

PR #347, opened 2019-12-13
 added RFC 0347-proof-negotiation

CHempel-esatus:


swcurran:
While the details of the of the tutorial can probably be surmised from the json example, it would be helpful to include an explicit use case/story that describes the request in plain language and business terms (e.g. "a bank needs a customer to prove their residency status, citizenship and time at their current address...It offers a number of ways to prove meet these requirements, including..."). That would make it a lot easier to understand the items and values in the presentation preview json chunks.

Is the expectation that the prover and verifier could do all this with a `presentation-proposal` negotiation?  For example, the verifier sends an initial proposal that outlines all the options (e.g. the first json example), and the prover responds with a proposal that is only what they plan to provide (e.g. no options, but explicit claims and credential definitions). Once satisfied, the verifier constructs a specific proof request for the prover that matches the prover's proposal on the last negotiation iteration. The prover would have to understand while they are not giving the verifier their data, their are providing metadata about credentials they have (probably) been issued...

In other words, can we accomplish the goal of this RFC by extending the "proposal" message (or more specifically, the preview), and adding prover functionality that allows the prover to select credentials they have to satisfy the request, just as is done in the proof request/proof process, but with the specific goal of only sending back a proposal, not a proof?

swcurran:
@CHempel-esatus - are you happy this is ready to be merged?  It will still be in the proposed states, so changes are easy (more PRs), but I'm wondering if you are happy enough with this to do a merge?

If ready, you will need to update the index again right before the merge is done.

Thanks

swcurran:
@dhh1128 - could you work your magic on the index.md so that this can be merged?  Not sure why it has not been merged yet.

dhh1128:
The normal recipe for resolving merge conflicts did not work for me. I'm not sure why; it may be that the autogeneration of index.md is interfering in some way I don't understand.

Anyway, I created a new PR that contains @CHempel-esatus 's 3 commits (so it's a duplicate of this PR), but with the commits applied after latest on master, instead of earlier. I merged that PR (#375), so this one is now redundant and can be closed.

Issue #225, opened 2019-09-18
Update message type and protocol identifier URIs to https


troyronda:
With the consensus in the Aries WG call to update protocol URI to use https, our next step is proposing a concrete mechanism. The prior discussion on this topic was held in https://github.com/hyperledger/aries-rfcs/issues/129

We would like to use a Hyperledger Aries URI string such that:
 - browses to the appropriate protocol documentation.
 - protocol documentation is up-to-date with the GitHub repo hyperledger/aries-rfcs

Example of protocol type URI:
- https://aries.hyperledger.org/rfcs/didexchange/1.0/invitation

Example of redirecting to the source documentation (for the above):
- https://aries.hyperledger.org/rfcs/didexchange/1.0/ -> https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/README.md
- A nice-to-have would be including the suffix as a fragment (e.g., https://github.com/hyperledger/aries-rfcs/blob/master/features/0023-did-exchange/README.md#Introduction).
 
Some options to explore for hosting:
- GitHub Pages
- Redirect to GitHub markdown page
- Read the Docs
- Hyperledger web server

Also need to check with Hyperledger to explore registering the aries.hyperledger.org subdomain.

dhh1128:
I have opened a conversation with HL's community architects about how to approach the rewriting/hosting/redirection feature that we need. I will report back.

dhh1128:
Update: Hyperledger IT is asking us for nginx rewrite rules. I am working on it.

troyronda:
@dhh1128 Thanks very much for your efforts. Are there any updates to share for this topic?

I know it's not directly coupled, but it would be nice to progress on updating the message types before we get too far down the path of community interop claims discussed in the test suite issue.

dhh1128:
I'm stalled again. I could give HL IT people nginx rewrite rules, using either of two strategies:

1. HTTP 301 redirects from aries.hyperledger.org/something to github.com/hyperledger/aries-rfcs/something -- aries.hyperledger.org never serves content, but redirects 100% of the time.

2. aries.hyperledger.org hosts static content and uses redirects to expand short URIs into long ones.

Strategy 1 requires less work by HL IT, but causes content to be penalized in Google's PageRank.

Strategy 2 requires HL IT to allocate a web server, and it means we have to modify our repo's CI/CD to generate static content.

It is not obvious which we should do.

[...]

PR #348, opened 2019-12-13
Propose a way to transition from the message type prefix of did:sov... to https://...

swcurran:


swcurran:
Please review - @TelegramSam @dhh1128 @troyronda.  Would this work as a transition plan for this change?

I noticed in looking at #225 that we really haven't settled on the precise name of the new prefix. I assume we have to settle on that before we proceed.  From my perspective, I'm OK with a change, but I would REALLY like to see the link be real, and to the spec.

In making this transition, I'm proposing that we would NOT update the RFCs to reference the new prefix until we move to Step 3 of the transition plan - e.g. when we start to change to defaulting to the new prefix.

Sam, I did not put in the line about "In step 1 you can send out a new message type if you receive one" because that directly contradicts the guidance that you "MUST only send out the old message type". How would one ever receive a new message if no one is allowed to send one out?

Also, Sam, note the new "Step 0" note that I put in about getting any RFC ducks in a row before beginning Step 1, and the use of the status values to track the steps - "Accepted", "Adopted" and "Retired" status updates trigger the start of Steps 1, 2 and 3, respectively.

dhh1128:
I don't think we need to retain ";spec" in the new form. It would just be "https://didcomm.org/protocol/1.0". The ";spec" part was only necessary when we were planning to do DID resolution and needed an identifier "spec" for the service endpoint.

swcurran:
> I don't think we need to retain ";spec" in the new form. It would just be "https://didcomm.org/protocol/1.0". The ";spec" part was only necessary when we were planning to do DID resolution and needed an identifier "spec" for the service endpoint.

Fixed

PR #350, opened 2019-12-13
Updates to the AIP RFC and the contents of AIP 1.0.0 following the Connectathon and Aries Working Group discussion

swcurran:


swcurran:
Update links to latest non-protocol changing versions of the RFCs in AIP.  The only changes to all of these protocols are clarifications:

* 0004-agents
* 0011-decorators (now "Accepted")
* 0047-json-LD-compatibility
* 0015-acks
* 0025-didcomm-transports
* 0035-report-problems

Did **not** update the link to 0036-issue-credential that has a more recent version because it is to a new patch version of the protocol - adding some optional fields.  Given we are a month out from the expected change, should we update the link to that?

NTH a program that generates a report of what has changed since this was set would be nice. I did this checking manually, with a lot of mouse clicks... :-)

esplinr:
I appreciate @swcurran making these changes based on the discussion during the Aries Connect-a-thon. I agree with all of the changes, except the addition of the Service Decorator.

* It isn't required for credential exchange interoperability
* There are some open questions about the RFC that make it hard to implement in a generic way

We recognize that this decorator is important for supporting the VC-Auth-OIDC flow in ACA-py. We don't discourage anyone from supporting that flow, and hope to include support for that in our products in short order. But we don't think the Service Decorator RFC is mature enough for inclusion in an interoperability profile because it leaves open questions that could impact interoperable implementations.

I don't have an opinion regarding the question about the link to 0036-issue-credential.

swcurran:
So how do you want to proceed?  At least some others building to AIP 1.0.0 support it and it will be required for anything that we do, so in a practical sense, we can drop it, or the apps can just not support it, but it will just mean apps that don't support it won't be used in many of the use cases that will be initiated in the next few months. 

Comments from others that are building agents?

esplinr:
I recognize it will be required for BC.gov use cases, but it won't be required by other uses cases for which we want community interoperability (for example Evernym's use cases).

To proceed, I would like to accept the changes in this pull request excluding the Service Decorator. We can then discuss adding the Service Decorator in a future interopability profile after we have provided additional feedback on that RFC.

(My response shouldn't be interpreted to discourage comments from others who are building agents.)

swcurran:
I disagree with the exclusion of that entry, as it will discourage other agents created by other parties from being interoperable in key use cases unless they have the inside scoop about what is really need for an important "today" use case.

I also don't understand supporting ~service in the vc-authn-oidc, but not claiming to support it.

dhh1128:
>I also don't understand supporting ~service in the vc-authn-oidc use case, but not claiming to support it.

@swcurran : I think the concern from EV is that existing impls of ~service may only support one layer of routing in `routingKeys`, because most other agents are not using agencies as a mediator. If this is the case, it means an EV edge agent behind a cloud agent behind an agency (2 layers of routing keys) wouldn't be a valid destination, even though impls touting AIP 1.0 support would be claiming to implement RFC 0056. The concern would color all existing deployments of C.M; C.M wouldn't be a routable target.

This concern may not be relevant; perhaps existing impls DO support multiple layers of routing keys already.

If they don't, then pushing the service decorator out of scope properly conveys the idea that, while its support is helpful, it is not core to what the first AIP targets.

Alternatively, we might consider the following:

* Update the service decorator RFC to specify that only one level of routing is supported, and to clarify what this constraint implies about which types of agents are reachable.
* Update the definition of AIP 1.0 such that it points to this updated RFC.

Then we could start work on implementing a more robust version of the service decorator, and tie it to a different AIP version.

dhh1128:
Merging per conversation on community call. Note that the inclusion of the `~service` decorator is still controversial, so that particular item may be removed in a follow-up PR. However, this PR does remove some items that are not controversial, so it's a step in the right direction.

PR #351, opened 2019-12-16
RFC 0351: Purpose Decorator

Moopli:
This RFC defines a new decorator, ~purpose, which indicates the usage and meaning of a message in a machine-parsable way. Agents can filter behavior on the purpose decorator, including filtering messages to be sent to client applications.


swcurran:
I think there a couple of things missing from this proposal.

First, there are no actual use cases for how this would be used. Do you have any ideas on where you plan to use this?

Second. there is no detail on the purpose or use of the non-DIDComm software in the scenario.  For example, I'm wondering who is doing the decryption of the message?  If there is a non-DIDComm agent that is intended to be the recipient of the message, what does the DIDComm know about the message?  Do they decrypt the full message?  What goes into the did:peer DIDDoc?

swcurran:
Looks like the proposal is still being evolved.  Perhaps better to do that locally to get a complete first pass done?

Moopli:
The use case I'm sure about involves the HTTP over DIDComm RFC I parked and will rewrite - allowing protocols built over HTTP (eg, REST servers) to operate over DIDComm, using DIDComm as a transparent layer.

The expectation is that the sender owns the sender-agent, and the recipient owns the receiver-agent, so the encryption of the envelope between sender and sender-agent is a matter for them to decide, same for the recipient, and the payload can be encrypted by the sender for the recipient as well, but these are questions answered by different layers.

For all DIDComm purposes, the receiver-agent (perhaps I should replace this with 'recipient agent') is the true recipient of the message, the owner of the peer DID, the provider of the DID doc during DID exchange, etc.

Moopli:
Thanks for the questions by the way, I'll work the answers into the RFC. It was indeed ready for review, the changes you might have caught were from rebasing away a merge commit :)

troyronda:
> The use case I'm sure about involves the HTTP over DIDComm RFC I parked and will rewrite
(@Moopli)

@swcurran FYI: the HTTP over DIDComm RFC PR is here:
https://github.com/hyperledger/aries-rfcs/pull/338

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,114 @@
...
+This protocol defines one DIDComm message format, of type `https://didcomm.org/json/1.0/message`. A `message` only requires (in addition to the standard contents of DIDComm messages) a `data-type` field and some number of JSON attachments, under the optional fields `data`, `data~attach`, and `~attach`. 
+
+The `"data-type"` field indicates the type of data contained under the attachments.
+
+The attachment fields are all optional, though there should be at least one.
+
+```json
+{
+  "@id": "123456780",
+  "@type": "https://didcomm.org/json/1.0/message",
```
troyronda: a topic field would be useful as an additional mechanism for recipients to more flexibly register/filter incoming messages.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,114 @@
...
+  "data-type": "",
+  "data": {},
+  "data~attach": {},
+  "~attach": []
+}
+```
+
+##### `data-type`
+
+The sender must provide a `data-type` identification string constant, which indicates, for the receiver, how the data is to be processed, and indicates, for the receiver-agent, which receiver(s) to send the message to.
```
troyronda: We should recommend that the data-type value be appropriately namespaced.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Summary
+
+This protocol defines a format for sending arbitrary JSON payloads using DIDComm, for use cases where business logic or external (to agent) applications need to communicate over DIDComm.
+
+## Motivation
+
+Aries users may need to support communication that doesn't fit within a particular existing Aries protocol. In cases where machine-readable data needs to be transmitted, a defined format for sending such messages allows those users to transmit data over DIDComm, without making custom agents with custom DIDComm message formats that might collide with future, standard DIDComm message formats.
+
+Beyond allowing agents to implement custom protocols, this protocol allows clients that aren't Aries agents to communicate JSON messages over DIDComm using Aries agents as [mediators or relays](/concepts/0046-mediators-and-relays/README.md).
```
troyronda: An additional motivation is to allow a generic message development, such that an Agent codebase doesn't need to be changed for each message type.

This concept works in situations where only the "business logic" (as opposed to the Agent) needs to maintain state. In other words, this concept works for messages where the Agent only needs to temporarily maintain threading state.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+The payload transported in this message is placed in an [attachment](/concepts/0017-attachments/README.md). If the attachment is embedded or appended, the receiver is provided with the entire attachment descriptor.
+
+The receiver-agent must be able to process all three fields.
+
+#### Threading
+
+A protocol built on top of generic JSON messages can include [`~thread` decorators](/concepts/0008-message-id-and-threading/README.md) in the messages. For example, this allows the receiver-agent to relay a reply from the receiver, so the receiver is able to reply to a message.
+
+For threading with senders or receivers that aren't aries agents, the sender-agent or receiver-agent must maintain context to correlate the DIDComm message thread, and the message thread of the communication protocol with the non-aries communicator. The lifetime of this context, for a custom protocol built on top of generic-JSON-message, is defined by the custom protocol.
```
troyronda: Timing decorators could also be useful for the message lifetimes.
... but that's a generic statement :).

Moopli: I mentioned timing decorators in HTTP over DIDComm, I could move that here, make this protocol describe an optional mechanism for supporting duplex communication, and in HTTP over DIDComm refer to this mechanism for the request-response nature of the protocol.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Drawbacks
+
+## Rationale and alternatives
+
+- Why don't we standardize data-type labels against content schemas and format definitions? The intent here is to allow custom protocols and messaging schemes to operate over DIDComm. For this we provide a message format that allows them to be constructed without a chance of collision (between message names, for example) with other DIDComm message types, and we allow custom formats to be designed without needing to standardize them through the RFC process. The data-type label allows implementors to ensure that they will not collide between any of their message types in their specific implementations. A custom schema could be defined in its own RFC.
+
+## Prior art
+
+- Compare against the [basic-message](/features/0095-basic-message/README.md) protocol, which uses aries agents as relays for person-person (or machine-person) communication. In contrast, this protocol uses aries agents as relays for machine-machine communication.
+- The non-normative data type definition (as opposed to a standardized system like MIME types) is similar to Intents in Android development.
```
troyronda: You may want to more strongly highlight similarities to Intents.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+##### `purpose`
+
+The `purpose` of a message is a machine-readable string value which describes the *semantics* of a message - the role it plays within a protocol implemented using this RFC, for example.
+
+The purpose is one mechanism for determining which recipient(s) should be sent a message:
+
+The recipient registers on a particular purpose value with the recipient-agent, in which case the recipient-agent must relay any message with that value to the recipient. Multiple recipients can register with a single recipient-agent on a single purpose value, and a recipient can register with multiple recipient-agents.
```
troyronda: I think this field needs to support multiple values. We should specify the format.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+##### `data-type`
+
+The sender must provide a `data-type` identification string constant, which indicates, for the recipient, the format of the data and how it is to be processed. 
+
+##### `data`, `data~attach`, `~attach`
+
+These 3 fields are optional, although at least one must be present. They carry attachments (inline, embedded, and appended, respectively).
```
troyronda: This line could use a bit more explanation.

features/0351-generic-json-message/README.md
troyronda: Makes sense. I think you should retain a summary in the HTTP one as well.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The `purpose` of a message is a machine-readable string value which describes the *semantics* of a message - the role it plays within a protocol implemented using this RFC, for example.
+
+The purpose is one mechanism for determining which recipient(s) should be sent a message:
+
+The recipient registers on a particular purpose value with the recipient-agent, in which case the recipient-agent must relay any message with that value to the recipient. Multiple recipients can register with a single recipient-agent on a single purpose value, and a recipient can register with multiple recipient-agents.
+
+##### `data-type`
+
+The sender must provide a `data-type` identification string constant, which indicates, for the recipient, the format of the data and how it is to be processed.
```
troyronda: I see the companion RFC uses `http/request`. Are you supplying guidance in the generic RFC on formatting of this value?

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+This is the generic JSON message protocol, defined uniquely by the URI:
+
+    "https://didcomm.org/json/1.0"
+
+#### Key Concepts
+
+> This RFC assumes familiarity with [mediators and relays](/concepts/0046-mediators-and-relays/README.md), [attachments](/concepts/0017-attachments/README.md), and [message threading](/concepts/0008-message-id-and-threading/README.md)
+
+This protocol has two primary roles, the **sender** and the **recipient**. The sender constructs a message with a JSON payload, which the recipient parses. This protocol supports optional threading, so the recipient can send a response to the sender.
```
troyronda: i also think the optional message timing decorator would be useful.

features/0351-generic-json-message/README.md
kdenhartog: I'm not sure I fully understand what you're after. Are you looking for the ability to specify the `@type` field still with a unique URI then also have the ability to include other details about the specifics of the message so it can be parsed differently?

If so, this could be defined within the data model without concern. If we need to move it to a shared attribute that can't be collided with then I think it would require more input and consideration from others.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,130 @@
...
+- The messages in your protocol may have a message-type identifier, which could be used for the `data-type` value.
+
+## Drawbacks
+
+## Rationale and alternatives
+
+- Why don't we standardize data-type labels against content schemas and format definitions? The intent here is to allow custom protocols and messaging schemes to operate over DIDComm. For this we provide a message format that allows them to be constructed without a chance of collision (between message names, for example) with other DIDComm message types, and we allow custom formats to be designed without needing to standardize them through the RFC process. The data-type label allows implementors to ensure that they will not collide between any of their message types in their specific implementations. A custom schema could be defined in its own RFC.
+
+## Prior art
+
```
troyronda: DIF Identity Hubs have Actions. https://github.com/decentralized-identity/identity-hub/blob/master/explainer.md#actions

However, this RFC is describing an agent-based mechanism rather than a Hub one.

features/0351-generic-json-message/README.md
troyronda: @kdenhartog my comment was not intended to be about the type field. @Moopli added a purpose field into the data model.

You can also see how DIF Identity Hubs defined Actions (comment below - https://github.com/hyperledger/aries-rfcs/pull/351/files#r362265169)

troyronda: @kdenhartog  Note: from the above there are (at least) three possible variations:
1. `@type`=`https://didcomm.org/json/1.0/message`
  - `data-type` defines the schema of the payload.
  - `purpose` defines the Intent allowing external apps to register.
  - attachment has the payload (and meta info like mime-type).
2. `@type` defines the schema.
  - `~purpose` (as a decorator) defines the Intent allowing external apps to register.
  - attachment has the payload (and meta info like mime-type).
3. `@type` contains `https://didcomm.org/json/1.0/message` and also some schema information.
  - `purpose` defines the Intent allowing external apps to register.
  - attachment has the payload (and meta info like mime-type).

The RFC currently uses the first variation. Between 2 & 3, I would assume the decorator style (2) would be more consistent.

features/0351-generic-json-message/README.md
Moopli: Enforcing a path-like hierarchical purpose format means we also need to be more specific about how recipients subscribe - for example, in MQTT you can subscribe to a particular value on a particular level, like `/+/+/Abcde` for all `Abcde` on the third level down from `/`. 

As another example (which I think would not work here) in Apple's Uniform Type Identifiers, every type is explicitly defined, including defining which types they 'inherit' from (as in, an instance of type A is also an instance of type B). We can't do some sort of registry, so subscription flexibility needs to be inherent in the purpose format.

features/0351-generic-json-message/README.md
```diff
@@ -0,0 +1,146 @@
...
+The `~purpose` decorator is a JSON array which describes the *semantics* of a message - the role it plays within a protocol implemented using this RFC, for example, or the type or meaning of the data contained within.
+The purpose is the mechanism for determining which recipient(s) should be sent a message.
+
+Example: `"~purpose": ["team:01453", "delivery", "geotag", "cred"]`
+
+Each element of the purpose array is a string. An agent provides some means for recipients to *register* on a purpose, or class of purposes, by indicating the particular string values they are interested in.
+
+The particular registration semantics left to the implementation. Some possible examples include:
+- A tagging system, where if a recipient registers on a list `"foo", "bar"`, it will be forwarded messages with purposes `["foo", "quux"]` and `["baz", "bar"]`
+- A hierarchical system, where if a recipient registers on a list `"foo", "bar"`, it will receive any message with purpose `["foo", "bar", ...]` but not `["foo", "baz", ...]` or `["baz", "foo", "bar", ...]`
```
llorllale: @Moopli is this an *ordered* array? Does the order affect the comparator function?

Moopli: Yes it is, and it could indeed, if we make the comparator care about order. One of my examples (the hierarchical one) _does_ care about order.

features/0351-purpose-decorator/README.md
```diff
@@ -0,0 +1,125 @@
+# Aries RFC 0351: Purpose Decorator
+- Authors: [Filip Burlacu](filip.burlacu@securekey.com)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2019-12-16
+- Status Note: implementation is being explored by SecureKey
+- Supersedes:
+- Start Date: 2018-12-05
```
Moopli: 2019 not 2018

features/0351-purpose-decorator/README.md
```diff
@@ -0,0 +1,125 @@
...
+### The `~purpose` Decorator
+
+The `~purpose` decorator is a JSON array which describes the *semantics* of a message - the role it plays within a protocol implemented using this RFC, for example, or the type or meaning of the data contained within.
+The purpose is the mechanism for determining which recipient(s) should be sent a message.
+
+Example: `"~purpose": ["team:01453", "delivery", "geotag", "cred"]`
+
+Each element of the purpose array is a string. An agent provides some means for recipients to *register* on a purpose, or class of purposes, by indicating the particular string values they are interested in.
+
+The particular registration semantics left to the implementation. Some possible examples include:
```
llorllale: How can an Agent differentiate between the intended semantics (if it supports more than one)?

features/0351-purpose-decorator/README.md
```diff
@@ -0,0 +1,125 @@
...
+This example protocol wraps the message within the `data` field. 
+
+```json
+{
+  "@id": "123456789",
+  "@type": "https://example.org/didcomm-message",
+  "~purpose": [],
+  "data" : {}
+}
+```
```
llorllale: Example of the message format using this decorator:

```json
{
  "@id": "123456789",
  "@type": https://example.org/didcomm-message",
  "~purpose": ["metrics", "latency"],
  "data": {"mean": 346457, "median": 2344}
}
```

Moopli:
@swcurran This is ready to merge (I can't add the label as I'm not a repo contributor)

PR #359, opened 2019-12-19
Added notes to say that RFCs that currently use the new format should continue to do so

swcurran:


swcurran:
@troyronda - please confirm that these changes address your concern.  If so, these will be merged and I'll submit a new PR to start the "Accepted" status change clock.

features/0348-transition-msg-type-to-https/README.md
```diff
@@ -11,15 +11,16 @@
 ## Summary
 
 Per issue [#225](https://github.com/hyperledger/aries-rfcs/issues/225), the
-Aries community has agreed to change the prefix for protocol message types from
-`did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/` to `https://didcomm.org/`. Examples of the two message types forms are:
+Aries community has agreed to change the prefix for protocol message types that currently use the
```
troyronda: I also think newly written RFCs should also use the new format. i.e., we shouldn't need to transition new efforts.

troyronda:
@swcurran Thanks for the update! I think the new text nicely addresses the already existing RFCs. I would also suggest that new RFCs should also just start with the new format (commented inline).

swcurran:
Good catch @troyronda - added.

dhh1128:
I'm cool with merging this.

PR #360, opened 2019-12-19
New RFC proposing the use of the did:key method in place of plain public keys

swcurran:


features/0360-use-did-key/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Motivation
+
+When one Aries agent inserts a public key into the JSON of an Aries message (for example, the `~service` decorator), it assumes that the recipient agent will use the key in the intended way. At the time this RFC is being written, this is easy because only one key type is in use by all agents. However, in order to enable the use of different cryptography algorithms, the public references must be extended to at least include the key type. The preferred and concise way to do that is the use of the [multicodec](https://github.com/multiformats/multicodec) mechanism, which provides a registry of encodings for known key types that are prefixed to the public key in a standard and concise way. [did:key](https://digitalbazaar.github.io/did-method-key) extends that mechanism by providing a templated way to transform the combination of public key and key type into a [DID-standard DIDDoc](https://digitalbazaar.github.io/did-method-key/#example-2-a-did-document-derived-from-a-did-key).
```
troyronda: hopefully the W3C repo becomes populated: 
- https://github.com/w3c-ccg/did-method-key
- https://w3c-ccg.github.io/did-method-key

(should have a TODO to update the references & links, once available.)

features/0360-use-did-key/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `did:key` method uses the strings that are the DID, public key and key type to construct ("resolve") a DIDDoc based on a template defined by the `did:key` specification. Further, the `did:key` resolver generates, in the case of an ed25519 public signing key, a key that can be used as part of a Diffie-Hellman exchange appropriate for encryption in the `keyAgreement` section of the DIDDoc. Presumably, as `did:key` supports other key types, similar DIDDoc templates will become part of the specification. Key types that don't support a signing/key exchange transformation would not have a `keyAgreement` entry in the resolved DIDDoc.
+
+The following currently implemented RFCs would be affected by adoption of this RFC. In these RFCs, the JSON items that currently contain naked public keys (mostly the items `recipientKeys` and `routingKeys`) would be changed to use `did:key` references where applicable. Note that in these items public DIDs could also be used if applicable for a given use case.
```
troyronda: I think we should separate the discussion of "naked" keys in the envelope from the other "naked" keys topics.

In particular, I think the JWE envelope ([RFC 360](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope)) should continue to support (and use) JWKs. 
- A JWK representation is well-known (and implemented) for JWEs: https://tools.ietf.org/html/rfc7516#section-4.1.5. 
- The underlying spec for ECDH-1PU references JWK for the "epk" field: https://tools.ietf.org/html/rfc7518#section-4.6.1.1

swcurran: @troyronda - Removed encryption envelope from list of "to be changed" protocols; added to do about discussing the use or not of did:key with the encryption envelope.

features/0360-use-did-key/README.md
swcurran: Added To Do.

dhh1128:
I thought that when we discussed this at the connectathon, we agreed that this would be a temporary compromise. The way the RFC is written makes it sound as if this is a solid long-term solution. I'd like a paragraph that says this is a "good enough for the time being" solution and that we want to look at better solutions in the long term.

swcurran:
I don't recall that, but I think I missed one of the discussions on this. @tplooker - do you agree the phrasing that @dhh1128 recommends be added?

It doesn't really make sense to me, as I think that phrasing applies to all RFCs/Protocols.  If there is a specific direction that is known that we should mention, then it makes sense to include, but a vague "there must be a better way" phrase seems odd.

dhh1128:
>It doesn't really make sense to me, as I think that phrasing applies to all RFCs/Protocols. If there is a specific direction that is known that we should mention, then it makes sense to include, but a vague "there must be a better way" phrase seems odd.

It is true that all RFCs express the best idea we have today, and are replaceable by a better idea if we find one tomorrow. However, not all RFCs are temporary compromises that we know to be weird.

When the thing that we want is semantically a key, providing a DID instead is weird. The thing we're trying to provide when we propose to use did:key is a key, not a DID. If we had a cleaner way to specify a key, that would be preferable to this mechanism.

I have never bought off on Tobias's assertion that we want the extra sections of a DID doc that tell us how the key may be used. To me, this is an ugly consequence of a workaround, not a desirable feature.

swcurran:
Updated the RFC to add the comment about using a DID to represent a public key, and the notion of changing to a new representation (when found) that is "just a key" with additional beneficial properties (and not a DID).

dhh1128:
Okay, I like this with the extra proviso that got added.

PR #363, opened 2019-12-30
tweak: describe coprotocols, relationship to APIs

dhh1128:


dhh1128:
As this is a minor tweak that just clarifies some tribal knowledge, I am merging it directly.

swcurran:
The co-protocol concept is interesting.  What happens to the threading fields in a co-protocol scenario? Is it that there is no ~thread item? If there is a thread item, how is it connected to the existing thread, but is not assumed to be a parent thread?

PR #365, opened 2020-01-08
revise RFC 0104: "Delegatable" --> "Chained"

dhh1128:
This explains delegation in the more general context of data provenance, and explains how chained credentials can be applied to delegation but also to other use cases. The RFC is significantly revised, although parts of it are still recognizable.

Since nobody is implementing this RFC yet, and it is still in its original PROPOSED state, I hope the change is not controversial.


dhh1128:
Thanks for the excellent feedback, @swcurran. I agree with all of it and will address each item. I knew this PR was imperfect when I raised it, but I was trying to get some thinking into print instead of letting draft documents languish forever, so I hoped this PR would be acceptable more or less as-is, with a follow-on PR to give additional examples and so forth.

@swcurran , if I update the PR with some quick fixes, would you be okay to merge, with the understanding that a subsequent PR would be raised for the more substantive improvements?

swcurran:
I'm trying to merge it now (even without the tweaks), but the circle/ci
process is not completing...

PR #366, opened 2020-01-08
didexchange RFC0023: PROPOSED --> DEMONSTRATED

llorllale:

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -1,7 +1,7 @@
 # Aries RFC 0023: DID Exchange Protocol 1.0
 
 - Authors: [Ryan West](ryan.west@sovrin.org), [Daniel Bluhm](daniel.bluhm@sovrin.org), Matthew Hailstone, Stephen Curran, [Sam Curren](sam@sovrin.org)
-- Status: [PROPOSED](/README.md#proposed)
+- Status: [DEMONSTRATED](/README.md#demonstrated)
 - Since: 2019-05-27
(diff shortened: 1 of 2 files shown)


TelegramSam:
I'm fairly uncomfortable with this status change given the need to apply the key representation stuff not yet applied to this RFC.

swcurran:
I would also like to extend the discussion from last week's call about moving the invitation message out of the Connections/DID-Exchange protocols to be either on it's own or part of a more general family of connection-less/plaintext messages.

kdenhartog:
I'd be curious what the takeaways and learnings were on the key representation based on trying to implement this. I assume the implementation made a few assumptions in order to progress, so some insights from it would be useful.

llorllale:
@Baha-sk @Moopli can share their insights

Baha-sk:
Key representation is still wip.

Since we're aiming to provide pluggable crypto/kms services and our default implementation will be using Tink, keys will be handled/managed internally, ie only an Uint32 values will be used as the keyId (generated and managed by Tink). Aries Agents don't need to worry about expiry or rotation of keys as they will be fully managed by the underlying Tink framework. The status of this change is still wip and will be pushed when ready. 

the link you posted above @kdenhartog represents keysets with encoding and envelope encryption equivalent to the current state of ACA-py. The new key format is not published yet.

but outside of crypto and key represenation work, DIDExchange is completed as far as Aries-framework-go is concerned. What @swcurran is proposing (moving invitation outside of DIDExchange protocol) can be included in a separate (new or existing) RFC.

dhh1128:
I agree that key rep and moving invitation out would be desirable before did exchange becomes accepted.

TelegramSam:
Discussed in Aries WG Call 2020-02-12. Will close until the state matures sufficiently then reopen this change.

PR #368, opened 2020-01-10
Tweak diff ordering in AIP script; update version of 0003, 0005, 0160 for non-breaking changes, remove 0031 and 0048

swcurran:


esplinr:
Thank you Stephen. I'm comfortable with merging this. Then I'll raise an issue to continue the discussion about the Service Decorator.

PR #371, opened 2020-01-13
allow empty commits on action

TelegramSam:
this will prevent errors when a PR results in no change to the index.


TelegramSam:
bad commit for DCO

PR #376, opened 2020-01-14
Referents for attributes in presentation proposals

sklump:
Distinguish between a desire for a proof_req with `requested_attributes`/`referent`/`names` and one with `requested_attributes`/`referent`/`name` over multiple distinct referents.

I believe this approach represents the minimum possible change to the existing spec.

### DIFF
--- a/features/0037-present-proof/README.md
+++ b/features/0037-present-proof/README.md
@@ -123,7 +123,8 @@ This is not a message but an inner object for other messages in this protocol. I
             "name": "<attribute_name>",
             "cred_def_id": "<cred_def_id>",
             "mime-type": "<type>",
-            "value": "<value>"
+            "value": "<value>",
+            "referent": "<referent>"
         },
         // more attributes
     ],
@@ -165,8 +166,41 @@ The optional `value`, when present, holds the value of the attribute to reveal i
 An attribute specification must specify a `value`, a `cred_def_id`, or both: 
 
 * if `value` is present and `cred_def_id` is absent, the preview proposes a self-attested attribute;
-* if `value` and `cred_def_id` are both present, the preview proposes verifiable claim to reveal in the presentation;
-* if `value` is absent and `cred_def_id` is present, the preview proposes verifiable claim not to reveal in the presentation.
+* if `value` and `cred_def_id` are both present, the preview proposes a verifiable claim to reveal in the presentation;
+* if `value` is absent and `cred_def_id` is present, the preview proposes a verifiable claim not to reveal in the presentation.
+
+##### Referent
+
+The optional `referent` can be useful in specifying multiple-credential presentations. Its value indicates which credential 
+will supply the attribute in the presentation. Sharing a `referent` value between multiple attribute specifications indicates that the holder's same credential supplies the attribute.
+
+Any attribute specification using a `referent` must also have a `cred_def_id`; any attribute specifications sharing a common `referent` value must all have the same `cred_def_id` value.
+
+For example, a holder with multiple account credentials could use a presentation preview such as
+
+```jsonc
+{
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/presentation-preview",
+    "attributes": [
+        {
+            "name": "account",
+            "cred_def_id": "BzCbsNYhMrjHiqZDTUASHg:3:CL:1234:tag",
+            "value": "12345678",
+            "referent": "0"
+        },
+        {
+            "name": "streetAddress",
+            "cred_def_id": "BzCbsNYhMrjHiqZDTUASHg:3:CL:1234:tag",
+            "value": "123 Main Street",
+            "referent": "0"
+        },
+    ],
+    "predicates": [
+    ]
+}
+```
+
+to prompt a verifier to request proof of account number and street address from the same account, rather than potentially an account number and street address from distinct accounts.
 
 #### Predicates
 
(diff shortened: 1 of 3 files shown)


swcurran:
I believe this change would require a change in the message format, and so a version change.  Would you agree?

sklump:
Definitely. 1.01? 1.1? 1.0.1? I wouldn't likely know best.

swcurran:
@dhh1128 @TelegramSam - I'm not quite sure on the versioning we should use. Could you jump in on what the new version should be?

sklump:
I see RFC 0036 uses 1.01 in its latest tweak?

dhh1128:
[Protocols use semver rules for versioning]. Semver says "1.01" is illegal; you must use either 1.1 or 1.0.1.

In general, I believe the patch section of semver is not used with protocols, so I'd advise 1.1 as the answer.

swcurran:
So does that mean we can only have 10 versions, or that after 1.9 we get
1.10.  I guess that makes sense - these are strings not numbers.  Better
yet, I should read the semver rules more closely.

dhh1128:
I take back part of what I said about versioning. Yes, we use semver. However, I don't think it applies in this case. After reviewing @sklump 's proposed change more thoughtfully, we're not actually changing anything in the protocol; we're simply clarifying a rule that has been true all along. I wouldn't bump the version of the protocol at all for a simple clarification to docs.

dhh1128:
Wait. Now I'm embarrassed. I guess the diff I just looked at was not a diff of the PR, but rather a diff between two commits in the PR. The PR *does* introduce a new field in a message, so everything I said about semver *does* come into play. This makes it v 1.1 instead of v 1.0. Sorry to be going back and forth.

sklump:
> 
> 
> Wait. Now I'm embarrassed. I guess the diff I just looked at was not a diff of the PR, but rather a diff between two commits in the PR. The PR _does_ introduce a new field in a message, so everything I said about semver _does_ come into play. This makes it v 1.1 instead of v 1.0. Sorry to be going back and forth.

It's OK to think aloud, few ideas are fully formed out of the gate.

sklump:
(... changed example to something less contrived, inspired by John Jordan's musing)

dhh1128:
So, I think we could merge this if the versioning were taken care of. IIRC, @swcurran suggested that we update the doc with its current version stamp (1.1) but put a section {at the top?} named {what?} that pointed to previous versions of the same protocol by git commit hash...

swcurran:
There is an example of the change log section here: https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential#version-change-log

I'll get that added to the appropriate document in the header.

sklump:
Absent version numbering discussions, are there any comments about the content of this update?

dhh1128:
I'm very happy with the content. I just want the versioning thing taken care of.

features/0037-present-proof/README.md
```diff
@@ -165,8 +166,41 @@ The optional `value`, when present, holds the value of the attribute to reveal i
...
-* if `value` is absent and `cred_def_id` is present, the preview proposes verifiable claim not to reveal in the presentation.
+* if `value` and `cred_def_id` are both present, the preview proposes a verifiable claim to reveal in the presentation;
+* if `value` is absent and `cred_def_id` is present, the preview proposes a verifiable claim not to reveal in the presentation.
+
+##### Referent
+
+The optional `referent` can be useful in specifying multiple-credential presentations. Its value indicates which credential 
+will supply the attribute in the presentation. Sharing a `referent` value between multiple attribute specifications indicates that the holder's same credential supplies the attribute.
+
+Any attribute specification using a `referent` must also have a `cred_def_id`; any attribute specifications sharing a common `referent` value must all have the same `cred_def_id` value.
```
dhh1128: Why? Wouldn't the more logical requirement be that there must be at least one cred_def_id value across all items with a common referent--and that all of them that do appear must be consistent? In other words, if I used the same referent 6 times, but only mentioned the cred_def_id once, wouldn't that be good enough?

What happens when this isn't an Indy-style cred, so it lacks a cred def? Couldn't other cred types use a credential id (not cred *def* id) as the thing they stitch together?

sklump: Note that the holder already has credentials from which he creates the proposal, so the holder necessarily knows the cred def id. Elsewhere in the RFC, recall that skipping the cred def id marks the attribute as self-attested. So the other 5 attributes would count as self-attested here. Self-attested attributes don't come from a credential, unless I've completely spaced on something fundamental.

If it wasn't clear to you, it won't be clear to anyone else: I will amend the wording to be explicit here, and submit via this PR, Monday.

dhh1128: This explanation makes perfect sense. Thank you. I agree that a clarifying sentence or two would be helpful.

sklump: Updated accordingly.

sklump:
Sorry folks, git is mystifying me this morning. Files I never meant to touch have shown up in the PR:
features/0335-http-over-didcomm/README.md and features/0348-transition-msg-type-to-https/README.md.

Would it be OK if I started from master again and proposed a new PR for this work?

dhh1128:
@sklump When I get in situations like this (happens periodically), I typically like to start over rather than continuing to fix stuff with anomalous commits. You can do it on the existing PR if your git kung fu is really good, but it's easier to just start clean and cherry-pick the commits you want into something fresh.

sklump:
Superseded by https://github.com/hyperledger/aries-rfcs/pull/403.

Issue #377, opened 2020-01-15
RFC 0036 version 1.01 is not semver compliant


sklump:
"1.01" violates https://semver.org/#spec-item-2

I believe it ought to be "1.1"?

swcurran:
Corrected with #379

PR #379, opened 2020-01-15
Change version of the propose-credential from 1.01 to 1.1 to be semvar compliant

swcurran:


swcurran:
@smithbk especially, and others - please note this change to correct the use of an incorrect (according to semvar) version number.  I think it was me that suggested the version be "1.01" for the updated protocol, but that is wrong - it should be "1.1" according to semvar rules.

Hopefully this doesn't break anything that is out in the wild and can be easily fixed. Note that AIP 1.0.0 does not point to this version of the protocol, it uses the previous one, so there shouldn't be anything out in the wild using this - hence the proposed correction.

Please respond and if OK we can merge this sooner than later.  Reference #377.

smithbk:
Agreed

PR #387, opened 2020-01-22
RFC 0302 AIP 1.0.0 --> ACCEPTED; Deadline: 2020.01.30

swcurran:
Implementation Builders should vote yay/nay on this PR for merging.

@tmarkovski @tplooker @esplinr @troyronda @johncallahan @jadhavajay
@TelegramSam @dbluhm @smithbk @idramp 

and any others in the community.


swcurran:
Also on the PR - an update to the link to include a clarification in RFC 0003 Protocols.

swcurran:
This PR is now ready for merging.  As you can see from the list of commits made, I've:

- Created a special commit of the 0036 credential issue RFC that has the claims encoding info but not the credential proposal 1.1.
  - AIP 1.0 points to that commit
  - A new commit was added that adds back in the credential proposal 1.1 text.
- Updated the links to the latest versions of 0019 encryption envelope and 0160 Connections to include base64URL decoding.
- Updated the link to the latest 0037 present-proof RFC to include claims encoding guidance.
- Dropped the .patch element of the versioning
- Updated the dates and status

I think we're ready to have AIP 1.0 accepted, per the discussion on the Aries WG call from today.  Please review and merge.

Issue #370, opened 2020-01-13
RFC 0036 v1.1 does not include schema issuer DID?


sklump:
Version 1.1 broadened the credential proposal to accommodate the prospective holder having only partial information regarding the credential in proposition.

However, the update misses the schema issuer DID as a criterion. Is this an oversight, on purpose, or maybe just a typo?

TelegramSam:
Appears resolved via RFC update as indicated.

PR #393, opened 2020-01-23
schema issuer DID is a criterion we must support; e.g., Ontarian university transcript, n…

sklump:
…ot Elbonian

Addresses https://github.com/hyperledger/aries-rfcs/issues/370.

### DIFF
--- a/features/0036-issue-credential/README.md
+++ b/features/0036-issue-credential/README.md
@@ -87,6 +87,7 @@ An optional message sent by the potential Holder to the Issuer to initiate the p
     "@id": "<uuid-of-propose-message>",
     "comment": "some comment",
     "credential_proposal": <json-ld object>,
+    "schema_issuer_did": "DID of the proposed schema issuer",
     "schema_id": "Schema ID string",
     "schema_name": "Schema name string",
     "schema_version": "Schema version string",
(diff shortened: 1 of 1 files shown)


features/0036-issue-credential/README.md
```diff
@@ -99,6 +100,7 @@ Description of attributes:
 
 * `comment` -- an optional field that provides human readable information about this Credential Proposal, so the proposal can be evaluated by human judgment. Follows [DIDComm conventions for l10n](../0043-l10n/README.md).
 * `credential_proposal` -- an optional JSON-LD object that represents the credential data that Prover wants to receive. It matches the schema of [Credential Preview](#preview-credential).
+* `schema_issuer_did` -- optional filter to request credential based on a particular Schema issuer DID.
```
dhh1128: The concept of this field is fine. However, I think it's misnamed. Schemas aren't issued; they're written--and not necessarily by credential issuers. Furthermore, the writer of a schema and the party who registers it on the ledger aren't guaranteed to be the same, either. So a better name for this might be "schema_registrant_did" or "schema_submitter_did".

sklump: Indy-sdk uses schema_issuer_did in its restrictions, and deviating from the term will introduce a distinction without a difference. I'm OK with introducing as much confusion as possible, since I'm a contractor who lives in the space and it makes for more trade secrets for me, but ideally the time to discuss what to call it was about 2016. Recall that I had proposed origin for schema-issuer under VON anchor, but that got no traction and I adopted indy-sdk nomenclature in the end.

I can go either way.

dhh1128: Hmm. Now I don't know what to do. @jovfer , what do you think? How strong is this terminology in Indy?

jovfer: @dhh1128 it's a part of filters / tags in the wallet. We can do some migration for new versions of libindy, but it's a typical compatibility problem.
I think the value of the renaming is not worth solving of compatibility problem in current versions.
I'd prefer to change it in 1.1 or at the time of Rich Schema release

dhh1128: Okay. Yuck, but okay.

PR #394, opened 2020-01-23
RFC 0335: Clarifications and addenda

Moopli:
- No longer enforces that the purpose is based on the request URI
- Added benefits of DIDComm for trust and blinding
- Client and server can use HTTP/2 to talk to their agents
- Client and server can use HTTPS to talk to their agents
- Describes rationale of not doing TLS over DIDComm
- Client and server side are decoupled, they don't both
  have to be non-didcomm apps using http to talk through
  an agent
- Message body field is optional

### DIFF
--- a/features/0335-http-over-didcomm/README.md
+++ b/features/0335-http-over-didcomm/README.md
@@ -20,6 +20,10 @@ Carl wants to apply for a car loan from his friendly neighborhood used car deale
 
 HTTP over DIDComm turns a dev + ops problem, of redesigning and deploying your server and client to use DID communication, into an ops problem - deploying Aries infrastructure in front of your server and to your clients.
 
+Using HTTP over DIDComm as opposed to HTTPS between a client and server offers some key benefits:
+- The client and server can use methods provided by Aries agents to verify their trust in the other party - for example, by presenting verifiable credential proofs. In particular, this allows decentralized client verification and trust, as opposed to client certs.
+- The client and server can be blind to each others' identities (for example, using fresh peer DIDs and communicating through a router), even while using their agents to ensure trust.
+
 ## Tutorial
 
 #### Name and Version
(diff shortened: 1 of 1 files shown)


Moopli:
@dhh1128 this PR answers your questions regarding HTTP/1.x vs HTTP/2 and using HTTP on only one side of the protocol (or possibly neither, if a 0335-compatible client talks to a 0335-compatible server).
@troyronda this PR takes "purpose is part of the resource URI" out of the spec and into an example, so the spec works for server/client blinding.

features/0335-http-over-didcomm/README.md
```diff
@@ -77,16 +81,16 @@ DIDComm messages for this protocol look as follows:
 {
   "@type": "https://didcomm.org/http-over-didcomm/1.0/request",
   "@id": "2a0ec6db-471d-42ed-84ee-f9544db9da4b",
-  "~purpose": [<partial resource uri: scheme://userinfo@host:port/path>],
+  "~purpose": [],
   "method": <method>,
   "resource-uri": <resource uri value>,
```
sudeshrshetty: `resource-uri` can also be optional in case where agent wants to hide identity of http server.

Moopli: Done.

features/0335-http-over-didcomm/README.md
```diff
@@ -77,16 +81,18 @@ DIDComm messages for this protocol look as follows:
...
   "version": <version>,
   "headers": [],
```
sudeshrshetty: http headers should be in key-value, not array.

dhh1128: I agree that this would be more natural from some perspectives. However, DIDComm best practice says to use an array of tuples rather than key-value. For an explanation of why, see https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0074-didcomm-best-practices#dont-treat-objects-as-associative-arrays.

PR #410, opened 2020-01-31
Add terms and deny message and clarify scope

TelegramSam:


TelegramSam:
@dhh1128 How do these changes look?

features/0211-route-coordination/README.md
```diff
@@ -39,10 +45,27 @@ This message serves as a request from the recipient to the router, asking for th
 ```jsonc
 {
     "@id": "123456781",
-    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routecoordination/1.0/route-request"
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/routecoordination/1.0/route-request",
+    "terms": [],
```
dhh1128: The distinction "terms" vs. "user terms" feels client-server-ish and centralized to me. How about "mediator_terms" and "mediated_terms" instead?

dhh1128: Or alternately, "router_terms" and "routed_terms"?

dhh1128: (To me "routing" includes both mediation and relay, but terms of service make a lot more sense for mediators. This is because neither the sender nor the recipient is guaranteed to know about relays -- but both have to know about mediators. Mediators decrypt something, and under the enhanced forwarding protocol that I might eventually raise as an RFC proposal, they do more than just push a message along. They might deliver it after a delay, hold it until it's retrieved, etc.)

features/0211-route-coordination/README.md
```diff
@@ -26,7 +30,9 @@ A recipient may discover an agent capable of routing using the Feature Discovery
 
 First, the _recipient_ sends a `route_request` message to the _router_. If the _router_ is willing to route messages, it will respond with a `route_grant` message. The _recipient_ will share the routing information in the grant message with other contacts.
 
-When a new key is used by the _recipient_, it must be registered with the _router_ to enable route identitifcation. This is done with a `keylist_update` message.
+If the _router_ requires the _recipient_  to agree to terms prior to service, a `route_deny` message will be returned listing the URIs of terms that the user must agree to. Term agreement is indicated by including the same URIs in the `terms` attribute of the `route_requeste` message. Terms the _recipient_ wish the _router_ to agree to are indicated in the `user_terms` attribute of the `route_request` message. The *router* may indicate which user terms they support in the `supported_user_terms` attribute of the `Route Deny` message.
```
dhh1128: See my comment a few lines below about why I don't think we should be promoting the term "router" or "routing" here.

dhh1128: Typo: "route_requeste" --> "route_request"

features/0211-route-coordination/README.md
TelegramSam: `router_terms` and 'recipient_terms`? This would match the language elsewhere in this RFC.

features/0211-route-coordination/README.md
TelegramSam: On router vs mediator: This protocol could technically coordinate either mediators or relays. Relays are not _required_ to be known to the recipient, but they could be.

features/0211-route-coordination/README.md
dhh1128: It seems to me that you're trying to shift terminology, popularizing "router" instead of "mediator". What I'm trying to point out is that routing is bigger than mediation, but the concepts you're talking about are pure mediation. So I'm still feeling dissonance.

features/0211-route-coordination/README.md
swcurran: Regards this - "Relays are not required to be known to the recipient, but they could be." If they are known to the recipient, they are by definition a mediator and not a relay. AFAIK - that is the difference between the two.

dhh1128: Ooh, this is getting nuanced. :-)

A mediator *must* be known to both parties, because accommodations must be made in the crypto to account for mediation; these bidirectional accommodations are what makes them a mediator. A relay *could* be known, but generally wouldn't be, because there is no point. (Relays don't have keys and make no accommodations, so there's nothing to tell the other party about or agree to.) Thus, a party that's doing relay, that's known, still isn't a mediator if it doesn't need to be accounted for in the crypto.

To say it another way: the defining characteristic of a mediator is that they receive encrypted messages, *encrypted for them*, do a decryption, and then take intelligent action on the plaintext they see (typically the plaintext is a `forward` message with a further-encrypted payload). The mediator's behavior is something that the sender and recipient have both agreed upon.

dhh1128:
I think this should be merged. However, I think it's worth noting that this particular protocol probably belongs in the DIDComm Messaging spec as part of the routing content; it is unlikely to live a long and productive life as an independent Aries RFC.

PR #411, opened 2020-02-02
0289: Mention TLS and give an example where it does not suffice

geonnave:
While reading the motivation section I felt the need to contrast the ToIP proposal against existing techniques such as TLS. As I understand that a detailed comparison could be large enough to reside on a separarte document, I briefly mention TLS and its importance so far, and give an example to help the reader have a sense of where TLS and ToIP might differ.


concepts/0289-toip-stack/README.md
```diff
@@ -13,8 +13,11 @@ This Aries concept RFC introduces a complete architecture for Internet-scale dig
...
+Without a doubt, implementations of the TCP/IP stack, followed by SSL/TLS, have driven a tremendous amount of innovation over the last 30 years.
+However, although protocols such as TLS offer world-class security, the architecture over which they have been built upon leaves a significant and widely-recognized gap: a means for any peer to establish trust over these digital connections. For example, while TLS does allow a user to trust she is accessing the right website, it does not offer, at least in an usable way, a way for the user to _log in_, or prove her identity, to the website. This gap has often been referred to as "the Internet's missing identity layer" [3].
```
dhh1128: "over which they have been built upon" has too many prepositions; shouldn't we remove either "over" or "upon"?

geonnave: Sure, I appreciate the correction (not a native speaker). I have removed the "upon".

TelegramSam:
@talltree ^

concepts/0289-toip-stack/README.md
talltree: Sorry to take so long to get to this, but I have reviewed it now and I think it makes a good point about SSL/TLS in the evolutionary progress. I'm in favor of merging this PR.

geonnave: No problem, thank you!

Issue #408, opened 2020-01-31
RFCs 0036 and 0037 (and others?): need formal state names


dhh1128:
I was in these RFCs, and noticed that although they have nice choreography diagrams, they don't provide formal names for their states. This is probably because they didn't build state machine matrices, which is fine (the matrices are just an optional diagram; the choreography diagrams are an experiment to see if the same content can be better represented the other way). Whatever diagram we do or do not use, we need formal names for the states, because problem reports need to report state in an interoperable way. This is the basis of DIDComm features like "I experienced an error while in <state 7>, and I'm now dropping back to <state 6>", and they are also [implied by RFC 0317 please_ack](https://github.com/hyperledger/aries-rfcs/tree/master/features/0317-please-ack#advanced-features-experimental). Today this is mostly tribal knowledge about future direction, not a feature we depend on. But I think this will become important as protocols become more complex, and as our standards for robustness and error handling go up.

So I recommend that we agree to assign formal state names in each place that they are lacking. If we do, I will then take an action item to submit PR(s) to do so.

@rajeshkalaria80, @dkulic

TelegramSam:
I agree that formalized names make sense. @dhh1128 I think an example PR would make this pretty clear and promote any discussion we need to have.

dhh1128:
Agreed. I will prepare an example that we can discuss on Wednesday's community call.

dhh1128:
I have raised PR #412 as an example of what I'm thinking. I have also prepared a [slide deck](https://docs.google.com/presentation/d/17hk6QqLW5M9E4TBPZwXIUBu9eEinMNXReceyZTF4LpA/edit) to explore the issue in a community call.

swcurran:
Closing this.  There is a new PR that is in the same vein for Present Proof (0037), but we can deal with this without having this issue open.

PR #412, opened 2020-02-03
tweak: add named states to RFCs 0036, 0037 per issue #408

dhh1128:


concepts/0003-protocols/template.md
```diff
@@ -93,7 +93,7 @@ the matrix form is used in many early RFCs. We leave it up to
 the community to settle on whether it wants to strongly recommend specific
 diagram types.
 
-The formal names for each state are important, as they are used in [`ack`s]( https://github.com/hyperledger/indy-hipe/pull/77)
+Regardless of the type of diagram that's used, [the formal names for each state are important](https://docs.google.com/presentation/d/17hk6QqLW5M9E4TBPZwXIUBu9eEinMNXReceyZTF4LpA/edit?usp=sharing), as they are used in [`ack`s]( https://github.com/hyperledger/indy-hipe/pull/77)
```
swcurran: This reference is to a HIPE PR. Pretty sure it should be to the Aries RFC.

concepts/0003-protocols/template.md
```diff
@@ -93,7 +93,7 @@ the matrix form is used in many early RFCs. We leave it up to
...
 and [`problem-report`s](https://github.com/hyperledger/indy-hipe/pull/65)).
```
swcurran: Link to HIPE again.

features/0036-issue-credential/README.md
```diff
@@ -35,9 +35,25 @@ There are two roles in this protocol: Issuer and Holder. Technically, the latter
...
+
+#### states for Holder
+
+* proposal-sent
+* offer-received
+* request-sent
+* credential-received
+* done
+
+Errors might occur in various places. For example, an Issuer might offer a credential for a price that the Holder is unwilling to pay. All errors are modeled with a `problem-report` message. Easy-to-anticipate errors reset the flow as shown in the diagrams, and use the code `issuance-abandoned`; more exotic errors (e.g., server crashed at Issuer headquarters in the middle of a workflow) may have different codes but still cause the flow to be abandoned in the same way. That is, in this version of the protocol, all errors cause the state of both parties (the sender and the receiver of the `problem-report`) to revert to `null` (meaning it is no longer engaged in the protocol at all). Future versions of the protocol may allow more granular choices (e.g., requesting and receiving a (re-)send of the `issue-credential` message if the Holder times out while waiting in the `request-sent` state).
```
swcurran: I see that list is the happy path set of states, but I think limiting it to just Happy Path states is not needed.  I think there should be a formally named non-happy path terminal state in the list - e.g. "issuance-abandoned". Or is there a reason not to make the list include such states?

To clarify, the lists should be all states we expect to use, and the reference to "happy path" in introducing the list should be changed.

features/0037-present-proof/README.md
```diff
@@ -35,6 +35,32 @@ Diagrams in this protocol were made in draw.io. To make changes:
...
+* proposal-sent
+* presentation-sent
+* reject-sent
+* done
+
+For the most part, these states map onto the transitions shown in the choreography diagram in obvious ways. However, a few subtleties are worth highlighting:
+
+* The final transitions for the Verifier (Send Presentation Reject, Send Presentation Ack, and Receive Presentation Reject) all result in a final `done` state, but this `done` may or may not be the Verifier's intended outcome. We may want to tweak this in a future version of the protocol.
+* A similar situation exists with the final transitions for the Prover and its final transitions (Send Presentation Reject, Receive Presentation Ack, and Receive Presentation Reject).
+* When a Prover makes a (counter-)proposal, it transitions to the `proposal-sent` state. This state is only present by implication in the choreography diagram; it essentially equates to the null or begin state in that the Prover does nothing until a presentation request arrives, triggering the leftmost transition for the Prover.
```
swcurran: I believe an intended feature of the "proposal" message is that the prover can initiate the protocol using that message. In that case, the last bullet point would not apply, and there would be both "proposal-sent" and "proposal-received" states.

concepts/0003-protocols/template.md
dhh1128: I'll fix these. However, I'd like to point out that I didn't introduce the links to HIPEs; they were present in the doc I was editing, and I just left them in. Unfortunately, the github diff makes them look like I added them.

I wonder if there's a good way to find other examples of this problem, so we can gradually drive it out of the corpus...

swcurran: @dhh1128 understand about the HIPE links being there already.

I'm wondering if these changes should go into the new 0000-template-protocol.md and we delete this file?  I was going to go back and fix these, but I think we can just eliminate the file.

What do you think?

dhh1128: In my latest push into this PR, I have largely removed the old doc. I kept a stub so we could retain the diagram of structure, but everything else is gone. Also, all references to it have been shifted. I had to move some more detail about adopted messages into the template. Tell me what you think, @swcurran .

concepts/0003-protocols/template.md
dhh1128: fixed

dhh1128:
@swcurran : I think I've now resolved all of your requested changes, except for a non-happy-path state. Not sure what to do with that. Do you have suggestions?

swcurran:
Sorry for the delay in merging this.

PR #413, opened 2020-02-06
ToIP RFC update

talltree:
Incorporated feedback and comments on the original RFC. Added a section on governance framework structure. Modified the last section to make it more general and inclusive.


dhh1128:
I don't think anything in this is controversial; it consists of some wordsmithing, some improved references, and some clarifications.

PR #414, opened 2020-02-07
added single message pickup feature

dgunseli:
Added single message pickup feature for querying messages with a message_id value. This feature might be useful for push notification communication because of the agent message sizes can be exceed data limits for push notifications very easily.


dhh1128:
I like the latest update.

@TelegramSam , as a person who's advocated pickup often, what feedback do you have?

dgunseli:
Should there be any "register" function? I guess all the messages of the all users should not be stored and provided by this function. User should request for registering themselves for using message pickup feature.

features/0212-pickup/README.md
```diff
@@ -70,7 +70,57 @@ A message that contains multiple waiting messages.
 {
     "@id": "123456781",
     "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/messagepickup/1.0/batch",
-    "messages": []
```
TelegramSam: How do you feel about `messages~attach` as the attribute/attachment method?

features/0212-pickup/README.md
```diff
@@ -70,7 +70,57 @@ A message that contains multiple waiting messages.
...
+}
+```
+### Message Query With Message Id List
+A request to read single or multiple messages with a message message id array.
+```json=
+{
+    "@id": "123456781",
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/messagepickup/1.0/single_pickup",
+    "message_id": "06ca25f6-d3c5-48ac-8eee-1a9e29120c31"
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/messagepickup/1.0/list_pickup",
```
TelegramSam: This example is malformed. It seems like you intended to include two examples that got combined by mistake.

TelegramSam:
> @TelegramSam and @dhh1128 , should there be any "register" function? I guess all the messages of the all users should not be stored and provided by this function. User should request for registering themselves for using message pickup feature.

Users can only pick up messages intended for them. This is verified either by the key used for the pickup request, or via keys registered for routing as part of that protocol.

features/0212-pickup/README.md
dhh1128: +1 to @TelegramSam 's suggestion.

dgunseli:
> > @TelegramSam and @dhh1128 , should there be any "register" function? I guess all the messages of the all users should not be stored and provided by this function. User should request for registering themselves for using message pickup feature.
> 
> Users can only pick up messages intended for them. This is verified either by the key used for the pickup request, or via keys registered for routing as part of that protocol.

I mean, user will be able to query their stored messages with id or gather with batch operation. But, as a mediator, system shouldn't store messages of all the users. I think each user should register themselves for pickup feature and then the system should start to store their messages and provide with pickup feature.

TelegramSam:
The registration option seems well covered by [RFC 0211](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md). Was there something different than you imagined?

dgunseli:
@TelegramSam and @dhh1128 ,
Should there be any handshaking protocol? Let's assume this scenario:

Edge agent queries messages with batch query and mediator returns messages but an error occurs while edge agent is processing some of the messages. When edge agent queries its messages again with batch query, messages whose process not completed by the edge agent should be return by the mediator. So, edge agent should send "ack" after it completes process each messages.

dhh1128:
>Should there be any handshaking protocol? Let's assume this scenario...

You are now describing something that is very, very close to the IMAP protocol. I think we could learn some things from it. It explicitly separates pickup from the release-from-server-storage process. That is, you fetch messages. Entirely separate from that, you can delete messages held on an IMAP server. You might delete messages you've fetched, or messages you've not fetched. They are just two separate operations.

Although I think the above may be useful input, I confess that as I see us inventing something so similar to IMAP, I am becoming uncomfortable about the feature. Maybe we should abandon it entirely, and simply tell agents who want this behavior to use IMAP as their transport. All this complexity would vanish, and there would be a lot less code to write.

dgunseli:
> > Should there be any handshaking protocol? Let's assume this scenario...
> 
> You are now describing something that is very, very close to the IMAP protocol. I think we could learn some things from it. It explicitly separates pickup from the release-from-server-storage process. That is, you fetch messages. Entirely separate from that, you can delete messages held on an IMAP server. You might delete messages you've fetched, or messages you've not fetched. They are just two separate operations.
> 
> Although I think the above may be useful input, I confess that as I see us inventing something so similar to IMAP, I am becoming uncomfortable about the feature. Maybe we should abandon it entirely, and simply tell agents who want this behavior to use IMAP as their transport. All this complexity would vanish, and there would be a lot less code to write.

You right @dhh1128, it is close to IMAP, but defining a standard for message storing and forwarding  mechanism would be very useful for mediator approach and also this approach seems like easy for integrating edge agents, isn't it?

dhh1128:
@dgunseli : : What would you like to do about this PR? I am trying not to let it languish without attention...

denizgunseli:
@dhh1128 and @TelegramSam,

All requests were done. Can you review the changes?

dhh1128:
I am okay with merging this. @TelegramSam ?

TelegramSam:
@dhh1128 can you approve the requested changes?

PR #415, opened 2020-02-07
cloud agent mediator for a mobile edge agent

sukalpomitra:
Signed-off-by: Sukalpo Mitra <sukalpomitra@gmail.com>


swcurran:
Resubmitted to address DCO issues.  Key comment from that previous #402 PR from @dhh1128:

> It is super exciting to see that you're experimenting with this use case, Sukal -- and that you're having some good success!
> 
> I think this idea does need to be documented. However, I think it needs to be connected in some way to the DID exchange protocol and to the RFC where the current forward message is described. See my inline comments.

swcurran:
I think this has some good stuff in it, but I think it goes too far. It appears to me that there is an assumption that Bob and Alice are going to use the same mediator mechanism. So, when Alice is preparing the message it gets to Bob because he also uses the same mediator mechanism. I don't think we should assume that all mobile agents and their mediators will operate the same way.  Am I misinterpreting this?

Rather, I think we want to assume that Alice and Bob are operating independently, and that Bob tells Alice in the did:peer DIDDoc he provides how messages are to be delivered to him - one endpoint, one receipientKey and 0 or more routingKeys.

I do like the idea that Alice's mobile agent could in a standard way connect to a mobile agent mediator service that holds messages for her rather than requiring that the mobile agent and the mediator be from a single vendor.

sukalpomitra:
@swcurran I think you have misinterpreted my point. My intention was to showcase the mediator works if Bob has a cloud agent of his own (if he is a mobile agent), can be a different cloud agent mediator or the same mediator used by Alice or has a service endpoint of his own (web agent).

PR #434, opened 2020-02-27
New Out of Band RFC a protocol for messages sent when no DIDComm connection is available

swcurran:


features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
+# Aries RFC 0434: Out of Band Protocol 1.0
+
+- Authors: [Ryan West](ryan.west@sovrin.org), [Daniel Bluhm](daniel.bluhm@sovrin.org), Matthew Hailstone, [Stephen Curran](swcurran@cloudcompass.ca), [Sam Curren](sam@sovrin.org), [George Aristy](george.aristy@securekey.com)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-03-01
+- Status Note: This RFC extracts the `invitation` messages from the [DID Exchange](../../features/0023-did-exchange/README.md) protocol (and perhaps [Connection](../../features/0160-connection-protocol/README.md)), and replaces the combined `present_proof/1.0/request` combined with the `~service` decorator to define an ephemeral (connection-less) challenge.
```
llorllale: nit: this note looks like it belongs in the summary section

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+- Supersedes: Invitation Message in [0160-Connections](https://github.com/hyperledger/aries-rfcs/blob/9b0aaa39df7e8bd434126c4b33c097aae78d65bf/features/0160-connection-protocol/README.md#0-invitation-to-connect) and Invitation Message in [0023-DID-Exchange](https://github.com/hyperledger/aries-rfcs/blob/9b0aaa39df7e8bd434126c4b33c097aae78d65bf/features/0023-did-exchange/README.md#0-invitation-to-exchange).
+- Start Date: 2020-03-01
+- Tags: [feature](/tags.md#feature), [protocol](/tags.md#protocol)
+- Protocol Name: Out of Band
+- Version: 1.0
+- URI: `http://didcomm.org/out_of_band/VER/MsgType`
```
llorllale: I believe we're using hypens in protocol names

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+The agent that generates the out of band message and makes it available to the other party.
+
+#### receiver
+
+The agent that receives the out of band message and decides how to respond. There is no out of band protocol message with which the receiver will respond. Rather, if they respond, they will use a message from another protocol that the sender will understand.
+
+### States
+
+The state machines for the sender and receiver are a bit odd for the out of band protocol because of it being a single message protocol with a response from a different protocol. In the following state machine diagrams we generically describe the response message from the *sender* as being a DIDComm message without associating the message with a specific protocol.
```
llorllale:
```suggestion
The state machines for the sender and receiver are a bit odd for the out of band protocol because of it being a single message protocol with a response from a different protocol. In the following state machine diagrams we generically describe the response message from the *receiver* as being a DIDComm message without associating the message with a specific protocol.
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+We need the ability to reuse a connection.
+
+### Connection Establishment Versioning
+
+In the existing Connections and DID Exchange `invitation` handling, the _inviter_ dictates what connection establishment protocol all _invitee_'s will use. A more sustainable approach is for the _inviter_ to offer the _invitee_ a list of support protocols and allow the _invitee_ to use one that it supports. This pattern enables the community deprecation of protocol versions.
+
+### Handling of all Out of Band Messages
+
+We currently have two sets of out of band messages that cannot be delivered via DIDComm because there is no channel. We'd like to align those message into a single "out of band" protocol so that their handling can be centralized, and a common QR code handling mechanism can be used.
```
llorllale: > We currently have two sets of out of band messages that cannot be delivered via DIDComm

We should name those. I'm aware of the invitation. What's the other one?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+While the _receiver_ is expected to respond with an initiating message from the chosen protocol using a specified service, the receiver may respond by reusing an existing connection. Specifically, if a connection they have was created from an out of band `invitation` from the same public DID of a new `invitation`, the receiver **SHOULD** use the existing connection in responding to the `invitation`. The selected `protocol` being used by the receiver may have a specific message type to use for this purpose.
+
+> **To Do**: Can we expand this to be able to reuse an inline `service` entry?
+> **To Do**: Update the `did-exchange` (and perhaps `connection`) protocol to have a `continue` message to be used for this purpose.
+
+#### Message Type: `https://didcomm.org/outofband/VER/request`
+
+```jsonc
+{
+  "@type": "https://didcomm.org/outofband/VER/request",
```
llorllale:
```suggestion
  "@type": "https://didcomm.org/outofband/1.0/request",
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+As mentioned in the description, the `service` item array is intended to be analogous to the `service` block of a DIDDoc. There are two forms of entries in the `service` item array:
+
+- a public DID that is resolved to retrieve it's DIDDoc service block, and
+- an inline service block.
+
+The following is an example of a two entry array, one of each form:
+
+```jsonc
+{
+  "@type": "https://didcomm.org/outofband/VER/invitation",
```
llorllale:
```suggestion
  "@type": "https://didcomm.org/outofband/1.0/invitation",
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+- `service` - an item that is the equivalent of the service block of a DIDDoc that the receiver is to use in responding to the message. Additional details below.
+
+While the _receiver_ is expected to respond with an initiating message from the chosen protocol using a specified service, the receiver may respond by reusing an existing connection. Specifically, if a connection they have was created from an out of band `invitation` from the same public DID of a new `invitation`, the receiver **SHOULD** use the existing connection in responding to the `invitation`. The selected `protocol` being used by the receiver may have a specific message type to use for this purpose.
+
+> **To Do**: Can we expand this to be able to reuse an inline `service` entry?
+> **To Do**: Update the `did-exchange` (and perhaps `connection`) protocol to have a `continue` message to be used for this purpose.
+
+#### Message Type: `https://didcomm.org/outofband/VER/request`
```
llorllale:
```suggestion
#### Message Type: `https://didcomm.org/outofband/1.0/request`
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+  "service": ["did:sov:LjgpST2rjsoxYegQDRm7EL"]
+}
+```
+
+The items in the message are:
+
+- `@type` - the DIDComm message type
+- `@id` - the unique ID of the message. The ID should be used as the **parent** thread ID (`pthid`) for the response message, rather than the more common thread ID (`thid`) of the response message. This enables multiple uses of a single `invitation` message.
+- `label` - a self-attested string that the receiver may want to display to the user.
+- `protocols` - array of protocols in the order of preference of the sender that the receiver can use in responding to the message. As implied by the message type of invitation, these are not arbitrary protocols but rather protocols that result in the establishment of a connection.
```
llorllale: > in the order of preference

Cannot rely on ordering across all JSON libraries. Suggest we punt this priority ordering?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+```jsonc
+{
+  "@type": "https://didcomm.org/outofband/VER/invitation",
+  "@id": "<id used for context as pthid>",
+  "label": "Faber College"
+  "protocols": ["https://didcomm.org/did-exchange/1.0"]
+  "service": [
+      {
+        "id": "#inline"
+        "type": "didcomm",
```
llorllale:
```suggestion
        "type": "did-communication",
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+        "serviceEndpoint": "https://example.com:5000"
+      },
+      "did:sov:LjgpST2rjsoxYegQDRm7EL"
+  ]
+}
+```
+
+The processing rules for the service block are:
+
+- Use only entries where the `type` is equal to `didcomm`.
```
llorllale:
```suggestion
- Use only entries where the `type` is equal to `did-communication`.
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+The service endpoint used to transmit the response is either present in the out of band message or available in the DID Document of a presented DID. If the endpoint is itself a DID, the `serviceEndpoint` in the DIDDoc of the resolved DID **MUST** be a URI, and the `recipientKeys` must contain a single key. That key is appended to the end of the list of `routingKeys` for processing. For more information about message forwarding and routing, see [RFC 0094 Cross Domain Messaging](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0094-cross-domain-messaging).
+
+### Adoption Messages
+
+The `problem_report` message **MAY** be adopted by the out of band protocol if the agent wants to respond with problem reports to invalid messages, such as attempting to reuse a single-use invitation.
+
+### Constraints
+
+An existing connection can only be reused based on a public DID in an out of banc message. It would be nice to have a way to be able to reuse an existing connection to the same entity even if there is no public DID in the out of band message.
```
llorllale:
```suggestion
An existing connection can only be reused based on a public DID in an out of band message. It would be nice to have a way to be able to reuse an existing connection to the same entity even if there is no public DID in the out of band message.
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+#### Example Out of Band Message Encoding
+
+Invitation:
+
+```json
+{
+  "@type": "https://didcomm.org/outofband/1.0/invitation",
+  "@id": "69212a3a-d068-4f9d-a2dd-4741bca89af3",
+  "label": "Alice",
+  "protocols": ["https://didcomm.org/did-exchange/1.0", https://didcomm.org/connections/1.0]
```
llorllale:
```suggestion
  "protocols": ["https://didcomm.org/did-exchange/1.0", "https://didcomm.org/connections/1.0"],
```

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+![Example QR Code](ExampleQRCode.png)
+
+#### URL Shortening
+
+It seems inevitable that the length of some out of band message will be too long to produce a useable QR code. Techniques to avoid unusable QR codes have been presented above, including using attachment links for requests, minimizing the routing of the response and eliminating unnecessary whitespace in the JSON. However, at some point an _sender_ may need generate a very long URL. In that case, a URL shortner redirection should be implemented by the sender as follows:
+
+- the sender should generate and track a GUID for the out of band message URL
+- The `c_i` field in the URL should replaced with the GUID, as follows:
+  - `http://example.com/ssi?c_i=5f0e3ffb-3f92-4648-9868-0d6f8889e6f3`
+- If a GUID is found in the `c_i` parameter, do an HTTP GET to retrieve the associated, encoded out of band message
```
llorllale: Suggest we use a different parameter name.

features/0434-outofband/README.md
swcurran: For transition things, I like to leave them out of the main narrative since eventually this is "normal" and new people reading it don't to know how things used to be. That opinion is a nit as well :-)

Willing to change, but that's the reason I put it there.

features/0434-outofband/README.md
swcurran: Fixed

features/0434-outofband/README.md
swcurran: Fixed.  Doh.

features/0434-outofband/README.md
swcurran: It's talked about in the document - the so-called ephemeral challenge.  That is presentation-request + ~service decorator.  BC Gov has a bunch of verifiable-credentials for OIDC uses for that.

features/0434-outofband/README.md
swcurran: We've talked about only putting the version in the header so there is no confusion about what is the current version.

Are you OK with that, or do we include the actual version in the JSON examples?

features/0434-outofband/README.md
swcurran: Good point. I think I'm OK with dropping the ordering/priority requirement, as the community update process should allow for the deprecating protocols.

Others have a comment on this?  @andrewwhitehead @TelegramSam @KitHat ?

features/0434-outofband/README.md
swcurran: Arrgghhh... :-)  Thank you! Fixed.

features/0434-outofband/README.md
swcurran: Fixed

features/0434-outofband/README.md
llorllale: Right - I forgot to double back to this comment when I reached the end.

features/0434-outofband/README.md
swcurran: Fixed

features/0434-outofband/README.md
llorllale: Oh - I wasn't aware of that discussion.

features/0434-outofband/README.md
swcurran: Fixed all the way through.  As well "did-exchange" should have been "didexchange", so that was changed as well.  Regenerated the rest of the examples based on the new message.

features/0434-outofband/README.md
swcurran: I didn't see a copy later in the RFC, but assume we would similarly drop the ordering of references related to the DIDComm service entries?

features/0434-outofband/README.md
swcurran: Good point. I was kind of testing this out to see what people think.

@dhh1128 @TelegramSam - what do you think of using this approach to version, the current version number is only in the header?

swcurran: After this, we should think about adjusting the protocol template further.  I think it is better, but not perfect.  I definitely didn't like the messages and state sections being in both the tutorial and the reference.  I think we should stick with one, or provide better guidance on what puts in one or the other.

features/0434-outofband/README.md
andrewwhitehead: The priority order is only a suggestion to the receiver on which protocol or service to choose. In the unlikely case where the elements are reordered there shouldn't be any negative impact.

That said, it might be an improvement to allow an object representation for protocols as well as the string representation, so that additional properties can be supported in future. Maybe just `{"name": "<protocol name>"}`.

features/0434-outofband/README.md
dhh1128: I'm okay with a placeholder, but I don't think "VER" in all caps is quite good enough. The reason is that this just looks like another URL segment that happens to be capitalized. Could we use a delimiter that would have to be URL-encoded, so no person could believe that what they're reading is an actual URL? Something like `%VER`, since we know `%` can't appear followed by `VER` in a valid URL?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+A common pattern we have seen in the early days of Aries agents is a user with a browser getting to a point where a connection is needed between the website's (enterprise) agent and the user's mobile agent. A QR invitation is displayed, scanned and a protocol is executed to establish a connection. Life is good!
+
+However, with the current invitation processes, when the same user returns to the same page, the same process is executed (QR code, scan, etc.) and a new connection is created between the two agents. There is no way for the user's agent to say "Hey, I've already got a connection with you. Let's use that one!"
+
+We need the ability to reuse a connection.
+
+### Connection Establishment Versioning
+
+In the existing Connections and DID Exchange `invitation` handling, the _inviter_ dictates what connection establishment protocol all _invitee_'s will use. A more sustainable approach is for the _inviter_ to offer the _invitee_ a list of support protocols and allow the _invitee_ to use one that it supports. This pattern enables the community deprecation of protocol versions.
```
dhh1128: I think this paragraph is out of alignment with the description of protocol version negotiation in https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/semver.md#recipient-rules. It has never been the case that inviters dictate versions as far as the specs are concerned; it's just the case that nobody in the community has implemented the version negotiation logic we already described.

If you believe what I am asserting, then the value of this as a rationale is lower, and the advisability of sending a list of supported protocols is questionable.

features/0434-outofband/README.md
dhh1128: "so that their handling can be centralized" --> "so that their handling can be harmonized inside an agent" (I'm trying to avoid unjustified negative reactions to the word "centralized", which is perfectly innocent as used here, but which might raise eyebrows from people who don't know the context)

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+### URLs and QR Code Handling
+
+We'd like to have the specification of QR handling centralized into a single RFC (this one).
+
+## Tutorial
+
+### Key Concepts
+
+The out of band protocol is used when you don't know if you have a connection with another user. This could be because you are trying to establish a new connection with that agent, you have connections but don't know who the other party is, or if you want to have a connection-less interaction. Since there is no DIDComm connection to use for the messages of this protocol, the messages are plaintext and sent out of band, such as via a QR code. Since the delivery of out of band messages will often be via QR codes, this RFC also covers the use of QR codes.
```
dhh1128: Can we give one other example besides QR codes, so it's clear that's not the only use case? How about email? ("such as via email or a QR code")

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+### States
+
+The state machines for the sender and receiver are a bit odd for the out of band protocol because of it being a single message protocol with a response from a different protocol. In the following state machine diagrams we generically describe the response message from the *receiver* as being a DIDComm message without associating the message with a specific protocol.
+
+The *sender* state machine is as follows:
+
+[![sender state machine google doc](state-machine-sender.png)](https://docs.google.com/spreadsheets/d/1o1szGCMx6FzrxfAKV76IEYmeKRJqvV3gd1r3Qfv1Q9o/edit#gid=1176419697)
+
+Note the "optional" reference under the second event in the `await-response` state. That is to indicate that an out of band message might be a single use message with a transition to done, or reusable message (received by many receivers) with a transition back to `await-response`.
```
dhh1128: Do we need to worry at all about the possibility of a *resend*?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+There are two messages in the out of band protocol. They are both sent by the *sender*. Only one of the message types is ever used in an instance of the protocol.
+
+#### Message Type: `https://didcomm.org/outofband/VER/invitation`
+
+```jsonc
+{
+  "@type": "https://didcomm.org/outofband/VER/invitation",
+  "@id": "<id used for context as pthid>",
+  "label": "Faber College",
+  "protocols": [
```
dhh1128: I'm not sure this is a good idea. See my comment about version negotiation.

Now, having said that, I don't think version negotiation totally solves the problem you're attacking here. It's only about an 80% overlap. Specifically, it doesn't handle the case where a version is renamed. So we may need what you are proposing. I am just noting that I think we need to ponder the similarities in these two mechanisms, and do something to resolve them.

features/0434-outofband/README.md
dhh1128: Re. priority order: this is one of the obnoxious things about JSON-LD, IMO: it explicitly contradicts ordering rules in JSON. JSON parsers are required to maintain order of JSON arrays, or they're not JSON-compliant. So I don't actually think George's concern is likely to occur. Do we know for a fact that it does? Having said that, a JSON-LD parse that reads DIDComm messages would probably make the very mistake he predicts, so maybe it's still an issue.

On a different subject: I don't think the invitation with protocol names is enough info. We need an invitation to specify both the name of the protocol, and the name of the role in the protocol that the inviter proposes the invitee to take. In both of the protocols we're naming here, the role implied for the invitee is `requester`, but there's no guarantee that future connection-like protocols will work with that implicit assumption.

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+- `label` - a self-attested string that the receiver may want to display to the user.
+- `protocols` - array of protocols in the order of preference of the sender that the receiver can use in responding to the message. As implied by the message type of invitation, these are not arbitrary protocols but rather protocols that result in the establishment of a connection.
+- `service` - an item that is the equivalent of the service block of a DIDDoc that the receiver is to use in responding to the message. Additional details below.
+
+While the _receiver_ is expected to respond with an initiating message from the chosen protocol using a specified service, the receiver may respond by reusing an existing connection. Specifically, if a connection they have was created from an out of band `invitation` from the same public DID of a new `invitation`, the receiver **SHOULD** use the existing connection in responding to the `invitation`. The selected `protocol` being used by the receiver may have a specific message type to use for this purpose.
```
dhh1128: We've been using the word "recipient" instead of "receiver" for protocols. That's because recipients are usually people, whereas receivers are usually pieces of tech. (In phone tech, the receiver is hardware, the recipient of a call is a person. The recipient of an email message is Fred; the receiver is usually his MTA.) Maybe that distinction is not super important; this isn't a big deal to me, but I wanted to point it out in case you think it's worth doing something about.

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+When considering routing and options for out of band messages, keep in mind that the more detail in the message, the longer the URL will be and (if used) the more dense (and harder to scan) the QR code will be.
+
+###### Service Endpoint
+
+The service endpoint used to transmit the response is either present in the out of band message or available in the DID Document of a presented DID. If the endpoint is itself a DID, the `serviceEndpoint` in the DIDDoc of the resolved DID **MUST** be a URI, and the `recipientKeys` must contain a single key. That key is appended to the end of the list of `routingKeys` for processing. For more information about message forwarding and routing, see [RFC 0094 Cross Domain Messaging](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0094-cross-domain-messaging).
+
+### Adoption Messages
+
+The `problem_report` message **MAY** be adopted by the out of band protocol if the agent wants to respond with problem reports to invalid messages, such as attempting to reuse a single-use invitation.
```
dhh1128: I don't see how problem-report messages could be used if you don't yet have a connection. How would you send them? Over QR codes?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+[![receiver state machine google doc](state-machine-receiver.png)](https://docs.google.com/spreadsheets/d/1o1szGCMx6FzrxfAKV76IEYmeKRJqvV3gd1r3Qfv1Q9o/edit#gid=10874521)
+
+Worth noting is the first event of the `done` state, where the receiver may receive the message multiple times. This represents, for example, an agent returning to the same website and being greeted with the same QR code each time.
+
+### Messages
+
+There are two messages in the out of band protocol. They are both sent by the *sender*. Only one of the message types is ever used in an instance of the protocol.
+
+#### Message Type: `https://didcomm.org/outofband/VER/invitation`
```
dhh1128: I feel like we are missing out on an opportunity to generalize a little bit further. We have here an invitation message, but we are assuming that the invitation is always an invitation to connect. Why couldn't it be an invitation to do X, Y, or Z instead, where connecting is a precondition? We could do this by having a new property telling what the goal of the invitation is:

```jsonc
{
    // all the stuff that's already in an invitation goes here, including the existing
    // protocols field. Then we add:

    "goal": [
        // Each item in the "goal" array is an alternative "weighted protocol start spec":
        //     protocol = the protocol to start
        //     your_role = named role the recipient should take
        //     preference = value between 0 and 1, expressing how preferred the option is.
        //         This allows two options to be weighted the same from the inviter's perspective.
        {"protocol": "https://didcomm.org/issue-credential/2.0", "your_role": "issuer", "preference": 0.8}
        {"protocol": "https://didcomm.org/issue-credential/1.3", "your_role": "issuer", "preference": 0.5}
        {"protocol": "https://didcomm.org/cred-iss/0.9", "your_role": "creator", "preference": 0.2}
    ]
}
```
The recipient decides whether to connect based on whether they agree to the goal, which is better than just connecting for no reason (gives the recipient better reason to filter spam). And the recipient can see which way the goal will be accomplished, so they'll know whether the connection is DOA because they can't accomplish the goal. Also, now you have a way to explicitly tell the recipient to start the protocol, or to take a passive role and let the inviter start the protocol as soon as the connection is established. You also have a way to start protocols with more than 2 roles (e.g., auctioneer, bidder, referee).

If we also made the `protocols` array into a `connect` array, and gave it the same structure, we could solve the prioritization problem, and we get the flexibility to do funkier connection types (e.g., let's connect by asking Alice to introduce us). Without this, you can set up a connection, but you're hardwired into the assumption that the invitee is the requester, and into the assumption that after the connection is set up, the inviter will make the first move in the subsequent protocol.

features/0434-outofband/README.md
swcurran: Fixed

features/0434-outofband/README.md
swcurran: Changed. If we like that, we'll add that to the protocol template.

features/0434-outofband/README.md
swcurran: My vote is to leave the text about the ordering as in the unlikely event there is a re-ordering, there is no harm done, since any are by definition OK to use.

I disagree with the need for the role name to be included.  By definition, the recipient will be the initiator of whatever protocol is selected.  Further, an "invitation" is by definition being used to create a connection, so there is little need to specify the role. It would also increase the size of the invitation, making URL shortening more likely to be required.

From the others, what about Andrew's suggestion about `{"name": "<protocol name>"}`.

dhh1128: I disagree with the analysis about the role name; see my longer comment further down. Specifically, I don't agree that an invitation should by definition be viewed as being used to create a connection. Spammers would love that. We need to make the goal a lot more explicit than just "hey, let's connect."

features/0434-outofband/README.md
swcurran: I think we would all like to reduce the instances of negotiation in the connection establishment process. We could support it, but avoiding it would be MUCH better. Negotiation of protocols once a connection is established is perfectly reasonable, but I think it's a bad idea on creating the connection.

features/0434-outofband/README.md
swcurran: I'll think about that. I'm not certain what the email would look like :-). I guess just 

Dear Daniel,

Click [here](http://example.com/ssi?c_i=eyAiQHR5cGUiOiAiaHR0cHM6Ly9kaWRjb21tLm9yZy9vdXRvZmJhbmQvMS4wL2ludml0YXRpb24iLCAiQGlkIjogIjY5MjEyYTNhLWQwNjgtNGY5ZC1hMmRkLTQ3NDFiY2E4OWFmMyIsICJsYWJlbCI6ICJBbGljZSIsICJwcm90b2NvbHMiOiBbImh0dHBzOi8vZGlkY29tbS5vcmcvZGlkZXhjaGFuZ2UvMS4wIiwgImh0dHBzOi8vZGlkY29tbS5vcmcvY29ubmVjdGlvbnMvMS4wIl0sICJzZXJ2aWNlIjogWyJkaWQ6c292OkxqZ3BTVDJyanNveFllZ1FEUm03RUwiXSB9) to connect.

Click here to get an agent...

```

File: features/0434-outofband/README.md [hunk 21]
```diff
@@ -0,0 +1,417 @@
...
+While the _receiver_ is expected to respond with an initiating message from the chosen protocol using a specified service, the receiver may respond by reusing an existing connection. Specifically, if a connection they have was created from an out of band `invitation` from the same public DID of a new `invitation`, the receiver **SHOULD** use the existing connection in responding to the `invitation`. The selected `protocol` being used by the receiver may have a specific message type to use for this purpose.
+
+> **To Do**: Can we expand this to be able to reuse an inline `service` entry?
+> **To Do**: Update the `did-exchange` (and perhaps `connection`) protocol to have a `continue` message to be used for this purpose.
+
+#### Message Type: `https://didcomm.org/outofband/VER/request`
+
+```jsonc
+{
+  "@type": "https://didcomm.org/outofband/VER/request",
```
dhh1128: After reading the doc, I'm puzzled about the difference between these two message types. Do we really need both of them. Aren't they doing almost the same thing, except that the request message attaches some stuff? Is that difference worth a whole separate message?

A possible answer to my question might be: no, they are different. The invitation just asks to connect, whereas the request message can do arbitrary stuff. But if that's true, why shouldn't we use the arbitrary stuff to do connections? I can already attach things to an invitation anyway, since it's a didcomm message. Right?

features/0434-outofband/README.md
dhh1128: Please go read what was written. It's not a back-and-forth negotiation. It's a "here's the version of the protocol I want to use, as embodied by the first message I send" followed by "here's the second message in the protocol, but notice that i've degraded the version, so you're stuck with what I picked" sort of negotiation. It's not chatty at all. It introduces zero negotiation messages.

features/0434-outofband/README.md
swcurran: I tend to agree, but in the case mentioned (all the requirements met), they have sent you their DIDDoc for the relationship, and in that case you COULD respond with a DIDComm message.  Again - it's a pretty limited case.

dhh1128: Okay. Thanks for the clarification. Point withdrawn.

features/0434-outofband/README.md
swcurran: Fixed per above

features/0434-outofband/README.md
swcurran: Fixed per above

features/0434-outofband/README.md
swcurran: In the existing RFCs, the inviter picks the invitation to send - either Connection or DID Exchange, since they can only send one. By taking "invitation" out of those, the receiver can respond to either during the next transition. After that, the value goes away, particularly if there is only version changes going forward.

I think you are proposing that in the Out of Band, invitation we should not specify any protocol and the receiver can respond with any protocol--presumably limited to those meant to establish a connection?  I think if we drop it, we have to be very clear in the RFC on the intention.

It does not feel right to me, but what do others think?

While as you say, the "value is reduced", I think it is still worth doing, but it does come at the expense of the length of the message (and hence, more likely URL shortening).

features/0434-outofband/README.md
swcurran: Added.  Also fixed the "centralized" reference. There was another that I changed as well.

features/0434-outofband/README.md
swcurran: I don't understand the context of the question. Could you please clarify?

features/0434-outofband/README.md
swcurran: I'm guessing you mean the "protocols" field?  If so, I think we're covering the conversation above.

features/0434-outofband/README.md
swcurran: I tried a few things and my preference is receiver.

Others?

features/0434-outofband/README.md
swcurran: We discussed this and decided while potentially useful, it was more complex than necessary.  It was in a first cut, but dropped during a discussion about the resulting benefits.

The purpose of the sender getting the "pthid" of the follow on message is so that they can know the context of the invitation that they initiated for use after the connection is established.  With that information, they can initiate the follow on action, whatever that might be. In many cases - particularly in a website/QR code scenario, both parties will know what will come next.

Further, by keeping that complexity out of the message, we keep the out of band message smaller.

features/0434-outofband/README.md
swcurran: I don't think it is cut'n'dry that two are needed, but my opinion is that the two are different enough that it's worth having both.

Very little information is needed is needed in the `invitation` message because it's purpose is known; quite a lot is needed in the first `request` example we have - `presentation-request`, especially if we are going to support multiple credential exchange models. I expect other uses to likely be equally bulky.  I can't think of what the attachment would be contained in the `invitation` case, so it just wouldn't have it.  I think we want to make it very obvious to the controller dev if a connection is to be established or not as an outcome, and having the two messages accomplish that.

I think the benefits are there for the two.  Others?

features/0434-outofband/README.md
llorllale: I don't see an issue with listing the supported *protocols* in the out-of-band invitation. I don't see an issue with including the supported *versions* as well, but I don't see that as a critical point.

I think what @dhh1128 is doing is attacking the assertion in the highlighted paragraph that including the *version* "enables the community to deprecate protocol versions" when in fact such a mechanism is described in RFC0003. There's a point there, but this out-of-band invitation would be just another way to achieve the same goal (deprecation of protocol versions), although solely contained to the bootstrapping phase.

swcurran: Thanks, George.  I've removed that line.

features/0434-outofband/README.md
llorllale: Ok - I somehow missed this giant point. My bad.

The main driver for this protocol on our end was to kickstart workflows *for long term relationships*. So we need did-exchange + something that describes whatever follows next. I kind of assumed the `request` message was setup for this but it appears to be tailored to ephemeral challenges?

dhh1128: Apparently I'm struggling to make my point. Let me try again. I have three reasons why I'm recommending a change.

1. Spam

    Connecting, as we have done it thus far in the Connection and DID Exchange protocols, is often a bad goal, in and of itself. I get a steady stream of "let's connect" messages on LinkedIn, and I have to sort through them and decide whether I think a connection is a good idea. Usually, the way I decide is by looking at the specifics of the sender's intentions. A message saying, "Let's connect so we can collaborate on X, as we discussed in conversation Y" is a lot better than a messaging saying "let's connect." The latter is often spam. Spam with agents will be a thing.

    What we have in the current design is a protocol for proposing connections *with no further context*. This is problematic from a spam perspective. We need to be able to specify intentions beyond connection, or else we'll just have to invent a new out of band protocol later that has that feature. Spam will demand it.

2. Who starts the next interaction

    You said "The purpose of the sender getting the `pthid` of the follow on message is so that they can know the context of the invitation that they initiated for use after the connection is established. With that information, they can initiate the follow on action..." This is all well and good. But what if the follow-on action should be begun *by the recipient*? How does Alice say to Bob, "Hey, let's connect, so *you* can bid on my Rembrandt that I'm auctioning?" Right now we only support, "Hey, let's connect, so *I* can bid on your Rembrandt that you're auctioning."

3. Deciding whether the next interaction will be possible

    Building on the previous example, suppose the next planned interaction is auction bidding. Suppose that Alice only supports auction 2.0 protocol, and Bob only supports auction 1.0 protocol. It is likely that from the recipient's perspective, there is no point in connecting if the true goal for the connection (the auction, not just connecting) will be frustrated. Therefore, the OOB protocol ought to let the sender express the eventual goal, so that can be taken into account by the recipient when evaluating the desirability of connecting.

I'm not saying that every connection must have a further goal; just having a connection might be the goal. In that case, the `goal` array I proposed would be blank. What I'm saying is that we ought to make it possible to tell the difference between a connection without a goal, and a connection with a goal. This helps with spam; it helps make the next interaction one startable by other parties besides the sender; it helps a recipient decide if the connection is a dead end.

And I'm saying that some OOB messages might not have the goal of connecting at all; it's possible that the goal is to complete a connectionless interaction, in which case the `protocol` array (perhaps better renamed `connection_protocol` array) would be blank. Then the second message type you're imagining goes away. The OOB protocol collapses to one message type that supports one additional field.

swcurran: I get your argument, but I don't find it compelling.  As I said, we did talk about this and decided it was more complicated than worthwhile.

On the spam, the invitee has to take a step to get the out of band message.  I don't know how the inviter gets the unsolicited message.  Is that via email?  Either way, the invitee has already taken a step by accepting the OOB message.

On the initiator, I think in the majority of cases the inviter will initiate the next action. If not, I think rather than making OOB invitation more complex, we add a DIDComm message that supports that.  Andrew's "Action Menu" is our WIP on that - https://hackmd.io/jrWg2kw7QuWx9XS0R3_XPQ

The deciding if the next interaction is possible is IMHO another "too complex" issue for invitation.

That all said, George is now around to your side of the argument, so we'll have to figure out how to break through this.

We really need to nail this down soon as it is a very high priority item to get moving in the community. The lack of reuse of connections is painful.

features/0434-outofband/README.md
TelegramSam: I agree that JSON preserves order and 'higher-level' parsers that don't respect that are flawed.

features/0434-outofband/README.md
TelegramSam: Still reading the conversation, but I'm a fan of listing the protocols. I think it's unlikely that we won't have the need for supporting different connection protocols.

features/0434-outofband/README.md
TelegramSam: I should mention that further _features_ don't need a new protocol... a new version will do.

On spam: The primary fighting mechanism is indeed context, but that context is provided by a pthid of an invitation. A public DID can be used to jump into didexchange directly, but that method will be highly subject to spam.

Providing the 'goals' does add some color as to purpose, but I believe having a pthid of an outofband message will be the most useful method.

features/0434-outofband/README.md
TelegramSam: Agree on two. I think the benefit of generalization is lost for common cases (such as establishing a connection) where a dedicated message is both smaller and harder to screw up.

features/0434-outofband/README.md
TelegramSam: On inviting the remote party (recipient of outofband message) to start the interaction: I think there are a few ways to do this.
1. Include a message you want the recipient to send back to you. This was suggested by Andrew, but I think very dangerous. The need to strictly validate is high.
2. Include a protocol and role. I could be convinced of this over 1. Same value, much less dangerous. I still am voting that this becomes a future version feature.
3. Use Pthid as context to prompt an interaction. Sending a BidRequest message isn't much different than requesting the recipient play the role of bidder. I think this method has the advantage of making agents easier to develop in the short term.

dhh1128: I don't understand the reasoning on a pthid providing the link. That context-building capability is only useful in retrospect; that is, you can't consider the context when you're evaluating the invitation; you can only connect it later, after the connection is made. No spam benefits. That feels helpful to me, but WAY less helpful than setting the context at the outset. I'm really pretty strongly misaligned with this. I think we're making a mistake to treat the spam problem so casually.

features/0434-outofband/README.md
dhh1128: >Very little information is needed is needed in the invitation message because it's purpose is known

This is the disconnect. I'm claiming that the *real* purpose of the invitation is not known, and that the purpose we're currently assuming (just to connect for the sake of connecting) is indicative of immaturity of our thinking as a community; random agents who want to connect just for the sake of connecting should always be refused when the ecosystem is mature. We only think that's valid today because it's what we've been using to build a narrative for our demos. The goal of just connecting is almost never the real purpose. And I'm asking to make the real purpose known (which happens to eliminate the distinction with a `request` message), and getting lots of pushback.

We can certainly merge this protocol RFC, but I can't recommend to my own company or anyone in the community that we support it. It's DOA for us unless we take the spam issue more seriously.

features/0434-outofband/README.md
swcurran: The claim I'm making is that the `pthid` provides the link to the context of the invitation in the inviter's context - why it was created. It tells the invitee nothing.  So the inviter does not need anything else.

Further, I'm claiming that the invitee knows the context of the claim before they process the invitation because the had to do something to ingest the invitation - scan it, click on it, etc.  At most, in the invitation they get at most self-attested information from the inviter - a label, and if you win, some sort of indication of the purpose of the invitation.  My view  is that I don't see the benefit of the extra (self-attested) information within the invitation as spam prevention.

features/0434-outofband/README.md
swcurran: As per above, the invitee knows the purpose based on the action of ingesting an out of band message.  Totally agree that the purpose of the invitation is not to connect, but to do something after, but adding more self-attested data such as the name of a protocol to use next does not provide more context in my opinion. 

Daniel, we *really* don't want this to be DOA/vetoed, so how do we get around this.  Could you:

* explain in very concrete term spam examples -- how you see them happening.
* exactly what you want added to the invitation message?

On your latter point ("which happens to eliminate the distinction..."), my guess is what you will propose will not eliminate the request message, but I'll wait on that until I see exactly what it is you are proposing.

Maybe we can discuss at HGF?  Are you still here on Friday?

dhh1128: @swcurran , @andrewwhitehead , @llorllale , and @TelegramSam : here's my best attempt to provide more context about my spam concern: https://docs.google.com/document/d/1S3dC9NuIYx9IRJVWu5RUG1wLNluhVy0lhIinin6fRqw/edit

TelegramSam: Daniel, thanks for that writeup. I'm thinking that clarifying that outofband messages are for human reception and contextual evaluation solves the recipient spam problem sufficient for current needs. Non-human reception can be expanded in a future version that includes greater features to prevent the issues you highlight, or in a different protocol alltogether.

Future version of the protocol can also add additional features necessary for contextual evaluation if we deem the current approach insufficient.

dhh1128: That doesn't resolve my concern. Two reasons.

1. I'm thinking this "human only" guidance is not aligned with the intentions of @tplooker WRT ephemeral use cases. Isn't he planning on using ephemeral stuff with IoT devices, immediately?

2. I don't think we get to say that protocols are only for use by people. DIDComm is designed such that agents could be backed by people or by AIs or by automation, and we can't tell the difference; everything is a peer. If there is some technical reason why the protocol *can't* be used by tech, as opposed to simple intentions that it not be, that would make me feel more cheerful.

dhh1128: And reason 3: the human flooding problem isn't solved.

features/0434-outofband/README.md
dhh1128: Right. I agree with your logic on the sender knowing the context via the pthid.

I disagree with your logic on the invitee knowing the context. I don't agree that we can assume the invitee had to take a human action to receive the OOB message. I gave several examples of how this could not be true, in my writeup about the spam concern.

features/0434-outofband/README.md
dhh1128: @TelegramSam : what would resolve my concern is helping me realize that one of my assumptions is wrong. That could totally be the case. If it is, great. But if not, then a general statement that we can tamp down the worry about spam by telling senders and recipients to be wise isn't going to help me get comfortable. I have an immediate use case for this protocol in a context where millions of people will receive unsolicited invitations. These invitations will be associated with the reputation of an institution that cares enormously about its public trust. Furthermore, once this protocol is deployed in this use case, it will be difficult to upgrade it. So the "future version can add additional features if we deem the current approach insufficient" is already a fulfilled prophecy for me. I need a better solution, now.

I also truly don't get the complexity concern. What I proposed adds one requirement: the existence of a field in a message called `goals`. The field can have no content. The recipient of such a message can ignore the content if they like; nothing in the protocol requires either party to do anything significant with the field. Anybody who doesn't want the complexity can largely ignore it in their impls. So how does this add complexity is such an obnoxious way?

TelegramSam: > So how does this add complexity is such an obnoxious way?

Goals don't add complexity, generalizing/collapsing the two message types does. 

I do find goals to be fairly useless in spam prevention, given that they can just be lies.

> And reason 3: the human flooding problem isn't solved.

I don't find this problem to be real. In the AR example (from the linked doc), I find it silly to think that Alice's AR platform would not be capable of recording when and where (the context) that an outofband message was received.

> I don't think we get to say that protocols are only for use by people.

I think we can say that _this_ protocol is only for use by people. This prevents developers from needing to receive these messages over any inbound transport.

> I have an immediate use case for this protocol in a context where millions of people will receive unsolicited invitations. These invitations will be associated with the reputation of an institution that cares enormously about its public trust.

Knowing more about this use case will help us evaluate the design of the protocol. How are the invitations received? How do goals help establish trust?

TelegramSam: > I'm thinking this "human only" guidance is not aligned with the intentions of @tplooker WRT ephemeral use cases. Isn't he planning on using ephemeral stuff with IoT devices, immediately? (Tobias, please comment. Maybe this concern is not really a thing?)

Also interested in commentary from @tplooker to verify this.

dhh1128: Please read this comment carefully, and please respond carefully. I feel like I am being misunderstood repeatedly, and that a quick response will only make things worse. I will number the items in this comment so they are easier to refer to with shorthand.

1. I am not recommending that we collapse the two message types. If you go read my comments, you will see that I'm puzzled about the `request` message. I have also pointed out in a couple verbal contexts that it's hard to justify such a message if the `invitation` message becomes richer. And I've said in this PR that the purpose of it is pretty hard to understand from the written text. But I haven't advocated anywhere in this PR stream that we get rid of `request`. I have made one concrete proposal only: that we add `goal` as a property to the `invitation` message. And the pushback I'm getting is about complexity. @TelegramSam says he's not troubled by the complexity of having a `goal` field--only about collapsing the two message types, but I think @swcurran feels that `goal` is too complex even if we keep the message types separate. Is that true? If so, why? If not, can we please accept that suggestion? I don't feel nearly as interested in collapsing message types as in getting a goal into an invitation.

2. Sam says "I think we can say that this protocol is only for use by people. This prevents developers from needing to receive these messages over any inbound transport." But the key property of agents is that they represent people in a way that parties on the other side of the interaction don't get to dictate. We don't get to require Alice to process the code with her human intelligence; she can use an AI or a dumb policy or whatever she wants.

    Also, we don't ever get to assume in DIDComm that the transport used by the sender and the transport used by the receiver are identical. As a sender, you can certainly choose the QR code transport -- but you don't get to tell Alice that she can't convert a QR code to an emailed link, or a voice text, or whatever, via a relay. (My wife has a deaf friend. She calls on the phone; a relay converts her spoken words to typed text. How will a blind person process a QR code? Etc.) All you know as a sender is that you *sent* over the QR code channel. Alice will receive it however she wants, and will process it however she wants. This protocol would (and should) work over any transport. I gave several transport examples in my doc that explored the spam problem; I claim all of them are reasonable. It is possible that a protocol could be invented that is inherently transport-specific (e.g., a protocol that lets jazz musicians jam together requiring realtime audio steaming and disallowing pigeon transport), but the "let's connect" problem is not that sort of problem, and we shouldn't do headstands to bend it into one. Agents don't have to support transports they don't like, but protocols aren't written for the convenience of a few agents that want to be less than transport agnostic.

    Also, even if you throw all of that out the window, what property of the transport are you going to modify to declare it to be a human-only transport? Just verbiage that a human reads in an RFC? If so, what happens when you want to change your mind later? Does that require a version update? If so, why, since none of the machine-inspectable properties of the protocol are different? And if not, how will you tell whether you're dealing with an agent that expects a human partner or not?

    Bottom line: saying this protocol is for humans only is a BAD idea for lots of reasons.

3. Sam says, "I do find goals to be fairly useless in spam prevention, given that they can just be lies."

    A lie is just as good as a truth here, because *what we're after is something that becomes testable quickly, not something that's guaranteed to be true*. Suppose Alice gets invitation A with no goal, invitation B with the goal of issuing a credential, and invitation C with the goal of giving her free money. Suppose B is a truthful invitation, and C is a lie. Alice responds with a connection request to each one, and gets a connection response, so she's now built a connection for all 3 invitations. What happens next? The sender of A (no goal) is silent. Alice waits. How long should she wait? Forever -- since the sender isn't violating her expectations. The sender of B (truthful invitation to issue cred) immediately starts an issuance protocol. Alice is happy about B and knows she isn't wasting her time. How about the liar who sent invitation C? If the liar is silent, Alice deletes the connection after, say, 30 seconds, since they didn't behave according to their expressed goal. If the liar now tries to do anything other than start the free money protocol (e.g., sell Alice Viagra, but also ask Alice to prove something, propose a rendezvous, send a BasicMessage, etc) Alice can also immediately delete the connection, since she now knows C is a liar. The bad outcome here isn't C (that's just a temporary annoyance, whether the party lies or not); rather, the bad outcome is A (no expressed goal). Because now Alice has allocated a DID, keys, a peer DID doc, etc. Having one of these useless connections is no big deal, but having thousands of them might become problematic.

    There will always be connections that have no declared purpose. But adding a `goal` field lets Alice filter unsolicited invitations twice. First she throws away ones with goals she either dislikes or can't accomplish. Then she quickly and effortlessly throws away lies about intention. This can be done in automated fashion, without bothering her at all, if there's a goal. If not, Alice is forced to triage them all personally.

4. Sam says, "I don't find this problem to be real. In the AR example (from the linked doc), I find it silly to think that Alice's AR platform would not be capable of recording when and where (the context) that an outofband message was received."

I feel like this kind of misses the point. Of course her AI platform can record when and where an OOB message was received. But if it's just an invitation to connect, without any goal, then there's no data for her AI to reason from, other than when and where the invitation occurred. Was that jewelry store going to offer her a 90% discount on a second set of earrings if she's a returning customer -- redeemable any time she comes by this week? She might want to know about that when she gets home. Whereas she may have no interest in the Hog Heaven Road Gear shop she walked past, since that was touting a giveway of a free Harley but it was a time-limited offer that has expired. The point is that the context includes the time and circumstances when the invitation arrived, but that gets way more useful when software knows the asserted intention of the invitation. Now policy can operate on intention.

dhh1128:
I apologize for such lengthy feedback. I feel like this PR represents a very important idea, and I'm excited that the PR has been raised. Perhaps I've engaged on it too strongly because I've been too worried about details that will prove to be less important.

I wanted to ask three additional questions that are independent of my other comments. Then it's my intention to try to stop commenting and let community consensus govern.

1. Although the two message types described in this RFC have the out-of-band character in common, I wonder if it's smart to treat them as a single protocol. If consensus is that the request message and the invitation message solve different problems, then shouldn't those solutions be versioned independently? There doesn't seem to be any sequencing or semantic progression between the two... Doesn't that make them independent protocols that just happen to have an OOB attribute in common?

2. I think the message that each of *these* messages elicits is really not "part" of the protocol here. Rather, I think what we have here in both cases is a one-message, one-way notification-style protocol that has as its purpose nothing beyond conveying info that kicks off a more normal co-protocol. The connection-request message that comes back after an invitation is really the first message in a new, independent protocol.

3. Obviously an invitation or a request to process an OOB message can be time-limited using ~timing.expires. But are we sure they must be count-limited, too? If AcmeCorp invites Bob to connect, is it always intended that he not accept more than once? Right now we are assuming so; the invitee is supposed to re-use an existing connection instead of re-connecting. But I think there might be cases where multiple connections is also a valid behavior (e.g., each connection is a new entry in a raffle). How does an inviter tell the invitee that?

features/0434-outofband/README.md
swcurran: My $0.02CDN.  I added some comments to the Google Doc that shows where I don't agree with either assumptions or the scenarios. The biggest disagreement I have with the doc is that assumption that there will be an automatic response to an OOB message. Human or machine, a (possibly automated) decision will be made between receiving and accepting the OOB message. The bottom line is that I still don't see spam as a compelling concern with respect to this protocol.

If we want to add anything to address this, I think it would be preferred to add a "purpose" text/code field (with, of course, the option of ~l10n). Like Sam, I don't think it is needed and that it is potentially misleading because it is self-asserted.  The use of "goals" list of protocols Daniel proposes gives the human no help in knowing what the initial purpose of the connection is. For a machine it does (but no different than a code), but the machine handling will be conceptually different from that of a human.

The reason I think a text field might be useful is that the "label" field will become overloaded for that purpose. I expect the label will be used as an initial value for the connection (update-able by the user), and having the initial purpose for the connection in the label will be a pain. So separating them into two fields could be useful.

swcurran:
Can't see how to reply to @dhh1128 's last note.  This is a reply to that:

1. I think the delivery mechanism (e.g. not DIDComm) is enough to put them into the same protocol. We could separate them (as stated in the RFC - the relationship is not technically necessary), but it certainly feels better to me to keep them in the same protocol. My vote is to keep them together.

2. Completely agree.  I hope that has been stated in the RFC.

3. I think there is definitely a strong case for a count limited (particularly - 1) OOB message. It is less defined on time but rather a session. But none of that matters in the protocol definition. Those are business concerns, and the controllers involved will handle them.  Count or expiration are not included as part of the messages.

dhh1128:
I am cheerful about the idea to change `goal` to be `purpose` and changing its datatype to be a code that could be used to find human-friendly, localizable text. This actually harmonizes with similar concepts that have been floated for co-protocols declaring what problem they solve, and with an "action menu" concept that I think originated with @andrewwhitehead . If the problem(s) solved by a protocol, and the items in an action menu, and the optional goal/purpose of an invitation were all expressed with the same corpus of localizable codes, I think we'd have some real goodness. It addresses enough of my spam concern to make me happy, too.

>I think the delivery mechanism (e.g. not DIDComm) is enough to put them into the same protocol.

How about we keep them in the same doc, but make them separate protocols? This would let an agent support one, but not the other--and say so in feature discovery. And if we come up with a new message type that shares the same OOB context, we can add a new section to the RFC but not have to rev the version on the messages that don't change?

>a strong case for a count limited... Those are business concerns, and the controllers involved will handle them. Count or expiration are not included as part of the messages.

Expiration can be included in the message via the timing decorator; we can't prevent it. But my interest in this topic was mainly around the count. Obviously the controller of the sender knows what business rules they intend, and can act accordingly -- but how will controller of the receiver know the sender's intent? The answer, "They won't -- but the sender will let them know if they violate expectations" works when count = 1 (the sender enforces 1) -- but it doesn't work if the intent is count > 1, because the user won't know to consider that an option.

TelegramSam:
I'm cheerful about `purpose` replacing `goal`. What details do we need to discuss about what that purpose would contain?

swcurran:
I'll update the PR today with that wording.  I think it is a single
optional field that is a string but we encourage ~l10n support.

swcurran:
I've added a `purpose-cd` that is a generic protocol-level code for the purpose for the machine use case and `purpose` that is a context specific, human-targeted string. Both can/should be localized.  I choose two strings because the code for the context-specific string converted to a code would be too specific for a machine to know what's going on, and the code suitable for a machine would be too generic for a person.

It might be overkill since it's likely the sender would know if they were dealing with a machine or a person (although not for sure), so could send an appropriate code depending on the circumstance.

Let me know what you think @llorllale @dhh1128 @TelegramSam @andrewwhitehead @KitHat 

Thanks!

features/0434-outofband/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+Both the `purpose-cd` and `purpose` fields should be used with the [localization service decorator](../0043-l10n/README.md). The two fields are to enable both human and machine handling of the out of band message. `purpose-cd` is to specify a generic, protocol level reason for the connection (e.g. issue verifiable credential, request proof, etc.) that is suitable for machine handling and possibly human display, while `purpose` provides context specific guidance, targeting mainly a person controlling the receiver's agent.
+
+While the _receiver_ is expected to respond with an initiating message from the chosen protocol using a specified service, the receiver may be able to respond by reusing an existing connection. Specifically, if a connection they have was created from an out of band `invitation` from the same public DID of a new `invitation`, the receiver **SHOULD** use the existing connection in responding to the `invitation`. The selected `protocol` being used by the receiver may have a specific message type to use for this purpose.
+
+> **To Do**: Define the list of `purpose-cd`
```
llorllale: Suggest a registry instead of a "list"

llorllale:
@swcurran my only concern is the name clash with the `~purpose` decorator defined in another RFC

features/0434-outofband/README.md
swcurran: Once a registry mechanism has been defined, we can move the list to that.  For now, I put a table in the message catalog with a few initial values. I don't think they are perfect, but that will get the discussion started.  I hope that discussion will be a new issue and not part of this PR :-).

swcurran:
Per @llorllale suggestion, I've renamed `purpose-cd` and `purpose` to `goal-cd` and `goal` - good catch. I also added a table of `goal-cd` values that I hope will evolve via an issue after we merge this PR.

Folks, please resolve the issues you have raised that are resolved and let's see where we are in getting this PR merged, or where we need to focus next.

Great discussion!

llorllale:
@swcurran 

> Folks, please resolve the issues you have raised that are resolved and let's see where we are in getting this PR merged, or where we need to focus next.

Looks like I don't have enough permissions to resolve my own comments.

Anyway, PR looks good to me.

features/0434-outofband/README.md
```diff
@@ -0,0 +1,463 @@
...
+
+We need the ability to reuse a connection.
+
+### Connection Establishment Versioning
+
+In the existing Connections and DID Exchange `invitation` handling, the _inviter_ dictates what connection establishment protocol all _invitee_'s will use. A more sustainable approach is for the _inviter_ to offer the _invitee_ a list of support protocols and allow the _invitee_ to use one that it supports.
+
+### Handling of all Out of Band Messages
+
+We currently have two sets of out of band messages that cannot be delivered via DIDComm because there is no channel. We'd like to align those message into a single "out of band" protocol so that their handling can be harmonized inside an agent, and a common QR code handling mechanism can be used.
```
dhh1128: I'm fine with having one RFC, but I'm really dubious about having one *protocol*. The messages have no relationship to one another other than their out-of-band-ness. They should be versioned and implemented independently; an agent should be able to support one and not the other.

features/0434-outofband/README.md
dhh1128: I don't see any example other than QR codes. Am I missing something? I searched for a fix...

features/0434-outofband/README.md
dhh1128: Never mind. Not important enough to worry about.

features/0434-outofband/README.md
```diff
@@ -0,0 +1,463 @@
...
+
+The *receiver* state machine is as follows:
+
+[![receiver state machine google doc](state-machine-receiver.png)](https://docs.google.com/spreadsheets/d/1o1szGCMx6FzrxfAKV76IEYmeKRJqvV3gd1r3Qfv1Q9o/edit#gid=10874521)
+
+Worth noting is the first event of the `done` state, where the receiver may receive the message multiple times. This represents, for example, an agent returning to the same website and being greeted with the same QR code each time.
+
+### Messages
+
+There are two messages in the out of band protocol. They are both sent by the *sender*. Only one of the message types is ever used in an instance of the protocol.
```
dhh1128: Only one of the message types is ever used in an instance of the protocol --> that's because these are two different protocols.

features/0434-outofband/README.md
swcurran: See example email message directly above [this heading](https://github.com/hyperledger/aries-rfcs/blob/95135828ce722f61cbc6dae3633f8b38f441c8aa/features/0434-outofband/README.md#url-shortening), below the QR code.

If you were looking for something else, please be a little more specific of what you want to see.

features/0434-outofband/README.md
swcurran: Please propose exactly what change you want.  I don't see the benefit, but I'd like to get to the end of the discussion.

features/0434-outofband/README.md
dhh1128: Thank you for pointing me to it. I did genuinely look before I asked, but I guess I didn't look enough. Sorry to make you point out the obvious. The example is great.

dhh1128:
Can we change "goal-cd" to "goal-code"? Abbreviating for no good reason just adds unnecessary opacity.

features/0434-outofband/README.md
dhh1128: Okay. Will do.

swcurran:
Changed goal-cd to goal-code.

features/0434-outofband/README.md
dhh1128: Addressed in https://github.com/swcurran/aries-rfcs/pull/1

features/0434-outofband/README.md
dhh1128: Done. https://github.com/swcurran/aries-rfcs/pull/1

swcurran:
Folks, @dhh1128 has proposed that we have one RFC, but two protocols.  He has submitted [this change] (https://github.com/swcurran/aries-rfcs/pull/3/files), with the core of the change being that we have two protocols "oob-invitatation" and "oob-request", so that the messages become:

- https://didcomm.org/oob-invitation/%VER/invitation
- https://didcomm.org/oob-request/%VER/request

My $0.02CDN - I'm fine with the change. I don't think it is crucial, but am more than happy to accept the PR if others would like to see it merged.  Particularly interested in knowing this would impact  implementation.

@TelegramSam @andrewwhitehead @llorllale

andrewwhitehead:
Having the separated protocols makes sense to me. I expect there will be agents that only implement one or the other. That doesn't affect things like feature discovery because the messages can't be received over DIDComm, but it makes recognizing an OOB message slightly more complicated for other tools (you can still do it in one regex though).

llorllale:
@swcurran fine by me

swcurran:
Good stuff - that's been merged.  @dhh1128 has a grammatical change to make (or I'm happy to make it) and then are we ready?

features/0434-outofband/README.md
```diff
@@ -0,0 +1,465 @@
...
+
+A common pattern we have seen in the early days of Aries agents is a user with a browser getting to a point where a connection is needed between the website's (enterprise) agent and the user's mobile agent. A QR invitation is displayed, scanned and a protocol is executed to establish a connection. Life is good!
+
+However, with the current invitation processes, when the same user returns to the same page, the same process is executed (QR code, scan, etc.) and a new connection is created between the two agents. There is no way for the user's agent to say "Hey, I've already got a connection with you. Let's use that one!"
+
+We need the ability to reuse a connection.
+
+### Connection Establishment Versioning
+
+In the existing Connections and DID Exchange `invitation` handling, the _inviter_ dictates what connection establishment protocol all _invitee_'s will use. A more sustainable approach is for the _inviter_ to offer the _invitee_ a list of supported protocols and allow the _invitee_ to use one that it supports.
```
kdenhartog: It should be noted how failure should occur if the invitee does not support any of the provided protocols from the inviter.

features/0434-outofband/README.md
kdenhartog: > there is no harm done, since any are by definition OK to use.

If a protocol is supported for legacy purposes, but really shouldn't be the default then this could present an edge case where it causes a problem.

features/0434-outofband/README.md
```diff
@@ -0,0 +1,465 @@
...
+### Standard Out of Band Message Encoding
+
+Using a standard out of band message encoding allows for easier interoperability between multiple projects and software platforms. Using a URL for that standard encoding provides a built in fallback flow for users who are unable to automatically process the message. Those new users will load the URL in a browser as a default behavior, and may be presented with instructions on how to install software capable of processing the message. Already onboarded users will be able to process the message without loading in a browser via mobile app URL capture, or via capability detection after being loaded in a browser.
+
+The standard out of band message format is a URL with a Base64URLEncoded json object as a query parameter.
+
+The URL format is as follows, with some elements described below:
+
+```text
+https://<domain>/<path>?c_i=<outofbandMessage>
```
TelegramSam: `c_i` was short for connection invitation. Given the changed nature and naming of this protocol, I think it should be updated to something more appropriate. Perhaps `oob`?

swcurran: Fixed.

features/0434-outofband/README.md
swcurran: Interesting point - thanks. I've added what I think could be possible in this [commit](https://github.com/hyperledger/aries-rfcs/pull/434/commits/d65de658932403c6558f1413ec32d502b06297db).

Folks - please take a look and see what you think.

dhh1128: I like it. I'm glad this was added. Good improvement.

dhh1128:
I've raised a PR against this PR, suggesting that we change "out of band" to "out-of-band" in most places. This is because of a picky grammar/spelling rule that says to hyphenate when a phrase as a whole is used as an adjective. See http://bit.ly/2IDXE6S. This is not a substantive change; we can take it or leave it. It is not vital to accept such a change before merging this; I'm just offering it if people want.

PR #435, opened 2020-02-28
Update Message Types, DIDComm Transports and JSON-LD Compatibility RFCs

TimoGlastra:


concepts/0020-message-types/README.md
```diff
@@ -61,7 +61,7 @@ did:sov:123456789abcdefghi1234;spec/exampleprotocol/1.0/exampletype
 
 This example DID document shows a service endpoint that includes a name property (emphasized) whose purpose is to enable creation of DID references that can deterministically select that service in order to have an algorithmic transformation into a concrete URI.
 
-```json
+```jsonc
 {
   "@context": "https://w3id.org/did/v1",
   "id": "did:example:123456789abcdefghi",
```
swcurran: Could yo change this as well? An ID should (in general) be a GUID, not a DID.

swcurran:
All the rest of the changes are good - thanks for this.  If you could add the one additional change, we can merge this.  Or we can merge it now :-).

concepts/0020-message-types/README.md
dhh1128: @swcurran , this is an `id` in a DID doc, not an `id` in a DIDComm message. Therefore, having the `id` be a DID is actually normal.

dhh1128:
I'm pretty sure that the comment about message-types/README.md should be considered resolved. If you don't agree, @swcurran , let me know, and I'll go fix it myself. In the meantime, I'm merging to keep us moving.

concepts/0020-message-types/README.md
swcurran: Doh.  Sorry.  Trying to figure out why a DIDDoc would be in that document.  It will be deleted soon :-)

PR #437, opened 2020-02-29
RFC 0056: Service decorator requirements for multi-message protocols

smithbk:


features/0056-service-decorator/README.md
```diff
@@ -19,6 +19,12 @@ Examples include the Connect Protocol and Challenge Protocols.
 
 The `~service` decorator on a message contains the service definition that you might expect to find in a DID Document. These values function the same way.
 
+The `~service` decorator may also be used for protocols which involve more than a simple request and response.  For example, using the `~service` decorator with the
+Present Proof Protocol allows a prover to present a proof to a verifier when the prover and verifier do not have a pre-existing connection or relationship.
+In this case, the `~service` decorator must be used on all messages of the Present Proof Protocol.
+
+More generally, when a `~service` decorator is used for any single message of a protocol, the `~service` decorator MUST be present on ALL messages of that protocol.
```
swcurran: I'm not sure this matters because I think the need for the ~service decorator becomes unnecessary in the request/response use case if the Out of Band PR/RFC happens.  However, I would argue that when used with the Present Proof protocol, it is a simple request/response flow. Further, I'm not sire the MUST on that last statement is true and definitely not enforceable. The agents can/will do what they want in using the ~service decorator.  We've yet to see the ~service decorator used on other than a request/response, so I'm not sure of a practical use case for this.

smithbk: Stephen, what "Out of Band PR/RFC" are you referring to?  Can you elaborate?  It would help me (and perhaps others) to see specific examples of how the service decorator is being used today.  And if it should definitely not be used in specific cases, this should also be stated in the RFC.  And if some or all of this info is stated elsewhere, please point me to it and I'll be happy to create a PR to this RFC which references the others.

When used with the Present Proof Protocol, I'm thinking of the following message flow:

1. Verifier sends presentation request to prover

2. Prover sends presentation to verifier

3. Verifier sends ACK to prover

This is more than a single request/response flow given the additional ACK.  And using the service decorator in this flow has use cases I believe.

For example, Alice approaches a Bartender who needs proof that Alice is 21.  There is no connection between Alice and Bartender, so Alice begins the verification flow with a presentation proposal message to Bartender with a service decorator.

dhh1128:
@smithbk : What would you like to do about this PR? I am trying not to let it languish without attention...

smithbk:
@dhh1128 @swcurran My overall request that we define an interoperable way of doing connection-less issuance and verifications.  On last week's Arie's call, Tomislav @tmarkovski gave a nice presentation of connection-less issuance with example messages; however, this assumed a protocol like HTTP which supports the return route decorator.  I think we should somewhere define or provide examples of how to perform connection-less issuance and verification flows which work over any protocol.  If this exists already some place, please let me know and I will be happy to cancel this PR.

m00sey:
Coming out of the most recent AIP discussions a proposal was made to deprecate 0056, we should close this PR as it is superseded by https://github.com/hyperledger/aries-rfcs/tree/master/features/0434-outofband

swcurran:
Closing as AIP 1.0 will likely not include this protocol.  Please reopen if this is an inappropriate change.

PR #440, opened 2020-03-06
Proposed RFC 0440 - Storage RFC

mikelodder7:
This RFC describes how a KMS should be architected and covers many considerations when designing them.


concepts/0440-kms-architectures/README.md
```diff
@@ -0,0 +1,422 @@
...
+- Since: 2020-03-06 (date you submit your PR)
+- Status Note: Proposed
+- Supersedes: [Wallets](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0050-wallets/README.md), uses concepts from [LOX](https://github.com/hyperledger/aries-rfcs/blob/master/features/0042-lox/README.md)
+- Start Date: 2019-12-02
+- Tags: [concept](/tags.md#concept)
+
+## Summary
+
+A Key Management Service (KMS) is designed to handle protecting sensitive agent information
+like keys, credentials, protocol state, and other data.
```
Baha-sk:
```
credentials, protocol state, and other data
```

These are not part of a KMS concept/Architecture. As the name portrays it, it manages only keys. For other personal confidential data uses, we have an Encrypted Data Vault (or EDV) to store credentials and other confidential data. 

We can probably combine the two (KMS and EDV) under one main umbrella term, like Agent Vault (AV) or something more meaningful. 

I think it is best to use clear terminology to avoid reader/user confusion. Keys and key management systems are well known terms. IMHO, it's better to not expand its scope in Aries.

What do you think?

swcurran:
@mikelodder7 - could you please change `concepts` in the "Tags" field in the header to `concept`?  Once done, we can review (rubber-stamp) this PR and merge it.

Thanks

PR #442, opened 2020-03-11
DID Exchange Update with Complete and Reuse

TelegramSam:


llorllale:
@TelegramSam why not remove `invitation` from the protocol as well?

TelegramSam:
I have split the message as discussed in 2020-03-11 Aries B Call. 

Comments Welcome.

PR #446, opened 2020-03-16
Update existing Rich Schema RFCs; Propose RFCs for Rich Schema's Mapping and CredDef

ashcherbakov:
- Update existing Rich Schema RFCs according to the Rich Schema Common RFC;
- Propose RFCs for Rich Schema's Mapping and CredDef according to the Rich Schema Common RFC


ashcherbakov:
@dhh1128 @brentzundel Can you please review?

concepts/0420-rich-schemas-common/README.md
```diff
@@ -86,24 +103,45 @@ data registries (ledgers).
...
+- Mapping 
+- Presentation Definition
+
+Context object can also be in JSON-LD format.
+
+
+If a Rich Schema object is a JSON-LD object, the `content`'s `@id` field must be equal to the `id`.
+
+The only thing that we currently expect from json-ld processing is substitution of attributes by a fully-qualified ones.
+We may assume that contexts belonging to the current ledger only
```
kdenhartog: I'm curious to understand this a bit better. Does this mean that all contexts necessary to build and map a rich schema must be on the same ledger? Would this hold true for the use of schemas that are anchored in IPFS that sit alongside an IPFS implementation. If this is true, then does that mean that different implementations of sidetree could reuse contexts from "different registries"?

While I understand this question is a bit futuristic sounding, the main reason I ask is because it makes sense to design in a way that aligns with securekey's architecture which using sidetree with Fabric. For rich schemas, they wouldn't necessarily have to anchor them using the batch files in sidetree, and instead could use IPFS for this functionality.

features/0249-rich-schema-contexts/README.md
```diff
@@ -10,10 +10,14 @@
 ## Summary
 [summary]: #summary
 
-Every rich schema object has an associated `@context`. Contexts are JSON-LD
+Every rich schema object may have an associated `@context`. Contexts are JSON or JSON-LD
```
kdenhartog: How will it be known if it's a JSON vs JSON-LD object? For example, if the context is anchored in JSON-LD form, then it's acceptable that the context is IRIs that are expanded by the document loader. In the case of a JSON consumer though, I'm not certain how they would receive the necessary detail to handle processing of an IRI. Could we provide additional details on this?

concepts/0420-rich-schemas-common/README.md
ashcherbakov: We have this assumption in Indy, where we assume all Schemas and corresponding Contexts should be on the same Ledger. BTW this assumption in Indy may be for the first MVP implementation of Rich Schema feature only.

As for the Aries side, I think I will remove this assumption.

features/0446-rich-schema-cred-def/README.md
```diff
@@ -0,0 +1,162 @@
...
+This increases the complexity of issuing verifiable credentials and
+verifiying the accompanying verifiable presentations. 
+
+## Rationale and alternatives
+
+- A `schema` object is supported by the W3C Verifiable Credentials Data
+Model specification.
+- It supports rich schema capabilities for credentials.
+
+A possible alternative to this work has been presented by Workday.
```
brentzundel:
```suggestion
```

features/0446-rich-schema-cred-def/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+This increases the complexity of issuing verifiable credentials and
+verifiying the accompanying verifiable presentations. 
+
+## Rationale and alternatives
+
+- A `schema` object is supported by the W3C Verifiable Credentials Data
```
brentzundel:
```suggestion
- A `credentialSchema` object is supported by the W3C Verifiable Credentials Data
```

features/0249-rich-schema-contexts/README.md
ashcherbakov: Any Rich Schema (which is a JSON-LD object) may reference Contexts by their `id`. The Context's content can be used then for substituting the attributes with IRIs within the Rich Schema object.

The question is whether the Context's content itself must be a JSON-LD as well. 
As I can see from https://json-ld.org/spec/latest/json-ld/#the-context, it's not.
And as I can see from the example of Contexts that we have (@brentzundel can correct me if I'm wrong),  the only requirement for the Context's format is to have a `@context` property containing a map of terms to IRIs. A Context may not have `@id` and `@type` properties expecting from any JSON_LD object.

brentzundel: My understanding of the `@context` objects is that simple ones are JSON objects, but there may be more advanced options that require JSON-LD. 
It may be a moot question though, because the object that **has** a `@context` is a JSON-LD object and and processing that object will include dealing with the `@context`. I think its safe to assume that a JSON-LD processor will know how to deal with `@context` objects.

kdenhartog: Ahh good points. Thanks for the clarifications

dhh1128:
I'm cheerful about merging this, if others are.

kdenhartog:
> I'm cheerful about merging this, if others are.

Yup looks fine from my perspective.

PR #448, opened 2020-03-18
RFC 0017: Add JSON web signature to attachment decorator

andrewwhitehead:
Based on @dhh1128 's closed PR #267, this is meant to support signing of DIDDocs in the DID-Exchange protocol and supplant the current ~sig decorator in general.

The main differences from the previous revision are the use of detached content for the signature (allowing signatures over external linked data) and recommendation of did:key over JWK for the inlined public keys (for terseness and some consistency between RFCs).


concepts/0017-attachments/README.md
```diff
@@ -338,6 +336,72 @@ whichever mechanism(s) are best suited to its individual needs and capabilities.
...
+{
+  "@type": "https://didcomm.org/xhomebuy/1.0/home_insp",
+  "inspection_date": "2020-03-25",
+  "inspection_address": "123 Villa de Las Fuentes, Toledo, Spain",
+  "comment": "Here's that report you asked for.",
+  "report~attach": {
+    "mime-type": "application/pdf",
+    "filename": "Garcia-inspection-March-25.pdf",
+    "data": {
+      "base64": "eyJ0eXAiOiJKV1QiLA0KICJhbGciOiJIUzI1NiJ... (bytes omitted to shorten)",
```
dhh1128: I think the JWS should be inside "data", not a peer of it. That's because "data" already supports "hash"; "jws" is a close analog to "hash".

andrewwhitehead: I wondered about that, and had it in `data` originally but it was suggested on the Aries call to move it up a level. I agree it makes sense to have them at the same level, whether inside or outside the data block.

dhh1128: I don't think this is a signature over the mime type, the filename, etc -- only over the raw bytes of the data. So I think it belongs next to the raw bytes of the data, not at the outer level.

TelegramSam: I prefer the attachment level, but not strongly. I think of it as a 'signed attachment' not an 'attachment with signed data'.

swcurran: My $0.02CDN is that is should be within the `data` element as @dhh1128 suggests.

PR #452, opened 2020-03-23
tweak: RFC 0037 removes redundant state

soluchok:
1. Verifier cannot receive a `request`. According to that `request-received` state is not possible for the Verifier. 

2. Instead of `reject-sent` I suggest rename it into some more generic name e.g `abandoning`. The `abandoning` state is responsible for sending `report-problem` message. So, when an internal error occurs or some message of the protocol were rejected (e.g by another agent) protocol goes to the `abandoning` state and sends a report problem message with a specific code e.g (errorCodeInternal, errorCodeRejected etc.).

Signed-off-by: Andrii Soluk <isoluchok@gmail.com>

### DIFF
--- a/features/0037-present-proof/README.md
+++ b/features/0037-present-proof/README.md
@@ -41,8 +41,8 @@ Diagrams in this protocol were made in draw.io. To make changes:
 
 * request-sent
 * proposal-received
-* request-received
 * presentation-received
+* abandoning
 * done
 
 #### states for Prover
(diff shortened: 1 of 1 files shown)


dhh1128:
@soluchok : This PR has languished a bit. Sorry; it happened to arrive just after Hyperledger Global Forum as COVID-19 craziness was distracting us. If you are still interested in getting it merged, would you like to ask for some people to review it on the next community call?

PR #453, opened 2020-03-23
WIP New RFC: Issue Credential v2

swcurran:


swcurran:
First try at V2 of Issue Credential.  @troyronda @dhh1128 @TelegramSam @andrewwhitehead @KitHat 

Things I tried to address: 

- New major version, new RFC
- Credential Proposal uses attachment
- Discuss attachment handling
  - Option of using json vs. base64
  - State explicitly that the attachment array is not intended for supporting bulk (more than 1) issuance.
- Per implementation section with details about the attachments
   - Indy section added per message
   - Specific values supported for the `@id` field of the attachment
   - Would like to get input from @troyronda and team on having the credential model(s) supported by af-go added
- Remove some things about previous implementations
- Added some more `Questions`

swcurran:
Added feedback from a recent [Aries WG Call - March 25](https://wiki.hyperledger.org/pages/viewpage.action?pageId=31196283), notably:

- added the `formats` array to link an attachment to its format/version (per @dhh1128 suggestion)
- clarified the registry to make it clear that it is there per message/attachment type.
- clarified that entries in the registry may be format (e.g. Hyperledger Indy) or ledger specific (e.g. Sovrin) or anything else (per @nage suggestion).

Would be interested in getting more specific info:

- From @kdenhartog (and others) -- how would/should bulk issuance be handled?
  - Use case 1: Same credential type, but multiple instances (different claims)
  - Use case 2: Multiple credential types issued in one swell phoop
- From @llorllale and/or @troyronda 
  - Could you provide examples of the "per credential format" sections that AF-Go supports?  You can do pushes to my repo or just send me the text and I'll put it in.

And of course, feedback from everyone is encouraged.

swcurran:
This merge is ready.  @TelegramSam @llorllale @dhh1128 @kdenhartog - care to do the honours?

PR #455, opened 2020-03-25
DID Exchange Reuse-accepted message

TelegramSam:


TelegramSam:
On the Aries WG B Call 2020-03-25, we discussed replacing the `reuse-accept` message (and the `complete` message) with an adopted `ack` message. Please discuss here.

swcurran:
For this use I would like to see us go with `reuse-accept` and `complete`.

I'll summarize my view, expressed on the call and strengthened after reviewing [RFC 0015 Acks](https://github.com/hyperledger/aries-rfcs/tree/master/features/0015-acks). This is not a hill I will die on, but I just think adopted acks don't really make sense.

- This RFC (and others) must explicitly call out when required acknowledgements/confirmations are necessary. Little to no savings by adopting ack.
- Each protocol implementation must explicitly receive and update the state machine implementation (aka the protocol state object) for each required acknowledgement. Little or no savings by adopting ack.
- By using specific names for required acks (such as "reuse-accept"), it's clear what each message is for and it's expected impact, which is easier for the developer.  This would be particularly true in protocols like this one where there are multiple explicit acks.

Additional thoughts:

- The ack message provides so little information (a single field with a three value enum) that I expect (with no evidence :-) ) that the trade-off of a "standard" ack vs. case-specific types is irrelevant. With case-specific message types we can include extra items or have no items at all.
- I would still like to keep "ack", but for use primarily with the "~please_ack" decorator.
- I still would like "problem-report" to be adopted as originally suggested in this and other protocols. It is much richer/more complicated, and thus benefits more from standardization.

PR #456, opened 2020-03-25
DID Exchange - Application of signed attachments

TelegramSam:


features/0023-did-exchange/README.md
```diff
@@ -258,13 +258,17 @@ The _invitee_ will provision a new DID according to the DID method spec. For a P
...
-  "connection": {
-    "did": "B.did@B:A",
-    "did_doc": {
-        "@context": "https://w3id.org/did/v1"
-      	// DID Doc contents here.
-    }
-  }
+  "did": "B.did@B:A",
+  "did_doc~attach": {
+     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
```
llorllale: is the intention to force this form of attachments (base64)?

TelegramSam: It must be if it is to be signed. Well, any canonical form can be signed, but base64 is the only currently proposed one.

kdenhartog: can we add a definition of the format of base64 in here? E.g. it's Base64url with no padding (to stay consistent with jose)

features/0023-did-exchange/README.md
```diff
@@ -298,13 +301,17 @@ When a `request` responds to an implicit invitation, its `~thread.pthid` MUST co
...
-        "@context": "https://w3id.org/did/v1"
-        // DID Doc contents here.
-    }
-  }
+  "did": "B.did@B:A",
+  "did_doc~attach": {
+     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
+     "jws": {
+        "header": {
+           "kid": "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
```
kdenhartog, line 309: should this also include the key fragment of the signing key?

features/0023-did-exchange/README.md
```diff
@@ -298,13 +301,17 @@ When a `request` responds to an implicit invitation, its `~thread.pthid` MUST co
...
+        },
+        "protected": "eyJhbGciOiJFZERTQSIsImlhdCI6MTU4Mzg4... (bytes omitted)",
```
kdenhartog, line 311: what's the difference between header and protected here?

features/0023-did-exchange/README.md
TelegramSam: in did key, there is only one key. Should a fragment be needed there?

features/0023-did-exchange/README.md
TelegramSam: Any of those improvements should be made to the attachments signing RFC, then reflected here.

features/0023-did-exchange/README.md
TelegramSam: good question for the attachments signing RFC.

features/0023-did-exchange/README.md
kdenhartog: did:key with ed25519 has 2 keys. One signing key (ed25519) and one key agreement key (x25519) that's derived from the signing key. 

Notice the fragments are different in this example: https://w3c-ccg.github.io/did-method-key/#example-2-a-did-document-derived-from-a-did-key

it's because it's the fingerprint of the publickey. 

Here's an example of the implementation where it does the deriving: https://github.com/digitalbazaar/did-method-key-js/blob/master/lib/driver.js#L104

kdenhartog:
Would it be too troublesome to hold off on merging until we resolve questions in attachment singing RFC or would you rather merge this and then I update both later?

swcurran:
I think that this should go ahead and comments should focus on the attachments RFC. Anything in these changes are just examples of the attachments RFC and that is the definitive source for implementations. If we miss something in an example, we can correct it without changing the version of this protocol.

kdenhartog:
I'm good with that approach.

PR #474, opened 2020-05-06
Protocol Role Request initial

TelegramSam:

### DIFF
--- a/features/0474-protocol-role-request/index.md
+++ b/features/0474-protocol-role-request/index.md
@@ -1,4 +1,4 @@
-# Aries RFC 0470: Role Request 0.1
+# Aries RFC 0474: Role Request 0.1
 - Authors: [Sam Curren](telegramsam@gmail.com)
 - Since: 2020-05-06
 - Status Note: initial proposal
--- a/features/0470-protocol-role-request/index.md
+++ b/features/0470-protocol-role-request/index.md
@@ -1,0 +1,87 @@
+# Aries RFC 0470: Role Request 0.1
+- Authors: [Sam Curren](telegramsam@gmail.com)
+- Since: 2020-05-06
+- Status Note: initial proposal
+- Start Date: 2018-05-05 
+- Tags: feature, protocol
+
+## Summary
+
+This protocol requests the recipient to play a specified role in another protocol.
+
+## Motivation
+
+It is sometimes desirable to ask the other party to play a role within another protocol. Although some protocols already have a way to do this as an existing feature of the protocol, it is desirable have a general purpose way to communicate that intent.
+
+## Tutorial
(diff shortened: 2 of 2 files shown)


features/0474-protocol-role-request/index.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+
+
+**protocol**: The Protocol URI
+
+**role**: The role name specified in the specified protocol.
+
+Note that no other information is present in this message. If information is required to start the protocol in the role, it likely warrants a message within the protocol itself.
```
llorllale, line 59: This is exactly the realization I had with the proposed change to the Out-of-Band's `invitation` message - simply informing the other party what role the sender wishes them to perform is not enough. That role needs to be parametrized somehow. The sender including the request message is the most straightforward way to do this.

Think back to @dhh1128 's example from the last call: how do I ask you to connect and then bid on my Rembrandt? Asking you to play the role of *bidder* is not enough - I need to also instruct you to bid on the *Rembrandt*.

Functionally speaking, the out-of-band `request` message is enough for this purpose if we agree that parties may form  a permanent connection prior to the receiver sending the message.

swcurran: Two thoughts out of this:

- a `label` on the request could be sufficient.  Or we could reuse the `goal` and `goal-code` pair from OOB for that purpose.
- Once again we're back to OOB `request` "being sufficient" for everything, if we added the fields from the 'invitiation' message into it.  Or vice versa. We're doing development on OOB and it would be really nice to nail down the remaining issues with it.  My $0.02CDN is that we just have one OOB message and put everything and the kitchen sink in it :-). We keep moving in that direction.

llorllale: I agree on collapsing the two messages.

I don't think goal-code is sufficient to parametrize the role for the receiver. A parameter can be an identifier or a reference to a resource on the sender's side.

dhh1128: @llorllale: I am conflicted. I agree with you that goal may not be enough. However, including the message has some conceptual problems. Explaining them will take a page or two of text. Let me do a writeup and then I'll come back here and link to what I create. Then we can discuss further.

dhh1128: @llorllale (and @swcurran and @TelegramSam ): here is my answer about a more elegant way to do this: https://github.com/hyperledger/aries-rfcs/pull/482. Read RFC 0478 first, and 0482 second. This is only a draft, but I think it will spark an interesting conversation.

TelegramSam:
Given the already accepted replacement for this concept, I've decided to retire this without merging.

PR #475, opened 2020-05-12
Update OOB to be a single protocol with a single message

swcurran:
Given the evolution of the OOB invitation message, this proposal eliminates the two messages in favour of one that handles both purposes, plus a third where both a connection is established and a request is made for use after the connection is established.


features/0434-outofband/README.md
```diff
@@ -82,25 +80,34 @@ The *receiver* state machine is as follows:
...
       "https://didcomm.org/didexchange/1.0",
       "https://didcomm.org/connections/1.0"
-      ]
+      ],
+  "request~attach": [
+    {
+        "@id": "request-0",
+        "mime-type": "application/json",
+        "data": {
+            "base64": "<bytes for base64>"
```
TelegramSam, line 107: if we use `json` instead of `base64` data encoding, we avoid the 33% bloat incurred with re-base64 encoding something.

swcurran: I'll update.  We had the same issue in the "issue credential" v2.0 exchange.  Thanks.

llorllale: I didn't take that example as normative. Should I? How the sender encodes the attachment seems out of scope of this RFC.

llorllale:
LGTM

swcurran:
Applied feedback from the Aries WG --notably the change from "setup_protocols" to "handshake_protocols". Added some more on using both lists at once in a message -- notably that reuse of the connection should be on the request~attach.

Suggest we merge this and then get more in a separate PR on the URL shortening.

swcurran:
@llorllale -- please add a review vs. just a comment.

PR #481, opened 2020-05-21
0003 Protocols RFC - Adjust URI matching to not ignore case and punctuation; simplify; all content in README.md

swcurran:
A significant change to the "protocols" RFC, summarized here:

- The only substantive change in the protocol is the removal of the line that URIs matching should "ignore case and punctuation". Per the Aries WG call last week, the community does not want that flexibility and no one has (AFAWK) has implemented it.
- Inlined the external documents into the README.  In doing that, I have tried to simplify without changing the content.
   - Since the changes are significant, I have left the documents in place for now, and have updated them to match the content in the README so it is easier to compare old and new.
   - Before merging this, the external documents (e.g. uris.md, etc.) will be removed, and references to those documents in other RFCs will be updated.
- Simplifications (in my opinion) to other parts of the document. 
- Removed an "easy" folder within the tictactoe folder that I assume was user specific.

Note that no images are altered in the preparation of this PR.

@dhh1128 - you should definitely review the changes to see if you are OK with them.


concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
-# Roles, Participants, Parties, and Controllers
+### Roles, Participants, Parties, and Controllers
 
-## Roles
+#### Roles
 
-The __roles__ in a protocol are the perspectives (responsibilities, privileges) that parties take on an
-interaction. 
+The __roles__ in a protocol are the perspectives (responsibilities, privileges) that parties take i an
```
dhh1128, line 5: typo "take i an" --> "take in an"

concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
...
 This perspective is manifested in three general ways:
 
  * by the expectations that a party takes on in a protocol (ex. a role may be expected to do something to start a protocol).
  * by the messages that a party can and does use in the course of the protocol (some messages may be reserved for a single role, while other may used by some if not all roles).
  * by the state and the transition rules
- 
-Like parties, roles are normally known at the start of the protocol but this is not a requirement.
 
-In an auction protocol, there are only two roles--*auctioneer*
-and *bidder*--even though there may be many parties involved.
```
dhh1128, line 17: I don't like losing this example.

concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
...
+Like parties, roles are normally known at the start of the protocol but this is not a requirement.
 
-## Participants
+#### Participants
 
-The __participants__ in a protocol are the agents that consume and/or emit
```
dhh1128, line 21: I don't like losing this example. I have had several conversations with people who don't get the subtlety that's being described here.

concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
...
+The __participants__ in a protocol are the agents that send and/or receive
+[plaintext application-level messages](../../concepts/../features/0044-didcomm-file-and-mime-types/README.md)
+that embody the protocol's interaction.
 
-## Parties
+#### Parties
 
 The __parties__ to a protocol are the entities directly responsible for achieving the protocol's goals.
 When a protocol is high-level, parties are typically people or organizations; as protocols become lower-level,
-parties may be specific agents tasked with detail work through delegation.
```
dhh1128, line 39: I don't like losing this example. This is a concept RFC, not a impl spec; we need to explain the concepts well. The definition that remains is accurate, but too terse.

concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
...
-to play in the ocean on day 2. The concierge engages in a sub-sub-protocol to reserve the waverunner. The
-goal of this sub-sub-protocol is to reserve the equipment, not to book a hotel or arrange a vacation. The parties to this
-sub-sub-protocol are the concierge and the person or automated system that manages waverunners.
+parties may be specific agents delegated with the detail work.
 
 Often, parties are known at the start of a protocol; however, that is not a requirement. Some protocols might commence
 with some parties not yet known or assigned.
 
 For many protocols, there are only two parties, and they are in a pairwise relationship. Other protocols
-are more complex. Introductions involves three; an auction may involve many.
```
dhh1128, line 65: Again, losing important detail.

concepts/0003-protocols/roles-participants-etc.md
```diff
@@ -1,83 +1,41 @@
...
 
-The __controllers__ in a protocol are entities that make decisions. They may or may not be direct parties.
+#### Controllers
 
-Imagine a remote chess game between Bob and Carol, conducted with software agents. The chess protocol isn't
-technically about _how to select a wise chess move_; it's about _communicating the moves_ so parties achieve
-the shared goal of running a game to completion. Yet choices about moves are clearly made as the protocol
-unfolds. These choices are made by controllers--Bob and Carol--while the agents responsible for the work of
-moving the game forward wait with the protocol suspended.
+The __controllers__ in a protocol are entities that make decisions. They may or may not be direct parties. Controller supplies the business logic or rules that drive each specific instance of a protocol. They provide the personality of the agent suitable for the use case it implements.
```
dhh1128, line 36: Again, loses too much info.

dhh1128: Typo: "Controller supplies" doesn't make sense.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
-constitutes a [public API in the semver sense](https://semver.org/#spec-item-1).
-An individual message type can add new optional fields, or deprecate
-existing fields, [with only a change to its protocol's minor
-version](https://semver.org/#spec-item-7).
-Similarly, a protocol can add new message types (or [adopted
-ones](../../0000-template-protocol.md#adopted-messages)) with only a change
-to the minor version. It can announce deprecated fields. It can add additional
-semantics around optional decorators. These are all backwards-compatible
-changes, also requiring only a minor version update.
+The major and minor versions of protocols match semvar semantics:
```
dhh1128, line 9: "semvar" --> "semver"

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
+- An individual message type can add new optional fields, or deprecate
+existing fields, [with only a change to its protocol's minor
+version](https://semver.org/#spec-item-7). Similarly, a protocol can
+add new or deprecate message types with only a change to the minor version.
+- Changes that remove fields or message types, that make formerly optional
 things required, or that alter the state machine in incompatible
 ways, must result in an [increase of the major version of the protocol/primary
 message family](https://semver.org/#spec-item-8).
 
-Because protocol handling choices depend mainly on major and minor version
```
dhh1128, line 24: Although I agree that core protocols use only major.minor, I am not sure we should delete this paragraph. Are we saying that it's impossible to use patch with protocols -- or only that we don't use it in our our work? Do we want agents to support patch properly, or not? If not, we should explicitly say so. The previous statement about it not being used in core protocols is not sufficient to make this a global characteristic.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
-know about. Generally, all messages after the first use only major.minor
+- message types received with a minor versions below the minimum may be answered with a `report-problem` with code `version-not-supported`
+- message types received with a minor version at or higher than the minimum supported and less than the current minor version are processed, ideally with a response using the same minor version of the received message
+  - The recipient may want to send a warning `report-problem` with code `version-with-degraded-features`
+- message types received with a minor version higher than the current minor version are processed with any unrecognized fields ignored
+  - The recipient may want to send a warning `report-problem` with code `fields-ignored-due-to-version-mismatch` 
 
-[![version negotiation matrix](version-negotiation-matrix.png)](
-https://docs.google.com/spreadsheets/d/1W5KYOqCCqmTeU4Z7XZQH9_6_0TeP5Vf5TtsOZmioyB0/edit#gid=0)
+As documented in the semvar documentation, these requirements may not apply when
```
dhh1128, line 35: "semvar" --> "semver"

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
-
-Unless Alice's agent (the initiator of a protocol) knows from prior history
-that it should do something different, it should begin a protocol using the
-highest version number that it supports. For example, if A.1
-supports versions 2.0 through 2.2 of protocol X, it should use 2.2 as the
-version in the message type of its first message.
-
-#### Recipient Rules
-
-Agents for Bob (the recipient) should reject messages from protocols with major
```
dhh1128, line 45: This section is important. It has come up on community calls and is not well understood. It should not be deleted.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
-correctly by agents.
-
-### Version Negotiation
-
-The semver portion of a [message type or protocol identifier URI](uris.md) is
-used to establish the version of a protocol that parties use during an
-interaction.
-
-#### Initiator
-
```
dhh1128, line 36: This section should not be deleted.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
-feature set of the earlier version*. That is, a 1.0-capable agent can support
-1.0 features using a 1.2 message, though of course it will lose any features
-that 1.2 added. Thus, accepting such a message could have two possible outcomes:
-
-1. The message at version 1.2 might look and behave exactly like it did at version
-1.0, in which case the message will process without any trouble.
+Within a given major version of a protocol, an agent should:
 
-2. The message might contain some fields that are unrecognized and need to be ignored.
+- specify a minimum minor version supported, defaulting to "0"
```
dhh1128, line 24: How does it specify this? This simplification loses a lot of important detail. It is not equivalent; it's rather a substantive change to agreed upon rules.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
+- specify a current minor version that is fully supported and used by default
 
-In case 2, it is best practice for the recipient to send a `problem-report` that
-is a *warning*, not an *error*, announcing that some fields could not be processed
-(code = `fields-ignored-due-to-version-mismatch`). Such a message is *in addition
-to* any response that the protocol demands of the recipient.
+This leads to the following received message handling rules:
 
-If the recipient of a protocol's initial message generates a response, the response
-should use the latest major.minor protocol version that both parties support and
```
dhh1128, line 85: This statement should not be deleted.

concepts/0003-protocols/semver.md
dhh1128, line 35: "semvar" --> "semver"

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
+- message types received with a minor versions below the minimum may be answered with a `report-problem` with code `version-not-supported`
+- message types received with a minor version at or higher than the minimum supported and less than the current minor version are processed, ideally with a response using the same minor version of the received message
+  - The recipient may want to send a warning `report-problem` with code `version-with-degraded-features`
+- message types received with a minor version higher than the current minor version are processed with any unrecognized fields ignored
+  - The recipient may want to send a warning `report-problem` with code `fields-ignored-due-to-version-mismatch` 
 
-[![version negotiation matrix](version-negotiation-matrix.png)](
-https://docs.google.com/spreadsheets/d/1W5KYOqCCqmTeU4Z7XZQH9_6_0TeP5Vf5TtsOZmioyB0/edit#gid=0)
+As documented in the semvar documentation, these requirements may not apply when
+major version 0 is used. Specific rules associated each v0 protocol are left to
```
dhh1128, line 36: That's not what semver says. Semver says that prior to v. 1.0, every minor version is a major change. I don't agree with the idea of letting each protocol handle this differently.

concepts/0003-protocols/semver.md
```diff
@@ -1,93 +1,51 @@
...
+use when initiating an instance of the protocol.
 
+The [discover features](../../features/0031-discover-features/README.md) protocol
+can be used by one agent to query the versions of protocols supported by another agent,
+informing the selection of which version of a protocol to use in initiating an
+instance of a protocol. Agent receiving a discover features request may respond with a list of
+preferred protocols they support, as described in that protocol.
 
- 
+An agent should reject messages from protocols or unsupported protocol major versions with
```
dhh1128, line 48: These 3 lines feel pretty redundant with the lines just before them.

concepts/0003-protocols/state-details.md
```diff
@@ -1,80 +1,39 @@
...
+#### State machines
 
-By convention, our community describes state and sequence rules using the
+By convention, protocol state and sequence rules are described using the
 concept of state machines, and we encourage developers who implement
 protocols to build them that way.
 
 Among other benefits, this helps with error handling: when one agent
-sends a [`problem-report`](../../features/0035-report-problem/README.md)
+sends a [`report-problem`](../../features/0035-report-problem/README.md)
```
dhh1128, line 14: The protocol is named "report problem"; the message type is "problem-report".

concepts/0003-protocols/state-details.md
```diff
@@ -1,80 +1,39 @@
...
-state_machine.handle(message) # hooks do_something()
-...
-state_machine.handle(message) # hooks do_something_else()
-...
-state_machine.handle(message) # hooks do_yet_another_thing()
-```
-
-Under maintenance, the robustness of the state machine approach is compelling.
-Compare the approaches for how many places you might have to change if you had
-a bug in how state Y is handled... The state machine approach also provides a
```
dhh1128, line 53: Although the benefits of state machines as we use them are now tribal knowledge among some Aries implementers, I don't think it's a good idea to remove all the rationale here. It feels like  you're trying to make the RFC read faster. That's a noble goal. But for a concept RFC, I don't think that's the goal. Nor is simple conformance. Rather, the goal of a concept RFC is to guide a mental model. We don't accomplish that by suppressing all the discussion about why we arrived at the model we chose.

concepts/0003-protocols/uris.md
```diff
@@ -54,39 +53,35 @@ to look up the named and versioned protocol. Optionally and preferably, the
...
-   https://github.com/hyperledger/aries-rfcs/tree/18c4f82
-   )) and look for documentation on the `trust_ping/1.0` protocol.
-* `did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/trust_ping/1.0/ping` (MTURI) This
-   uses a DID reference to look up an endpoint named `spec` that serves
-   information about protocols. (The exact syntax of DID references--URIs
-   where the DID functions like a domain name, and additional info is
-   fetched from a DID Doc in much the same way IP address and hostname
-   definitions are fetched from a DNS record--is still being finalized.
-   See the latest [DID Spec](https://w3c-ccg.github.io/did-spec/) for details.)
+```
```
dhh1128, line 73: I'd prefer all of the old examples and none of the new ones. The new ones assume Aries conventions. This isn't a doc about how core protocols are written in Aries; it's a doc about how any protocols can be written, with Aries just being the place for a core subset.

concepts/0003-protocols/README.md
```diff
@@ -185,29 +217,263 @@ A protocol has the following ingredients:
...
+conventions.
+
+#### MTURI
+
+A __message type URI__ (MTURI) identifies message types unambiguously.
+Standardizing its format is important because it is parsed by agents that
+will map messages to handlers--basically, code will look at this string and
+say, "Do I have something that can handle this message type inside protocol
+*X* version *Y*?" When that analysis happens, it must do more than compare
+the string for exact equality; it may need to check for semver compatibility.
```
dhh1128, line 251: I think this phrase "more than compare the string for exact equality" accurately reflects the intent of the community after our last call, but we need to be more explicit about the comparison algorithm. We also have to be careful because semver compatibility and exact equality are contradictory concepts; semver is about predictable fuzziness.

URIs are case-sensitive after a domain name, but case-insensitive before. Per internet specs, "https://foo.com/myprotocol/1.0" and "https://FOO.COM/myprotocol/1.0" are identical URIs. "Exact equality" means something different in unicode than it does in normal human intuition (there are several byte sequences in UTF8 that map to lower-case-latin-e-with-acute-accent; all will render as visually identical, and all are considered normalized).

I suggest something like this:

Keep "A __message type URI___ ...*X* version *Y*?" as written. Then add a paragraph break, followed by:

"When this analysis happens, strings should be compared for byte-wise equality in all segments except version. This means that case, unicode normalization, and punctuation differences all matter. It is thus best practice to avoid protocol and message names that differ only in subtle, easy-to-mistake ways."

"Comparison of the version segment of an MTURI or PIURI should follow semver rules and is discussed elsewhere."

concepts/0003-protocols/roles-participants-etc.md
swcurran: fixed

concepts/0003-protocols/roles-participants-etc.md
swcurran: I don't think this example contributes to the understanding, because such a protocol introduces a host of other issues not accounted for in these protocols. A casual reader would either think this is inherently supported, or go into a rabbit hole of trying to figure out how to make an "auction" protocol work. IMHO, neither result is helpful in getting the basics of protocols.

concepts/0003-protocols/roles-participants-etc.md
swcurran: I think this example creates far more confusion about protocols, agents and DIDComm than it clarifies.  How are all those devices connected?  Where do the keys reside?  How do messages get to the different devices?  What about when the phones aren't close enough for Bluetooth?   IMHO, this is far too complex an example to be included in this intro.

What specific point is the example attempting to convey?

concepts/0003-protocols/roles-participants-etc.md
swcurran: I find this kind of example very confusing in trying to understand the core concepts and so detract from the overall understanding.

concepts/0003-protocols/semver.md
swcurran: Fixed

concepts/0003-protocols/semver.md
swcurran: Good catch on the distinction. My preference would be to eliminate them altogether vs. just for the core protocols,

In thinking about it, we really don't care either way.  If an agent is going to support a non-core protocol, it is going to get the protocol requirements for from some other place, and those requirements will indicate if support for handling a patch in the type is needed or not.  Adding that support might require changing common code for message type processing, but so be it.

I'm open to putting whatever is appropriate there -- e.g. "Core protocols must not use the patch component of a semver designation, but other protocols may include patch and handling should be included".

concepts/0003-protocols/semver.md
swcurran: I think it is covered above.

concepts/0003-protocols/semver.md
swcurran: I found this section to be extremely confusing.  I think the rules list above cover it sufficiently.

swcurran: The rules I mentioned might be below :-)

concepts/0003-protocols/semver.md
swcurran: Fixed

swcurran: Fixed

concepts/0003-protocols/semver.md
swcurran: Corrected.

concepts/0003-protocols/state-details.md
swcurran: Added "message" to the references.

concepts/0003-protocols/state-details.md
swcurran: My rationale in looking at this was that those interested in protocols would understand state-machines, but don't need to be sold on them. Those that want to see details/code can find that in the tictactoe example, which (I think) is more effective than the code snippets that were here.

concepts/0003-protocols/uris.md
swcurran: I would like at least some examples of how the URIs are actually used, hence the inclusion of the core ones here and the references.

I didn't understand why the examples removed were there.  We don't want non-core implementers to use those formats.  Where do the query parameters come from?  Are they supported?  By what?  What is the commit reference in the `trust-ping` example?  What is the point of having a URL that returns a 404?

I could not come up with a description of why they were there.  Open to adding but saying the are "valid" seems a stretch.

concepts/0003-protocols/README.md
```diff
@@ -26,9 +65,9 @@ We also need to show how a protocol is defined, so the analog to defining a Swag
 
 ## Tutorial
 
-#### What is a protocol?
+### What is an Aries protocol?
```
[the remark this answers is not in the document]
swcurran: I struggled with this as well.  I think a name is needed for this category of protocols, but the community has not come up with one.  I personally think Aries protocols is fine, but am definitely open to something else.  We can just leave "protocols" everywhere, but that too will be confusing to users.

concepts/0003-protocols/README.md
```diff
@@ -26,9 +65,9 @@ We also need to show how a protocol is defined, so the analog to defining a Swag
...
 
-A __protocol__ is a recipe for a stateful interaction. Protocols are all
+An Aries __protocol__ is a recipe for a stateful interaction. Protocols are all
```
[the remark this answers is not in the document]
swcurran: Removed

concepts/0003-protocols/README.md
```diff
@@ -69,48 +108,47 @@ The familiar world of web APIs is a world of protocols, but it comes with constr
...
-The most common protocol style in DID Communication is __request-response__.
-This style involve two *parties*, with the `requester` making the first move,
-and the `responder` completing the interaction. The [Discover Features Protocol](
-../../features/0031-discover-features/README.md) uses this style.
-
-![request-response](request-response.png)
+### Types of Protocols
 
-A second common pattern that's also important is __notification__. This style also
+The simplest protocol style in Aries is __notification__. This style
```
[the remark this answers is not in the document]
swcurran: Removed.  However, I think this is the section about "the protocols that shall not be named" protocols and I don't think we shouldn't flip between those protocols and any protocol.

concepts/0003-protocols/README.md
```diff
@@ -69,48 +108,47 @@ The familiar world of web APIs is a world of protocols, but it comes with constr
...
-Protocol](../../features/0035-report-problem/README.md) use this style.
+and the protocol ends when the `notified` receives it. The [basic message protocol](../../features/0095-basic-message/README.md) uses this style.
 
 ![notification](notification.png)
 
+Slightly more complex is the __request-response__ protocol style.
+This style involve two parties, with the `requester` making the first move,
+and the `responder` completing the interaction. The [Discover Features Protocol](
+../../features/0031-discover-features/README.md) uses this style. Note that
+with Aries protocols (and unlike an HTTP request), the request-response messages are
```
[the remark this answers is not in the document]
swcurran: changed

concepts/0003-protocols/README.md
```diff
@@ -120,17 +158,16 @@ When agents receive messages, they map the messages to a __protocol handler__
...
+### Composable
 
 Protocols are *composable*--meaning that you can build complex ones from simple ones.
 The protocol for asking someone to repeat their last sentence can be part of
-the protocol for ordering food at a restaurant. The protocols for
-reporting an error or arranging payment can be part of a protocol for
-issuing credentials.
+the protocol for ordering food at a restaurant. The protocol for
+proving the claims from some credentials (using the [present proof](../../features/0037-present-proof/README.md) protocol)
+can be part of an instance of the [issue credential](../../features/0036-issue-credential/README.md) protocol.
```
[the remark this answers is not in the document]
swcurran: Nice - that's way better.

concepts/0003-protocols/README.md
swcurran: Nice.  Changed.

concepts/0003-protocols/roles-participants-etc.md
dhh1128: Right. The intent of the comment is to suggest the host of other issues you're talking about. That's part of the scope of protocols. That's what we're trying to teach here -- that simplistic models don't account for all the complexity we're trying to account for.

concepts/0003-protocols/roles-participants-etc.md
dhh1128: The point is that agents who don't receive and emit messages are not participants, even if they belong to one of the parties that's interacting. What's at issue here is the definition of "participant" -- and we can't define it without explaining it accurately.

concepts/0003-protocols/roles-participants-etc.md
dhh1128: Disagree.

concepts/0003-protocols/uris.md
dhh1128: >We don't want non-core implementers to use those formats.

This sentence tells me that we have a fundamentally different assumption about purpose. We don't write Aries RFCs to guide core implementers only; we write them to guide all implementers, everywhere. Some implementers will be core implementers, and they should use the patterns we chosen for Aries core stuff. But some implementers won't be core implementers. They need to know how to follow guidelines even if their context is different. This RFC is guidance for all protocol implementers, not just core implementers.

concepts/0003-protocols/state-details.md
dhh1128: Okay, I can get on board with that.

swcurran:
@dhh1128 - As neither of us have time for this, I'll revert the updates you don't like.  I still lean to the changes improving the content, but it's not worth debating.

The important things are to clear up the section on semver, and (IMO) to get the content into a single document so that important details are not missed be readers.

concepts/0003-protocols/roles-participants-etc.md
swcurran: You disagree that I find it confusing?  :-)

concepts/0003-protocols/roles-participants-etc.md
swcurran: Reverted

concepts/0003-protocols/roles-participants-etc.md
swcurran: Reverted

concepts/0003-protocols/roles-participants-etc.md
swcurran: Reverted

concepts/0003-protocols/roles-participants-etc.md
swcurran: Reverted

concepts/0003-protocols/roles-participants-etc.md
swcurran: Reverted

concepts/0003-protocols/semver.md
swcurran: Clarification added.

concepts/0003-protocols/semver.md
swcurran: Added as "Semver Examples".

concepts/0003-protocols/semver.md
swcurran: Added back

concepts/0003-protocols/semver.md
swcurran: How is that a change?  I had made the statement that "all minor versions of a protocol must be supported", and someone (I thought you) had objected that an agent may refuse to support a minor version below a given number.  That is what is being conveyed.

In ACA-Py, we are putting into code:

- each major version of each protocol and for each:
    - the minimum minor version allowed
    - the supported/default minor version
    - a link to the commit in github of the RFC for the protocol major version.

All of that is our way to specify that information.  Others of course can do something different or nothing.

Changed the wording to "respond to" from "specify" in this and the next line.

concepts/0003-protocols/semver.md
swcurran: Reverted.

concepts/0003-protocols/semver.md
swcurran: Removed the previous lines.

concepts/0003-protocols/uris.md
swcurran: Reverted.  Still haven't a clue what to make of the examples.

concepts/0003-protocols/README.md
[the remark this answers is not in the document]
swcurran: Reverted

concepts/0003-protocols/README.md
```diff
@@ -141,42 +178,37 @@ a superprotocol from another (as when protocols are nested at least 3 deep).
 ![super- and subprotocols](super-sub.png)
 
 Commonly, protocols wait for subprotocols to complete, and then they continue.
-A good example of this is [ACKs](../../features/0015-acks/README.md),
-which are often used as a discrete step in a larger flow.
+A good example of this is mentioned above&mdash;starting an issue credential flow,
+but requiring the presentation of a proof before completing the process.
```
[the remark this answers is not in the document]
swcurran: Changed

swcurran:
Requested changes made.

concepts/0003-protocols/roles-participants-etc.md
dhh1128: I'm not disagreeing that you find it confusing; obviously you're the world authority on that assertion. :-) I'm disagreeing with the judgment call that says, "take it out, because the confusion is getting in the way of the goal." I'm claiming that if the example doesn't make sense, it's because of a deep problem in understanding. The example must be grappled with. If it can't be processed comfortably, then subtle misunderstandings remain. Therefore, a better change would be to rewrite the section such that it makes perfect sense to you -- not to delete it.

concepts/0003-protocols/semver.md
dhh1128: When I read the word "specify" I assumed you were talking about some behavior in the protocols themselves -- that something in messages sent by an agent would specify which versions it supported. We have no standard way of doing this, so when I read it that way, I got concerned. If all you meant by "specify" was "decide and document" then I am cool with that word.

Regarding substantive change: the new and simplified text changes the intent of this section from being about which version of a protocol is used, to being about which version of a protocol elicits a compatible response. Those are not the same thing. The effect of the original text would be to advocate the following behavior:

>A sends v 1.3 of a message to B. B, being more up-to-date, prefers version 1.5. However, B responds, and the response uses version 1.3, because the protocol version has been negotiated down to the least common denominator. Both A and B know and explicitly acknowledge that no features above 1.3 will take effect in the interaction.

The updated text allows the following:

>A sends v 1.3 of a message to B. B, being more up-to-date, prefers version 1.5. However, B responds. The response uses version 1.5. Now A has to decide whether to support 1.5 or not. The protocol version is still unclear; are they both speaking the same version of the protocol? Is B aware that any features in versions >1.3 are being dropped, or were they just careless about inspecting versions?

dhh1128:
@swcurran : I 'm content to merge this. However, you said in your initial comment:

>Before merging this, the external documents (e.g. uris.md, etc.) will be removed, and references to those documents in other RFCs will be updated.

Do you want to make a final round of changes to do that?

Also, there are some broken links at this point. These may organically resolve as you do the final pass, or they may need specific attention:

```sh
concepts/0003-protocols/uris.md:                                                
    [semver rules](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
    [should not change in...](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
concepts/0003-protocols/README.md:                                              
    [Semver Rules for Pro...](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
    [semver-compatible
  ...](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
    [semver section](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
    [semver rules](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
    [should not change in...](#semver-rules-for-protocols) #semver-rules-for-protocols not in text/markdown content
```

swcurran:
Sounds good.  I'll run the check links script to make sure I catch everything.

swcurran:
I think all the needed changes have been made - deleting the now obsolete files and updating the links in other RFCs.

PR #486, opened 2020-05-28
Present proof v2

swcurran:
Updates from v1 to v2 - the following lists the categories of changes in the PR:

- Added version change log
- Reorganized (a little) to match the new protocol template
- New wording to reflect that this work supersedes previous work 
- Added the `formats` item to all messages as used in the 0453 - Issue Credential
- Added an "Attachment Registry" section to each message; moved Indy specific material into the Indy section within the Registries.
- Added note that the attachment can be either JSON or base64, and used both in the different examples.

Substantive changes that should be carefully considered:

- add the attachment on "propose" is optional when the only issue is the format of the verifiable presentation, and not the content of the presentation request.
- made either Ack or Problem Report as **required** after the "presentation" message.

Several other To Do's are in the document and should be addressed and removed before merging.


features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,347 @@
...
+- There might need to be a way to associate a payment with the present proof protocol.
+- A number of `To Do` items are embedded in the content of the RFC.
+
+## Implementations
+
+The following lists the implementations (if any) of this RFC. Please do a pull request to add your implementation. If the implementation is open source, include a link to the repo or to the implementation within the repo. Please be consistent in the "Name" field so that a mechanical processing of the RFCs can generate a list of all RFCs supported by an Aries implementation.
+
+Name / Link | Implementation Notes
+--- | ---
+ |  |
```
dhh1128, line 347: This should have 1 vertical bar, not 2.

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,347 @@
...
+
+If the `proposal~attach` is not provided, the `attach_id` item in the `formats` array should not be provided. That form of the `propose-presentation` message is to indicate the presentation formats supported by the prover, independent of the verifiable presentation request content.
+
+#### Negotiation and Preview
+
+Negotiation prior to the delivery of the presentation can be done using the `propose-presentation` and `request-presentation` messages. The common negotiation use cases would be about the claims to go into the presentation and the format of the verifiable presentation.
+
+#### Propose Attachment Registry
+
+##### Hyperledger Indy Proposal Attachment
```
dhh1128, line 142: V1 of this protocol has been criticized as being Indy-centric. We know that was not the intent, but that's the way it reads to a non-Indy audience. It would be nice to include some a section for something besides Indy. If we can't add such a section because we don't know what it would contain, can we create a placeholder with "TODO" text? And can we put it before Indy?

dhh1128:
Minor request: can we add to this PR a quick update of RFC 0037 noting that it is now superseded by this version of the protocol?

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+    ]
+}
+```
+
+Description of fields:
+
+* `comment` -- a field that provides some human readable information about the proposed presentation.
+* `formats` -- contains an entry for each `filter~attach` array entry, including an optional value of the attachment `@id` (if attachments are present) and the verifiable presentation format and version of the attachment. Accepted values for the `format` items are provided in the per format "Attachment" sections immediately below.
```
soluchok, line 129: Please replace `filter~attach` with `proposal~attach`.

I see that the previous version has [presentation-preview](https://github.com/hyperledger/aries-rfcs/tree/master/features/0037-present-proof#presentation-preview) message.
Also, I noted that issue-credential-v2 has [preview-credential](https://github.com/hyperledger/aries-rfcs/tree/master/features/0453-issue-credential-v2#preview-credential) message.

Since those protocols are similar I assume that we missed something. I see two possible options:
1. We need to remove a preview-credential message from the issue-credential-v2.
2. We need to add a presentation-preview to the present-proof-v2.

Because we have attachments and formats It looks like the `1` option is what we need.
Please confirm whether I am right.

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,345 @@
...
+            "cred_def_id": "<cred_def_id>",
+            "predicate": "<predicate>",
+            "threshold": <threshold>
+        },
+        // more predicates
+    ]
+}
+```
+
+> To Do: Is this structure sufficient to achieve the desired level of negotiation?
```
tmarkovski, line 173: Starting a discussion here as we have just implemented v1 of this protocol. This message format seems to be a configuration for how the verifier should construct the proof request. Why not simply allow the proposer to submit an actual proof request format instead (with updated nonce)? It would clearly communicate the proposed intent, and add flexibility for fill restriction set per the WQL format, and non-revocation. Non-revocation is important, as the holder can propose a time they know they have a valid credential for.
I'm curious if this has been discussed before and what the pros and cons of using the same structure.

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,345 @@
...
+
+```json
+{
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/%VER/propose-presentation",
+    "@id": "<uuid-propose-presentation>",
+    "comment": "some comment",
+    "formats" : [
+        {
+            "attach_id" : "<attach@id value>",
+            "format" : "<format-and-version>",
```
llorllale, line 111: @swcurran the purpose of `format` seems to be to communicate the exact schema of the attachment's contents. This is useful in other protocols as well, so why not add a new field to the attachments RFC?

dhh1128: That's an interesting idea that I had not considered.

My mental model for attachments was heavily influenced by email attachments. There, the only typing you need is a MIME type. However, I knew we needed some extra precision in DIDComm, so I added the ability to connect a specific attachment to a specific field in a DIDComm message (e.g., in an insurance claim, you might want to say, "pictures of wind damage on the roof of my shed are in attachment #3" in a machine-readable way). So we have that.

What you're proposing here is attachments getting an additional "format" property that isn't a type, exactly; rather, it's a fine-grained distinction that splits one MIME type into many similar sub-formats. We probably didn't think of this before because we assumed only VC attachments--not photos or office docs or medical data--needed the extra level of detail. If we can satisfy ourselves that it's generally useful, then I guess I'm not opposed to updating the properties of attachments.

I'd be curious to hear Stephen's opinion.

llorllale: @dhh1128 even within VCs we have multiple schemas.

It's not enough to parse a VC generically and validate its proof. Oftentimes we need to parse it and extract data from it, and this is typically from custom, well-defined schemas.

And you're right, attachments can contain anything.

The second point is that this same structure is duplicated in issue-credential V2. So it seems extra-obvious to put this in the attachments RFC.

features/0454-present-proof-v2/README.md
llorllale: @swcurran there does seem to be an inconsistency. Is this intentional?

features/0454-present-proof-v2/README.md
dhh1128: The argument about VC schema doesn't feel convincing to me yet, but maybe say more about it in case I'm missing something? Of course we have different schemas for different VCs -- but we have a mechanism for communicating the schema (@context and/or the credential def from Indy); we don't need to declare it somewhere else. And besides, the value we would put into "format" isn't the schema of the cred anyway; it's something about whether the cred is JSON-LD-centric, JWT-centric, etc -- is it not? 

What would convince me for sure is what I said before: "if we can satisfy ourselves that [the mechanism is] generally useful." And we'd have to carefully explain what we intend by "format" and how it is different from MIME type.

llorllale: @dhh1128 sure - the VC example was poor.

Consider just a JSON object you're trying to get across. MimeType="application/json", format=??

Consider a string representation of something you're sending (eg. a DID (not a did doc)?). MimeType="text/plain", format=??

llorllale: I realized I'm mixing thoughts about how to express the purpose of an attachment vs its format for my use case. I've opened #505 to continue discussion on the purpose of attachments.

I still suspect there's value in this format idea.

> the value we would put into "format" isn't the schema of the cred anyway; it's something about whether the cred is JSON-LD-centric, JWT-centric, etc -- is it not?

It's the schema. From this PR:

> The Hyperledger Indy v1.0 propose filter attachment schema is as follows:

llorllale: To be clear: while I think this format idea probably has merits, this suggestion I've raised is in no way a blocker of this proposal as far as I am concerned.

features/0454-present-proof-v2/README.md
swcurran: Fixed

features/0454-present-proof-v2/README.md
swcurran: This will be added shortly (follow on PR) to include the DIF presentation definition.  Leaving off for now.

swcurran:
> Minor request: can we add to this PR a quick update of RFC 0037 noting that it is now superseded by this version of the protocol?

Update added to 0037 header.

features/0454-present-proof-v2/README.md
swcurran: Fixed the `filter~attach` issue.  The "presentation preview" content was initially moved over to the v2 protocol under a different name.  Per @tmarkovski suggestion below, we discussed switching from this "presentation preview" format to just using an Indy "presentation-request" data structure (same as for the request-presentation message.  This was discussed on the Aries WG call on 2020.06.24 and all agreed that's what we should do.  So, "presentation preview" is now gone in both name and content, replaced in the proposal message with the presentation request.

I think that resolves this question.  Note that we definitely don't want to remove the preview credential from the issue-credential propose message.

features/0454-present-proof-v2/README.md
swcurran: This was discussed on the Aries WG call on 2020.06.24 and all agreed that's what we should do. So, "presentation preview" is now gone in both name and content, replaced in the proposal message with the presentation request.

swcurran:
There remain three "To Do" items in the PR.  Do we want to resolve them before or after merging?

The first (definition of "done") is to be addressed in a follow up PR I have agreed to do about "verify" and "validate" and that at the protocol level a successful "done" means "validate" -- accepted at a business level.

The second is about problem-reports and the protocol state after a problem-report is sent.

The last is whether a message from the verifier to transition to "Done" is required or optional.

Thoughts?   @TelegramSam @dhh1128 @andrewwhitehead @tmarkovski @llorllale @smithbk

dhh1128:
>The last is whether a message from the verifier to transition to "Done" is required or optional.

Obviously we *want* to know that the verifier is done. The question is whether we should *require* it.

My first instinct was to say yes. It would provide such nice closure. But then I started thinking, "If we do that, then the verifier and the prover both have to stay online longer. I bet there are use cases where a prover wants to go offline as soon as proof is sent, or a verifier wants to go offline as soon as proof is received."

Then I thought, "Well, here is a perfect use case for ~please_ack." If a prover wants closure, they can ask for it.

Then I thought, "But how does a prover say, 'please DON'T ack'?" Or how does a verifier say, "I want you to prove this, and BTW, I have no plans of ACK'ing your proof"?

So now I'm not sure.

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+The present proof protocol consists of these messages:
+
+* `propose-presentation` - Prover to Verifier (optional) - propose a presentation or send a counter-proposal in response to a `request-presentation` message
+* `request-presentation` - Verifier to Prover - request a presentation
+* `presentation` - Prover to Verifier - provide a presentation in response to a request
+
+In addition, the [`ack`](../0015-acks/README.md) and [`problem-report`](../0035-report-problem/README.md) messages are adopted into the protocol for confirmation and error handling.
+
+The messages that include `~attach` attachments may use either form of the embedded attachment&mdash;`base64` or `json`. In the examples below, one or the other is arbitrarily used.
```
llorllale, line 95: What's the reason for this restriction? Why is `links` excluded?

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+* A similar situation exists with the final transitions for the Prover and its final transitions (Send Presentation Reject, Receive Presentation Ack, and Receive Presentation Reject).
+* When a Prover makes a (counter-)proposal, it transitions to the `proposal-sent` state. This state is only present by implication in the choreography diagram; it essentially equates to the null or begin state in that the Prover does nothing until a presentation request arrives, triggering the leftmost transition for the Prover.
+
+Errors might occur in various places. For example, a Verifier might time out waiting for the Prover to supply a presentation. Errors trigger a `problem-report`. In this version of the protocol, all errors cause the state of both parties (the sender and the receiver of the `problem-report`) to revert to `null` (meaning it is no longer engaged in the protocol at all). Future versions of the protocol may allow more granular choices.
+
+> To Do: Should the previous sentence be addressed in this version? Should it be removed?
+
+### Choreography Diagram
+
+![present proof](credential-presentation.png)
```
llorllale, line 81: The verifier's "Verify `presentation`" activity should be substituted for a decision point, eg. "Is `presentation` valid?" with Yes or No as outcomes.

llorllale: Is "Receive Presentation Reject" [here](https://github.com/hyperledger/aries-rfcs/pull/486/files#diff-6f5d41c9a81691c26389a4e0172b90a1R68) the "No" outcome for my suggestion above ?

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+* request-received
+* proposal-sent
+* presentation-sent
+* abandoned
+* done
+
+For the most part, these states map onto the transitions shown in the choreography diagram in obvious ways. However, a few subtleties are worth highlighting:
+
+* The final transitions for the verifier (Send Presentation Reject, Send Presentation Ack, and Receive Presentation Reject) all result in a final `done` state, but this `done` may or may not be the verifiers intended outcome. We may want to tweak this in a future version of the protocol.
```
llorllale, line 68: > Send Presentation Reject, Send Presentation Ack, and Receive Presentation Reject

What is the difference between "send presentation reject" and "receive presentation reject"? The mismatch of terms between the diagram and this sentence is confusing and makes it hard to compare the two.

Does "send presentation reject" refer to the verifier's transition "send request-presentation" -> "receive problem-report"?

Does "receive presentation reject" refer to a transition not illustrated in the diagram where the presentation is received and yet is not valid?

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+> To Do: Should the previous sentence be addressed? Removed?
+
+* A similar situation exists with the final transitions for the Prover and its final transitions (Send Presentation Reject, Receive Presentation Ack, and Receive Presentation Reject).
+* When a Prover makes a (counter-)proposal, it transitions to the `proposal-sent` state. This state is only present by implication in the choreography diagram; it essentially equates to the null or begin state in that the Prover does nothing until a presentation request arrives, triggering the leftmost transition for the Prover.
+
+Errors might occur in various places. For example, a Verifier might time out waiting for the Prover to supply a presentation. Errors trigger a `problem-report`. In this version of the protocol, all errors cause the state of both parties (the sender and the receiver of the `problem-report`) to revert to `null` (meaning it is no longer engaged in the protocol at all). Future versions of the protocol may allow more granular choices.
```
llorllale, line 75:
```suggestion
Errors might occur in various places. For example, a Verifier might time out waiting for the Prover to supply a presentation. Errors trigger a `problem-report`. In this version of the protocol, all errors cause the state of both parties (the sender and the receiver of the `problem-report`) transition to `abandoned` (meaning it is no longer engaged in the protocol at all). Future versions of the protocol may allow more granular choices.
```

features/0454-present-proof-v2/README.md
llorllale: @swcurran "`filter~attach`" is still there, ie. the line pointed out by @soluchok has not been fixed.

features/0454-present-proof-v2/README.md
llorllale: I think this may have been left as a TODO [here](https://github.com/hyperledger/aries-rfcs/pull/486#issuecomment-649129451)? @swcurran correct me if I'm wrong

llorllale:
@swcurran 

> The last is whether a message from the verifier to transition to "Done" is required or optional.

Requiring it would encumber ephemeral challenges.

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+The `~attach` array is to be used to enable a single presentation to be requested/delivered in different verifiable presentation formats. The ability to have multiple attachments must not be used to request/deliver multiple different presentations in a single instance of the protocol.
+
+### Propose Presentation
+
+An optional message sent by the prover to the verifier to initiate a proof presentation process, or in response to a `request-presentation` message when the prover wants to propose using a different presentation format or request. Schema:
+
+```json
+{
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/%VER/propose-presentation",
```
llorllale, line 105: wrong `@type`

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+The Hyperledger Indy v1.0 proposal attachment contains an Indy presentation request. The structure is the same as used in the Indy attachment to the `request-presentation` message. The structure comes directly from libindy and is base64-encoded. For more information see the [Libindy API](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L1214).
+
+### Request Presentation
+
+From a verifier to a prover, the `request-presentation` message describes values that need to be revealed and predicates that need to be fulfilled. Schema:
+
+```json
+{
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/%VER/request-presentation",
```
llorllale, line 154: wrong `@type`

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,289 @@
...
+
+The Hyperledger Indy v1.0 presentation request attachment contains data from libindy about the presentation request, base64-encoded, as returned from `libindy`. For more information see the [Libindy API](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L1214).
+
+### Presentation
+
+This message is a response to a Presentation Request message and contains signed presentations. Schema:
+
+```json
+{
+    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/%VER/presentation",
```
llorllale, line 197: wrong `@type`

features/0454-present-proof-v2/README.md
dhh1128: I think this was just a missing detail, and that `links` would still be valid, too. But let's see what others say.

swcurran: I had not thought the links style would be used because of the contents of some of the messages. Anything that is hash/linked is easily shared (accidentally or on purpose) by the message receiver. That is the same for embedded data, it's just (feels) easier with a URL. Not a strong argument :-).

Note that the attachment could be embedded using JSON which in turn contains a hashlink (or hash and link). 

I'm ok with adding it if others are good with it (and OK with leaving it out).

If added here, it should also be added to the v2.0 Issue Credential, right?

features/0454-present-proof-v2/README.md
swcurran: Based on the last Aries WG call (20200624) call, I think we want to add a paragraph on "Verify" (cryptographic) and an optional "Validate" (business rules) and suggest that the outcome from the Verifier be the outcome of the validation.

I think that is what you are saying as well in this comment, right @llorllale ?

I will put that into the RFC, and perhaps update the picture if I can figure out how.  Yes, I know, there are directions :-).

features/0454-present-proof-v2/README.md
swcurran: I didn't play with that section in V2 and will have another look.

features/0454-present-proof-v2/README.md
swcurran: Will make the change.  Can we drop the last sentence since that is true of every word and phrase in the RFC?

swcurran:
> @swcurran
> 
> > The last is whether a message from the verifier to transition to "Done" is required or optional.
> 
> Requiring it would encumber ephemeral challenges.

Not necessarily. The Trinsic folks are doing an ephemeral issue credential, so it is possible to have an ephemeral multi-message protocol. Plus this gets us out of the "using an existing connection" conundrum we are discussing about the OOB + reuse PR.

features/0454-present-proof-v2/README.md
swcurran: Fixed in next commit.

features/0454-present-proof-v2/README.md
swcurran: Fixed in next commit

features/0454-present-proof-v2/README.md
swcurran: Fixed in next commit

features/0454-present-proof-v2/README.md
swcurran: Fixed in next commit - removed the final sentence and the related "To Do".

features/0454-present-proof-v2/README.md
swcurran: As I interpret this, the "send presentation reject" is sent by the verifier and "receive presentation reject" is received by the verifier.  I'll reword to make that clear and that those send the state to "abandoned", not "done" as a terminal state.

Ah...I see.  You are thinking of "receive presentation reject" as another "receive presentation AND reject"?  Which I think is "send presentation reject".  Sigh... :-)

features/0454-present-proof-v2/README.md
llorllale: I think it should be included.

> If added here, it should also be added to the v2.0 Issue Credential, right?

Yes, for consistency.

features/0454-present-proof-v2/README.md
llorllale: > Based on the last Aries WG call (20200624) call, I think we want to add a paragraph on "Verify" (cryptographic) and an optional "Validate" (business rules) and suggest that the outcome from the Verifier be the outcome of the validation.

OK, but I'm not sure of making that distinction in the RFC...

features/0454-present-proof-v2/README.md
llorllale: > Can we drop the last sentence since that is true of every word and phrase in the RFC?

If you're referring to this then sure:

> Future versions of the protocol may allow more granular choices.

features/0454-present-proof-v2/README.md
dhh1128: Yes, the two should be consistent. But we shouldn't be adding something to allow this; we should be subtracting something -- specifically, the verbiage that says we only support a subset of the attachment capability. Let's just say that we use attachments -- and if clarification is needed, we can add "in all their forms."

llorllale: > Yes, the two should be consistent. But we shouldn't be adding something to allow this; we should be subtracting something -- specifically, the verbiage that says we only support a subset of the attachment capability. Let's just say that we use attachments -- and if clarification is needed, we can add "in all their forms."

Agreed.

swcurran:
My apologizes - I thought I had pushed all but the "attachment" type changes.  I think all the comments are covered from above, but @llorllale and @dhh1128 might want to do another pass.

I deliberately left off the DIF Presentation Request format from this because I haven't looked at it and would like to confirm it has the features necessary.  Of course, others that know more or welcome to add a PR to include it.

Love to have @llorllale follow up with attachment formats in the "Registry" sections of this document, once this PR has been merged.

Thanks -- sorry again for the delay.

features/0454-present-proof-v2/README.md
```diff
@@ -102,7 +100,7 @@ An optional message sent by the prover to the verifier to initiate a proof prese
 
 ```json
 {
-    "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/%VER/propose-presentation",
+    "@type": "http://didcomm.org/present-proof/%VER/propose-presentation",
```
llorllale, line 103: nit: it's "https"

features/0454-present-proof-v2/README.md
```diff
@@ -0,0 +1,288 @@
...
+- The details that are part of each message type about the different attachment formats serves as a registry of the known formats and versions.
+- Version 2.0 uses &lt;angle brackets&gt; explicitly to mark all values that may vary between instances, such as identifiers and comments.
+
+## Summary
+
+A protocol supporting a general purpose verifiable presentation exchange regardless of the specifics of the underlying verifiable presentation request and verifiable presentation format.
+
+## Motivation
+
+We need a standard protocol for a verifier to request a presentation from a prover, and for the prover to respond by presenting a proof to the verifier. We doing that exchange, we want to provide a mechanism for the participants to negotiate the underlying format and content of the proof.
```
llorllale, line 25:
```suggestion
We need a standard protocol for a verifier to request a presentation from a prover, and for the prover to respond by presenting a proof to the verifier. When doing that exchange, we want to provide a mechanism for the participants to negotiate the underlying format and content of the proof.
```

features/0454-present-proof-v2/README.md
llorllale: @swcurran I think this section (including the choreography diagram) needs cleanup. State transitions are critical for a protocol RFC, and there seem to be a few missing here as we've discussed above.

The existing diagram might get too cluttered if you add them. Consider perhaps switching to a state matrix.

llorllale:
@swcurran 

> I deliberately left off the DIF Presentation Request format from this because I haven't looked at it and would like to confirm it has the features necessary. Of course, others that know more or welcome to add a PR to include it.
> 
> Love to have @llorllale follow up with attachment formats in the "Registry" sections of this document, once this PR has been merged.

I opened #510 for this. I think the formats should be separate RFCs.

swcurran:
I knew I shouldn't have touched the choreography diagram. I used the draw.io github integration and it did some commits without signing.  Sigh...I wound up just deleting the files it added anyway.  Anyone know how to clean that up?  The changes were in my fork, but not my clone.  Maybe just open a new PR?

@llorllale @tmarkovski @TelegramSam @KitHat 

I added the state tables as requested, and adjusted the choreography diagram accordingly -- to get the state names aligned.  I labelled "abandoned" and "done" as terminal states, and did NOT propose that a problem report can go to some other state.

I added a field to the presentation-request "will_confirm" (from the verifier) so that the holder will know to expect (or not) the "presentation-ack" message.  I also added a note that the prover can add a "~please_ack" decorator to get an ack from the verifier.

In both  cases, the "presentation-ack" is the **business** result of the validation, not the cryptographic verification (which is of course included).

Please review these changes and let me know.  Would love feedback on the ack changes.

I still have the changes discussed today to the attachment registries and will do that by tomorrow.

llorllale:
@swcurran 

> I added a field to the presentation-request "will_confirm" (from the verifier) so that the holder will know to expect (or not) the "presentation-ack" message. I also added a note that the prover can add a "~please_ack" decorator to get an ack from the verifier.
> 
> In both cases, the "presentation-ack" is the business result of the validation, not the cryptographic verification (which is of course included).

I'd say the business result is whatever outcome the Prover wanted to achieve just by executing this protocol in the first place. I don't see the need for a "presentation-ack". The actual outcome - whether it's the display of a screen, issuance of a token, w/e - should follow from the `presentation` message.

swcurran:
> I'd say the business result is whatever outcome the Prover wanted to achieve just by executing this protocol in the first place. I don't see the need for a "presentation-ack". The actual outcome - whether it's the display of a screen, issuance of a token, w/e - should follow from the presentation message.

Yours is one view, and an equally acceptable view is that the protocol should allow the verifier to tell the prover what they thought of the proof -- was it sufficient.  That, for example, allows for a pure agent-to-agent interaction without out-of-band action (or even in addition to some out of band action).  That's why I propose adding the "will_confirm" field, so both flows can be easily supported.  The choreography has always indicated that an ACK should go back, but I agree that your view is valid that in some cases it is not needed and this is my proposal to rationalize those views.  I think it is a good one.

dhh1128:
Here's a simple use case that teases out the difference in your perspectives.

Suppose a radio station says, "We're giving away a free vacation to the Bahamas to the first person who can prove they are a resident of our town." The proof request is "Prove that you're a resident of our town." It CAN'T include "the first person" because this is a criterion only the radio station can evaluate. Alice and Bob both start the proving protocol. Alice finishes first and gets a result that says the proof was accepted; Bob finishes second and -- even though his proof was perfectly valid -- gets a result that says the proof was declined.

I think that being able to distinguish these two outcomes is useful -- AND it is useful to know whether the proof validated. Bob's proof validated; he just didn't satisfy some business constraints that are outside the bounds of the pure proving. Not sure what to do about this, but I think Stephen's idea is headed in the right direction.

jstclair2019:
“an equally acceptable view is that the protocol should allow the verifier to tell the prover what they thought of the proof -- was it sufficient"
Good point. Allows room for possible “risk scoring” and machine readable governance framework, so a feedback loop has rick mgmt. value for verifier/prover IMO

Best Regards,
________________________
Jim St. Clair
Chief Technology Officer
Dinocrates Group LLC
One Research Court, Suite 450
Rockville, MD 20850
Office: 240-403-4103
Mobile: 228-273-4893<tel:(228)%20273-4893>
jim.stclair@dinocrates.com<mailto:jim.stclair@dinocrates.com>
www.dinocrates.com<http://www.dinocrates.com>
Calendly link - https://calendly.com/jim-stclair<https://via.intercom-mail-100.com/e?ob=2it%2FEFwfiO0RnjFn65EB2QRujZBxGJaSUK4x3Adx5BN1UfRAPUb2nyP0ckyG%2FAKh&h=8e0e3ad8c52dd472bbb90ea5991927dbde1c3dd3-17690316241>
[d,d]
[cid:image009.png@01D660EE.D7200090]
[cid:image010.png@01D660EE.D7200090]







*******************************************************************************
This message (including any attachments) contains confidential information intended for a specific individual and purpose, and is protected by law. If you are not the intended recipient, you should delete this message and any disclosure, copying, or distribution of this message, or the taking of any action based on it, by you is strictly prohibited.

llorllale:
@swcurran @dhh1128 @jstclair2019 

Sorry, I guess I'm missing something here.

I'm not entirely convinced by Daniels scenario: if Alice loses she'll get a problem-report back (we already allow for this). If she wins she'll get *something* back (an email? a message from another protocol?).

To the point about risk scoring and machine-readable governance framework: I think the same thing applies. Whats more, Stephen's change here proposes that the *Prover* sets the terms on whether a response is required or not. In many cases in the real world, the Verifier will NOT wish to disclose the "risk score" they measured in the proof. I get the sense this belongs in a different (perhaps sub?) protocol.

swcurran:
> @swcurran @dhh1128 @jstclair2019
> 
> Sorry, I guess I'm missing something here.
> 
> I'm not entirely convinced by Daniels scenario: if Alice loses she'll get a problem-report back (we already allow for this). If she wins she'll get _something_ back (an email? a message from another protocol?).
> 
> To the point about risk scoring and machine-readable governance framework: I think the same thing applies. Whats more, Stephen's change here proposes that the _Prover_ sets the terms on whether a response is required or not. In many cases in the real world, the Verifier will NOT wish to disclose the "risk score" they measured in the proof. I get the sense this belongs in a different (perhaps sub?) protocol.

@llorllale your last point is not correct.  In my proposal the verifier says "I will confirm" --- telling the prover to expect an ACK.  I did add that the prover could use the "~please_ack" decorator, but the verifier is not required to respond to that -- it's a ask, not a requirement.

llorllale:
@swcurran 

> > @swcurran @dhh1128 @jstclair2019
> > Sorry, I guess I'm missing something here.
> > I'm not entirely convinced by Daniels scenario: if Alice loses she'll get a problem-report back (we already allow for this). If she wins she'll get _something_ back (an email? a message from another protocol?).
> > To the point about risk scoring and machine-readable governance framework: I think the same thing applies. Whats more, Stephen's change here proposes that the _Prover_ sets the terms on whether a response is required or not. In many cases in the real world, the Verifier will NOT wish to disclose the "risk score" they measured in the proof. I get the sense this belongs in a different (perhaps sub?) protocol.
> 
> @llorllale your last point is not correct. In my proposal the verifier says "I will confirm" --- telling the prover to expect an ACK. I did add that the prover could use the "~please_ack" decorator, but the verifier is not required to respond to that -- it's a ask, not a requirement.

Ah, I got that mixed up then.

Now, I am not too strongly opposed to this idea. I am just not seeing much ROI on this extra bit of complexity. The protocol is complex enough as it is already.

Don't let my objection block this PR.

swcurran:
@llorllale @dhh1128 -- I think I'm done.  States/Choreography cleaned up after change to make ack clearly optional. Registry table added.  I have to clean up the DCO (again!!! :-( ) and the circle/ci failing.  @dhh1128 -- if you have a moment could you peek and tell me what's wrong with the build?

Re: your last comment @llorllale about complexity.  The verifier does not have to use the ack at all. By never setting "will_confirm" and ignoring a "~please_ack", it is straight to "done".  The verifier also can't send a problem-report either though (prover is already "done").  The prover does have to handle the complexity though, and may have to clean up if the "ack" never comes...

Thanks!

swcurran:
OK - ready to go.  Build passes and (another!!) DCO cleanup done.  @dhh1128 -- have your change requests been addressed?  @llorllale approve?

Thanks.

dhh1128:
All my concerns have been resolved. I also pushed a change that addressed @llorllale 's request to fix a typo ("We" --> "When") and some code errors ("http://didcomm" --> "https://didcomm"). I also updated to pick up the latest changes from master. Given how long this PR has been open and how much effort has gone into it, and given @llorllale 's last comment about not letting his concern hold up the PR, I am going to merge it now. If we need to make subsequent tweaks, we can do that after the merge.

PR #495, opened 2020-06-05
Add replacement_id for credential replacement to credential issuance

TelegramSam:


TelegramSam:
Questions: Does this ID need to be used in more messages?

PR #502, opened 2020-06-17
Move reuse from did exchange to out of band

TelegramSam:


features/0434-outofband/README.md
```diff
@@ -130,7 +130,57 @@ If only the `request~attach` item is included, no new connection is expected to
...
 
-While the _receiver_ is expected to respond with an initiating message from a `handshake_protocols` or `request~attach` item using an offered service, the receiver may be able to respond by reusing an existing connection. Specifically, if a connection they have was created from an out-of-band `invitation` from the same public DID of a new `invitation` message, the receiver **SHOULD** use the existing connection in responding to the `invitation`. If a selected `handshake_protocols` item is used by the receiver, it may have a specific "reuse this connection" message type.
+### Reuse Messages
+
+While the _receiver_ is expected to respond with an initiating message from a `handshake_protocols` or `request~attach` item using an offered service, the receiver may be able to respond by reusing an existing connection. Specifically, if a connection they have was created from an out-of-band `invitation` from the same public DID of a new `invitation` message, the receiver **MAY** use the existing connection in responding to the `invitation`. This is accomplished using the `reuse` and `reuse-accepted` messages.
```
swcurran, line 135: I think this should only be used if there is only a `handshake_protocols` entry and no `request~attach`.  If there is a `request~attach` (even if there is also a `handshake_protocols` item) and an existing connection, the first message back should be the response to the attachment, whatever that is, using the existing connection.

features/0434-outofband/README.md
```diff
@@ -130,7 +130,57 @@ If only the `request~attach` item is included, no new connection is expected to
...
+
+- `@type` - the DIDComm message type
+- `@id` - the unique ID of the message. 
+- `pthid` - the @id of the invitation message. This provides the context link for the _inviter_ to prompt additional protocol interactions.
+
+Sending or receiving this message does not change the state of the existing connection.
+
+When the _inviter_ receives the `reuse` message, they **MUST** respond with a `reuse-accepted` message to notify that _invitee_ that the request to reuse the existing connection is successful.
+
+#### Reuse Accepted:  `https://didcomm.org/out-of-band/%VER/reuse-accepted`
```
swcurran, line 160: I would argue this is not needed.  The "done" (I think that's the name) is to complete the handshake. If the connection exists and is active, is there a need to have another confirmation message?

There is the scenario to consider if there being an existing connection, the invitation receiver sending the reuse, and the invitation sender having deleted the connection. However, I don't think that case is relevant as either:

- the invitation sender is using the public DID for the connection and understands the reuse message and can continue, or
- the invitation sender is using a peer DID for the connection, in which case the "reuse" message cannot be opened, and may not even get to the sender.

TelegramSam:
draw out that reuse messages must be encrypted.

features/0434-outofband/README.md
TelegramSam: Given the discussion in today's Aries WG call (2020-06-17), we are going to leave it here.

features/0434-outofband/README.md
TelegramSam: In today's meeting, I agreed that we should indicate this, but I think we need to confirm one thing.

The `reuse-accepted` message handles an explicit case where the _inviter_ no longer has the keys to receive a message over the existing (known to the _invitee_). 

How should that be handled in this case? If the attached message is processed, and a message sent back to the _inviter_ using a key it no longer has, do we fall in the same hole?

Even though redundant in the common case, it may help to still send a reuse message prior to engaging via the attached messages to avoid falling into the same hole.

Thoughts? @swcurran

llorllale: @swcurran 

> I think this should only be used if there is only a handshake_protocols entry and no request~attach. If there is a request~attach (even if there is also a handshake_protocols item) and an existing connection, the first message back should be the response to the attachment, whatever that is, using the existing connection.

Why?

@TelegramSam 

> How should that be handled in this case? If the attached message is processed, and a message sent back to the inviter using a key it no longer has, do we fall in the same hole?

Yes.

> Even though redundant in the common case, it may help to still send a reuse message prior to engaging via the attached messages to avoid falling into the same hole.

Agreed.

swcurran: @llorllale -- about why this: "I think this should only be used if there is only a handshake_protocols entry and no `request~attach`. If there is a `request~attach` (even if there is also a handshake_protocols item) and an existing connection, the first message back should be the response to the attachment, whatever that is, using the existing connection."

If you get a handshake AND a request attach AND you already have a connection, why not just skip to the response to the attach, and not bother with the handshake?

Sam has raised the issue that if the invitation Sender has deleted the connection, they won't be able to process the request, but (a) that's likely an 80-20 rare case and (b) the worst that happens is the message is lost and eventually someone realizes it.  I think we should optimize for the common case.

I'm reluctantly OK with going through the reuse dance every time (two extra messages!!! The horror!!!  :-) ) But if that's what people want, we can do that.  I don't think it's the end of the world either way.

swcurran: BTW --- what do we do if we have a connection and there is ONLY the `request~attach` message?  We get into the same hole, but with no "reuse", "confirmed" message pair to fall back on.  Do we accept that as OK?

What about this approach:

- If the `request~attach` response in turn has a response coming back to the invitee, the reuse/confirmed message pair is really a waste, so we don't use it.
- If there is not a response expected to the `request~attach` response, we recommend the use of `~please_ack` with the `request~attach` response so that we get a response if all is well.

Between the two, we achieve the same result in all cases and bypass the extra messages reuse/confirmed.

@TelegramSam @llorllale

TelegramSam:
TODO: Only use the reuse mechanism if no request~attach is present.

TelegramSam:
Updated language as discussed. @swcurran

PR #508, opened 2020-07-02
IBM completed step 1

smithbk:
Signed-off-by: Keith Smith <bksmith@us.ibm.com>


swcurran:
@dhh1128 --- see the failure above in the build script.  I don't understand the error.  Can you please take a look at it?  Everything looks fine to me.

dhh1128:
The test error was that when agents are declared in an impl table, they have to have a consistent name and URI across all impl tables in all RFCs. Keith's change used https://github.com/hyperledger/indy-sdk as the URI for IBM's agent; that URI was used as the URI for Indy SDK's impl in other RFCs. I fixed this by creating a URI for IBM's agent that didn't involve Indy SDK. (I first tried to use https://github.com/hyperledger/indy-sdk/libvcx as the URI, but that didn't work either, because the comparison is based on a base URI that clips off the "libvcx" part. So, Keith, feel free to submit a PR that uses an improved URI for IBM's agent. In the meantime, what I did will work.)

PR #509, opened 2020-07-08
Action menu

jcourt562:
Proposed draft of the action-menu protocol implemented in ACA-py and described at the following location :
https://hackmd.io/jrWg2kw7QuWx9XS0R3_XPQ?view


features/0509-action-menu/README.md
```diff
@@ -0,0 +1,211 @@
+# Aries RFC 0509: Action Menu Protocol 0.9
+
+- Authors: [Andrew Whitehead],[Stephen Curran]
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2019-12-26 (date you submit your PR)
```
swcurran, line 5: Please review any template instructions, and update the dates in the header.

features/0509-action-menu/README.md
```diff
@@ -0,0 +1,211 @@
+# Aries RFC 0509: Action Menu Protocol 0.9
+
+- Authors: [Andrew Whitehead],[Stephen Curran]
```
swcurran, line 3: Definitely include your own name here :-)!

features/0509-action-menu/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Messages
+
+#### menu
+
+A `requestor` is expected to display only one active _menu_ per connection when action menus are employed by the `responder`. A newly received _menu_ is not expected to interrupt a user, but rather be made available for the user to inspect possible actions related to the `responder`.
+
+```jsonc
+{
+  "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/action-menu/1.0/menu",
```
swcurran, line 67: There is a new form for types --- https://didcomm.org/action-menu/%VER/menu

The Version of the protocol (1.0, in this case) should be defined just after the header and then referenced with the "%VER".  That's so we only have one copy of the version.

features/0509-action-menu/README.md
```diff
@@ -0,0 +1,211 @@
...
+## Summary
+
+The action-menu protocol allows one Agent to present a set of heirarchical menus and actions to another user-facing Agent in a human friendly way. The protocol allows limited service discovery as well as simple data entry. While less flexible than HTML forms or a chat bot, it should be relatively easy to implement and provides a user interface which can be adapted for various platforms, including mobile agents.
+
+## Motivation
+
+Discovery of a peer Agent's capabilities or service offerings is currently reliant on knowledge obtained out-of-band.  There is no in-band DIDComm supported protocol for querying a peer to obtain a human freindly menu of their capabilities or service offerings. Whilst this protocol doesn't offer ledger wide discovery capabilities, it will allow one User Agent connected to another, to present a navigable menu and request offered services. The protocol also provides an interface definition language to define action menu display, selection and request submission.
+
+## Tutorial
+
```
swcurran, line 20: Should add the protocol name and version here:

```
### Name and Version

`present-proof`, version 2.0
```

features/0509-action-menu/README.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+##### Description of attributes
+
+* `name` -- the _menu_ option being requested. This is taken from the `name` attribute of the `options` array elements in the _menu_  
+* `params` -- optional dictionary containing any input parameters requested in a _menu_ option `form` section.  The dictionary key values are taken from the `name` attributes of `params` array elements in the _menu_ option `form`.
+
+## Drawbacks
+
+Why should we *not* do this?
```
swcurran, line 182: Same comment as above with removing the template instructions and putting in a blanks or "None" or "N/A" or something like that.

swcurran:
The Circle CI is failing because the state is proposed, but there is an implementation.  How about changing the state to "Demonstrated"?

swcurran:
@tmarkovski @CHempel-esatus --- what are your thoughts on implementing this protocol in your mobile agents?   I think this is a very powerful capability to allow an enterprise type agent to be able to send to a user on a mobile agent a menu of things the user can do.

tmarkovski:
I agree, this would be a cool feature to support in a mobile agent. I would categorize this together with basic messaging feature: both are really nice and extend Aries beyond the credential exchange.
My concern, from practical application, is about broad implementation for this feature, specifically getting backends ready to build and interact with dynamic menus. There seems to be a considerable amount of work to support this properly for end users, both on backend and frontend, that I always have to weigh in how much will this improve the overall adoption of SSI and push that agenda forward, before making commitment to support it.

The protocol itself isn't complex to add in Aries, but the interaction with the platform specific UI is non-trivial. I'm happy to hear other people's thoughts on this from application perspective.

features/0509-action-menu/README.md
jcourt562: Done :-)

features/0509-action-menu/README.md
jcourt562: Done

features/0509-action-menu/README.md
jcourt562: Done

features/0509-action-menu/README.md
jcourt562: Done

features/0509-action-menu/README.md
jcourt562: Done

swcurran:
@tmarkovski - thanks for the comments.  My concern is that without this, all enterprise to mobile interactions will require a web interface, often with a QR code scan, which I think people will not like.  With this functionality, there can be relatively rich interactions between such entities. Further, since the interactions are well known to developers - essential, simplified web interactions - getting support will not be that difficult.

Not sure what you are seeing (likely more than I), but I've had a number of questions about how interactions can happen directly with the wallet, and there isn't currently a way to do that.

BTW - basic message is insufficient for much.  I would not encourage it's use. It's way too...basic.

PR #511, opened 2020-07-22
PROPOSE: RFC0511 register DIF credential-manifest as attachment format in issue-credential V2

llorllale:

### DIFF
--- a/features/0511-dif-cred-manifest-attach/README.md
+++ b/features/0511-dif-cred-manifest-attach/README.md
@@ -1,0 +1,244 @@
+# Aries RFC 0511: Credential-Manifest Attachment format for requesting and presenting credentials
+- Authors: George Aristy (SecureKey Technologies)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-07-22
+- Status Note:
+- Supersedes:
+- Start Date: 2020-07-22
+- Tags: [feature](/tags.md#feature), [protocol](/tags.md#protocol), [credentials](/tags.md#credentials), [test-anomaly](/tags.md#test-anomaly)
+
+
+## Summary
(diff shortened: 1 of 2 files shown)


features/0511-dif-cred-manifest-attach/README.md
```diff
@@ -0,0 +1,292 @@
...
+{
+    "challenge": "...",
+    "domain": "...",
+    "credential_manifest": {
+        // credential manifest object
+    }
+}
+```
+
+Complete message example:
```
swcurran, line 93: I'm thinking a better approach would be to have a pointer to examples in the DIF document vs. putting examples inline here.  I think this document should provide context between the Issue Credential protocol and the DIF data model - how one is used with the other.

features/0511-dif-cred-manifest-attach/README.md
```diff
@@ -0,0 +1,292 @@
...
+                }
+            }
+        }
+    }]
+}
+```
+
+### `issue-credential` attachment format
+
+This specification does not register any format identifier for the `issue-credential` message. The Issuer SHOULD set the `format` to the value
```
swcurran, line 261: I don't know that much about the DIF document, but I don't understand why there would not be a format string for this.

features/0511-dif-cred-manifest-attach/README.md
TelegramSam: I think an intentionally incomplete example is useful. Show some outer structure, and indicate the incompleteness with a comment.

swcurran:
Hmm...no button to update the base branch. Back to you @llorllale

llorllale:
@swcurran updated - please check now

PR #513, opened 2020-07-23
Typo correction

kukgini:
If the original intent is 'at rest', you can ignore this PR.

Signed-off-by: kukgini <kukgini@gmail.com>


TimoGlastra:
I think 'at rest' is intended in this case. It means that when DIDComm message are not used (at rest) they should be stored encrypted.

llorllale:
"at rest" is definitely intended here

PR #516, opened 2020-07-27
Update index for new RFCs

swcurran:


swcurran:
@llorllale @TelegramSam - please merge to update the index file.

PR #519, opened 2020-07-28
propose RFC 0519: Goal Codes

dhh1128:
Define the format and semantics of goal codes.


concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+But in a mature SSI ecosystem, where unknown agents can contact one another for arbitrary reasons, this context is not always easy to deduce. Acme Corp's agent may support many different protocols, and Alice may interact with Acme in the capacity of customer or potential employee or vendor. Although we have [feature discovery](../../features/0031-discover-features/README.md) to learn what's possible, and we have [machine-readable governance frameworks](../0430-machine-readable-governance-frameworks/README.md) to tell us what rules might apply in a given context, we haven't had a way to establish the context in the first place. When Alice contacts Acme, a context is needed before a governance framework is selectable, and before we know which features are desirable.
+
+The key ingredient in context is __intent__. If Alice says to Acme, "I'd like to connect,", Acme wants to be able to trigger different behavior depending on whether Alice's intent is to be a customer, apply for a job, or audit Acme's taxes. This is the purpose of a goal code.
```
swcurran, line 23: Have you given any thought to combining this concept with Action Menu to achieve the desired goal? It seems like the power of Action Menu (corresponding to the simple power that was the generation 1 HTML `form` element) would merge nicely with goal-code to enable the machine handling of Action Menu forms.

concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+>TODO: should we make a decorator out of this, so protocols don't have to declare it, and so any message can have a goal code? Or should we just let protocols declare a field in whatever message makes sense?
+
+Protocols use fields of this type as a way to express the intent of the message sender, thus coloring the larger context. In a sense, goal codes are to DIDComm what the `subject:` field is to email -- except that goal codes have formalized meanings to make them recognizable to automation.
+
+Goal codes use a standard format. They are lower-cased, kebab-punctuated strings. ASCII and English are recommended, as they are intended to be read by the software developer community, not by human beings; however, full UTF-8 is allowed. They support hierarchical dotted notation, where more general categories are to the left of a dot, and more specific categories are to the right. Some example goal codes might be:
+
+* `aries.sell.consumer.fitness`
```
dongwangdw, line 35: This makes me think of domain name, like `mail.yourname.come, hr.youname.com zone-a.hr.yourname.com, zone-b.hr.youname.com. etc` We may need this kind of standard.

0000-template-protocol.md
```diff
@@ -104,6 +104,22 @@ case-sensitively.
 State management in protocols is a deep topic. For more information, please
 see [State Details and State Machines](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/state-details.md).
 
+### Goal Codes
+
+If your protocol defines any [goal codes](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0519-goal-codes/README.md), or if it uses goal codes defined elsewhere, and you can enumerate them, define or link them here. Use markup as shown in the sample block below.
```
llorllale, line 109: nit: use relative link

concepts/0519-goal-codes/README.md
llorllale: @swcurran in that case I'd propose modifying `action-menu` such that the `menu` message can be sent in response to *any* didcomm message although that would add complexity to the state machines...

I'd rather keep having the initiating party start with the goal code, more so when both are operating under a governance framework.

concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Goals for others
+
+Related to directionality, it may occasionally be desirable to propose goals to others, rather than adovcating your own: "Let &lt;parties = us = Alice, Bob, and Carol> &lt;goal = hold an auction&gt; -- I nominate Carol to be the &lt;role = auctioneer&gt; and get us started." The difference between a normal message and an unusual one like this is not visible in the goal code; it should be exposed in additional fields that associate the goal with a particular DID+role pair. Essentially, you are proposing a goal to another party, and these extra fields clarify who should receive the proposal, and what role/perspective they might take with respect to the goal.
```
llorllale, line 54: I would generalize DID to "ID" in "DID+role pair". Not all identifiers in DIDComm are DIDs.

concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+##### `aries.vc.verify`
+Verify or validate VC-based assertions.
+##### `aries.rel`
+Create, maintain, or end something that humans would consider a relationship. This should not to be confused with building a DIDComm channel. (Building a DIDComm channel is a low-level procedure, not a high-level goal.)
+##### `aries.rel.build`
+Create a relationship. Carries the meaning implied today by a LinkedIn invitation to connect or a Facebook "Friend" request.
+
+## Unresolved questions
+
+- What parts of the design do you expect to resolve through the
```
llorllale, line 132: editorial: these need to be removed if there aren't any unresolved questions

concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+The key ingredient in context is __intent__. If Alice says to Acme, "I'd like to connect,", Acme wants to be able to trigger different behavior depending on whether Alice's intent is to be a customer, apply for a job, or audit Acme's taxes. This is the purpose of a goal code.
+
+### The goal code datatype
+
+To express intent, this RFC formally introduces the goal code datatype. When a field in a [DIDComm](../0005-didcomm/README.md) message contains a goal code, its semantics and format match the description given here. (Goal codes are often declared via the `~thread` decorator, but may also appear in ordinary message fields. See the [Scope section](#scope) below. Convention is to name this field "goal_code" where possible; however, this is only a convention, and individual protocols may adapt to it however they wish.)
+
+>TODO: should we make a decorator out of this, so protocols don't have to declare it, and so any message can have a goal code? Or should we just let protocols declare a field in whatever message makes sense?
```
llorllale, line 29: We don't currently have the concept of "inheritable *fields*" so a decorator is the best we can do that can be applied across all protocols.

And since we would add a decorator anyway, I suggest we add it to `~thread` just like you did in your threading example above.

concepts/0519-goal-codes/README.md
```diff
@@ -0,0 +1,149 @@
...
+>##### `aries.sell.services.consumer`
+>Sell services of interest to general consumers.
+>##### `aries.sell.services.enterprise`
+>Sell services of interest to enterprises.
+
+#### In DIDComm-based Protocol Specs
+
+Occasionally, goal codes may have meaning only within the context of a specific protocol. In such cases, it may be appropriate to declare the goal codes directly in a protocol spec. This can be done using a section of the RFC as described above.
+
+More commonly, however, a protocol will *accomplish* one or more goals (e.g., when the protocol is fulfilling a co-protocol interface), or will require a participant to *identify* a goal at one or more points in a protocol flow. In such cases, the goal codes are probably declared external to the protocol. If they can be enumerated, they should still be referenced (hyperlinked to their respective definitions) in the protocol RFC.
```
llorllale, line 100: > a participant to *identify* a goal at one or more points in a protocol flow

In this case the goal codes the participant responds with are merely data - we should not confuse the actual goal of this particular interaction ("myprotocol.collect.goals").

llorllale: > *accomplish* one or more goals

Suggest we make it an array.

llorllale:
@dhh1128 did we drop the description of goal codes? It would be nice to have this single, optional, human-readable field displayed to the end user, even when multiple goal codes are involved, a bit analogous with mobile app permissions today.

0000-template-protocol.md
dhh1128: The reason this is not a relative link is that the link has to be valid from its original position at the root of the repo, as well as from its new position inside the /features/X or /concepts/X folder.

concepts/0008-message-id-and-threading/README.md
```diff
@@ -61,7 +61,8 @@ Message threading will be implemented as a [decorator](../0011-decorators/README
         "thid": "98fd8d72-80f6-4419-abc2-c65ea39d0f38",
         "pthid": "1e513ad4-48c9-444e-9e7e-5b8b45c5e325",
         "sender_order": 3,
-        "received_orders": {"did:sov:abcxyz":1}
+        "received_orders": {"did:sov:abcxyz":1},
+        "goal_code": "aries.credx.issue"
```
dhh1128, line 65: This should be "aries.vc.issue", to be consistent with what's declared elsewhere.

0000-template-protocol.md
```diff
@@ -104,6 +104,22 @@ case-sensitively.
 State management in protocols is a deep topic. For more information, please
 see [State Details and State Machines](https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0003-protocols/state-details.md).
 
+### Goal Codes
```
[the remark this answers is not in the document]
dhh1128: Removed.

concepts/0519-goal-codes/README.md
dhh1128: Updated to be more general.

concepts/0519-goal-codes/README.md
dhh1128: Removed.

concepts/0008-message-id-and-threading/README.md
dhh1128: Fixed.

dhh1128:
@talltree says he votes to rename these "purpose codes" because that's the terminology used in governance frameworks. Robert Mitwicki also noted the use of the term "purpose" in the data consent work he's been doing.

swcurran:
Where does that leave OOB, where we have used "goal"?  Do we change it?  Or is it a separate concept from this?

dhh1128:
I think if we rename this to "purpose codes", we'd want to consider renaming the field in OOB. However, the RFC already has language that says each use of the data type can appear in field names of a protocol's choosing, so it wouldn't be strictly required to do the rename.

PR #532, opened 2020-08-27
Update the RFC to transition to Step 2. Does not include update ALL RFCs to change from did:sov...

swcurran:


swcurran:
This step SHOULD include changing (almost) all references from "did:sov..." to "https://didcomm.org/".  Almost, because at least this RFC has some references that need to stay the same.

swcurran:
Should all of the other RFCs be updated in this PR or a follow up?

dhh1128 (approved):
Looks fine.

dhh1128:
Vote taken on community call on 23 Sep 2020: agreed to merge.

PR #542, opened 2020-09-25
RFC 0193: Change coin flip algorithm

pstuermlinger:
To follow PR #193, which proposes the Coin Flip Protocol, this changes the used algorithm to save one round of communication.

Basically, the Recorder provides the Caller two hashed options. One is the hash of `win<salt>` the other one of `lose<salt>`. The caller has no idea which one belongs to what. After the Caller made a blind decision, the Recorder reveals the salt and both parties can verify the result.

This is easily extendable to provide more than the given two alternatives.

I'm not sure if `Coin Flip` is still a good name, since we actually don't flip anything. It's just a blind choice between n alternatives. On the other hand, it still has the same intention and effect of a coin flip.


dhh1128 (approved):
This is an awesome enhancement. Love it!

dhh1128:
Since this is a simple enhancement to a not-yet-implemented protocol at the PROPOSED state, I believe it can be merged immediately.

pstuermlinger:
This was quick - thank you!

PR #548, opened 2020-10-05
Fix inconsistencies in mediator-coordination types

dbluhm:
Quick fixes in protocol name and message types to make this protocol more consistent with itself (consistent use of kebab-case vs. snake_case).


llorllale (dismissed):
If the community really cares about this then we need to have a discussion as it involves changes that will break interop. We must first agree to a style guide and submit an RFC for that. I see other RFCs using camelCase and snake_case (eg. RFC0056 and 0048).

dbluhm:
Given that this is still in proposed state, I personally didn't see a problem with making these corrections. If there are enough implementations of this in the wild that it disrupts interop, I agree that that should be taken into consideration. I admit, I didn't think that likely given the proposed status and did not account for it in my proposed changes. I do, however, think these changes are important for the quality of the protocol and recommend we do take the necessary steps to incorporate these corrections, whether that be with a version bump and associated coordination or not.

I think a style guide RFC is a good idea but disagree with the idea that this PR must sit open until a style guide RFC is proposed and accepted. These changes do not need to be tied to repo wide changes and if a style guide RFC is proposed and changes are needed for other protocols, the updates will have to occur one by one with coordination and version incrementing so I don't see a need to treat this differently.

dbluhm:
I'll add that these changes were more seeking to be consistent with the protocol itself than with other protocols and I don't consider these changes to be a statement about how other protocols should do things besides being internally consistent.

swcurran:
I don't see this as a global issue, but specific to this RFc and am happy to accept the changes.  If we want to make a larger change across the set of RFCs, that would need a lot more discussion, documentation and planning.

llorllale:
@swcurran @dbluhm I forgot this was in PROPOSED state.

Not to nitpick too much, but one of this PR's stated goals is to "make this protocol more consistent with other protocols", which leaves a door open for other arbitrary changes elsewhere and I urge us to be cautious there. We should agree to a style guide and stick with that.

swcurran:
I don't see any reason to hold up this PR for that.  It is changing nothingcase to kebob-case and improving the names of some variables on a protocol that no one has implemented.

Maybe change the PR's goal to "Improve the usability of this protocol", drop the line about "consistent with other protocols" and then we can be happy?

swcurran (approved):
LGTM

kdenhartog (approved):

llorllale:
@swcurran 

> I don't see any reason to hold up this PR for that.

To be clear: 21 hours ago I dismissed my earlier review requesting changes, removing any blockers on this PR.

What I am resisting against is opening the door to arbitrary changes done willy-nilly that do not necessarily improve the protocol in any substantial way. And yes - I agree that is a separate discussion that should be proposed through an RFC.

swcurran:
Sorry @llorllale -- I was confused by what you meant by the action and comment.  All good.

Issue #534, opened 2020-09-09
RFC0023: clarify signature in 'did_doc~attach'


llorllale:
The RFC merely states:

> The `did_doc~attach` contains the DID Doc associated with the DID as a signed attachment. If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc~attach` attribute is optional.

We should clarify what exactly is being signed. In particular, we would like to make this compatible with existing KMS interfaces where data is signed/verified by the remote KMS. Which would mean this signature should be done over a hash of the did doc's contents as opposed to the raw bytes.

So we should have a top-level `did_doc` attribute with the raw did doc bytes, and the attachment would have a hash of this doc.

sklump:
Just a line here to note that the text states that
_The `did_doc~attach` contains the DID Doc associated with the DID as a signed attachment_
with example
```
{
...
  "did_doc~attach": {
     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
     "jws": {
        "header": {
           "kid": "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
        },
        "protected": "eyJhbGciOiJFZERTQSIsImlhdCI6MTU4Mzg4... (bytes omitted)",
        "signature": "3dZWsuru7QAVFUCtTd0s7uc1peYEijx4eyt5... (bytes omitted)"
      }
   }
}
```

but this is not a signed attachment: it is the signed `"data"` component of a signed attachment; ref. RFC 17 example `report~attach` below:
```
{
  "@type": "https://didcomm.org/xhomebuy/1.0/home_insp",
  "inspection_date": "2020-03-25",
  "inspection_address": "123 Villa de Las Fuentes, Toledo, Spain",
  "comment": "Here's that report you asked for.",
  "report~attach": {
    "mime-type": "application/pdf",
    "filename": "Garcia-inspection-March-25.pdf",
    "data": {
      "base64": "eyJ0eXAiOiJKV1QiLA0KICJhbGciOiJIUzI1NiJ... (bytes omitted to shorten)",
      "jws": {
        // payload: ...,  <-- omitted: refer to base64 content when validating
        "header": {
          "kid": "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
        },
[...]

PR #552, opened 2020-10-08
Tweaks to DID Exchange - inviter/invitee references; Attach fields; headings lint cleanup

swcurran:
Some further cleanup on the DID Exchange RFC.

- Some references to the "inviter" and "invitee" were left from my last clean up `#fail`
- Adjusted the `attach~did-doc` field to match the RFC 0017 Attachments format
- Changed some titles to eliminate some markdown lint complaints because such issues bug me

I think a quick review and merge is likely OK with this one.

This does NOT address #534 but the example attachment change was a result of a comment in that issue.

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -143,7 +143,7 @@ The _requester_ may provision a new DID according to the DID method spec. For a 
   "label": "Bob",
   "did": "B.did@B:A",
   "did_doc~attach": {
-      "@id": "B.did@B:A",
+      "@id": "d2ab6f2b-5646-4de3-8c02-762f553ab804",
       "mime-type": "application/json",
       "data": {
          "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
@@ -193,7 +193,7 @@ When a `request` responds to an implicit invitation, its `~thread.pthid` MUST co
   "label": "Bob",
   "did": "B.did@B:A",
   "did_doc~attach": {
-      "@id": "B.did@B:A",
+      "@id": "d2ab6f2b-5646-4de3-8c02-762f553ab804",
       "mime-type": "application/json",
       "data": {
          "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
@@ -222,7 +222,7 @@ When a `request` responds to an implicit invitation, its `~thread.pthid` MUST co
   "label": "Bob",
   "did": "B.did@B:A",
   "did_doc~attach": {
-      "@id": "B.did@B:A",
+      "@id": "d2ab6f2b-5646-4de3-8c02-762f553ab804",
       "mime-type": "application/json",
       "data": {
          "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
@@ -294,7 +294,7 @@ The exchange response message is used to complete the exchange. This message is 
   },
   "did": "B.did@B:A",
   "did_doc~attach": {
-      "@id": "B.did@B:A",
+      "@id": "d2ab6f2b-5646-4de3-8c02-762f553ab804",
       "mime-type": "application/json",
       "data": {
          "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -143,13 +143,17 @@ The _requester_ may provision a new DID according to the DID method spec. For a 
   "label": "Bob",
   "did": "B.did@B:A",
   "did_doc~attach": {
-     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
-     "jws": {
-        "header": {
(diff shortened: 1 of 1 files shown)


swcurran:
@dhh1128 @TelegramSam @kdenhartog @llorllale -- please review and if ok, merge.

kdenhartog (approved):
Looks good to me. I'm +1 to moving from inviter/invitee to requestor/responder

llorllale (approved):
LGTM although we need to fix the attachments in the other examples as well

features/0023-did-exchange/README.md
```diff
@@ -282,25 +282,29 @@ The exchange response message is used to complete the exchange. This message is
...
-     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
-     "jws": {
-        "header": {
-           "kid": "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
-        },
-        "protected": "eyJhbGciOiJFZERTQSIsImlhdCI6MTU4Mzg4... (bytes omitted)",
-        "signature": "3dZWsuru7QAVFUCtTd0s7uc1peYEijx4eyt5... (bytes omitted)"
+      "@id": "B.did@B:A",
+      "mime-type": "application/json",
+      "data": {
```
sklump, line 287: Note that there are three more examples in the text that need this treatment (add `"@id"`, `"mime-type"` and subsume the rest into value for `"data"` key)

sklump (changes requested):
Look closely for three more did_doc~attach examples that need to align with RFC 17

swcurran:
Doh!  Thanks George. I'll get to it later today.

features/0023-did-exchange/README.md
```diff
@@ -330,15 +334,15 @@ Possible reasons:
...
+##### Response Processing Error
 
 - unknown processing error
 
 ## 3. Exchange Complete
 
 The exchange `complete` message is used to confirm the exchange to the _responder_. This message is **required** in the flow, as it marks the exchange complete. The _responder_ may then invoke any protocols desired based on the context expressed via the `pthid` in the DID Exchange protocol.
 
-#### Example
+### Complete Message Example
```
sklump, line 345: Could this section add a treatment of errors and corresponding problem report error codes for exhange-complete message processing errors; e.g., no such exchange at responder? In the weird case where the responder gets an completion with bad data, there is no direction here so far on how to proceed.

features/0023-did-exchange/README.md
```diff
@@ -313,13 +317,13 @@ When the message is sent, the _responder_ are now in the `response-sent` state.
 
 #### Response Processing
 
-When the _invitee_ receives the `response` message, they will verify the `change_sig` provided. After validation, they will update their wallet with the new information. If the endpoint was changed, they may wish to execute a Trust Ping to verify that new endpoint.
```
sklump, line 316: This is the first and only mention of `change_sig`: what is it?

features/0023-did-exchange/README.md
swcurran: Note: Not yet added.

features/0023-did-exchange/README.md
swcurran: I removed the reference and attempted to clarify above. Basically, if the responder had initiated the connection via an OOB invitation or a public DID and then switched to a peer DID, the DIDDoc signature should use the DID/key from the OOB or public DID.

features/0023-did-exchange/README.md
swcurran: Added the same error sections to Complete as with the other message type. For now, no reasons added, and only an "unknown processing error".

features/0023-did-exchange/README.md
```diff
@@ -143,19 +143,23 @@ The _requester_ may provision a new DID according to the DID method spec. For a
...
   "did": "B.did@B:A",
   "did_doc~attach": {
-     "base64": "eyJ0eXAiOiJKV1Qi... (bytes omitted)",
-     "jws": {
-        "header": {
-           "kid": "did:key:z6MkmjY8GnV5i9YTDtPETC2uUAW6ejw3nk5mXF5yci5ab7th"
-        },
-        "protected": "eyJhbGciOiJFZERTQSIsImlhdCI6MTU4Mzg4... (bytes omitted)",
-        "signature": "3dZWsuru7QAVFUCtTd0s7uc1peYEijx4eyt5... (bytes omitted)"
+      "@id": "B.did@B:A",
```
llorllale, line 146: Although this example is non-normative, I don't think it's a good idea to add `@id` and set its value to the DID's ID. That will probably mislead developers.

swcurran: I was certain on what to put there.  What do you suggest?

llorllale: Any UUID would suffice.

swcurran: Done

PR #557, opened 2020-10-16
New RFC: Feature Discovery 2.0 that supports goal codes, crypto, decorators, etc.

dhh1128:


features/0031-discover-features/README.md
```diff
@@ -1,244 +1,150 @@
...
+
+Implementations of this protocol must recognize the following values for `feature-type`: `protocol`, `goal-code`, `crypto`, and `decorator`/`header`. (The concept known as `decorator` in DIDComm v1 approximately maps to the concept known as `header` in DIDComm v2. The two values should be considered synonyms and must both be recognized.) Additional values of `feature-type` may be standardized by raising a PR against this RFC that defines the new type and increments the minor protocol version number; non-standardized values are also valid, but there is no guarantee that their semantics will be recognized.
 
-The `query` field may use the * wildcard. By itself, a query with just
-the wildcard says, "I'm interested in anything you want to share with
-me." But usually, this wildcard will be to match a prefix that's a little
-more specific, as in the example that matches any 1.x version.
+Identifiers for feature types vary. For protocols, identifiers are PIURIs. For goal codes, identifiers are goal code values. For crypto, identifiers are standardized suite or key type names from W3C standards.
```
llorllale, line 65: > For crypto, identifiers are standardized suite or key type names from W3C standards.

I don't quite understand this one.

At what level of the didcomm message does this crypto apply to? Would love to see an example of a query and a response.

Crypto suites defined by the the relevant W3C CCGs are scoped to the object's format eg. `Ed25519Signature2018`.

dhh1128: We're not discovering crypto at the message level or at the JWT level. We're discovering what crypto an agent supports. Does that help, or am I still being ambiguous?

llorllale: @dhh1128 

> We're discovering what crypto an agent supports. Does that help, or am I still being ambiguous?

I'm afraid so. :(

Can we start with an example?

dhh1128: A crypto-related query might be: "Do you support Ed25519Signature*". A response might be "Yes, I support Ed25519Signature2018". Similarly, we might ask an agent if it supports the goldilocks curve, or secp256k1, or RSA, etc. 

Your observation that values like "Ed25519Signature2018" is scoped to the object's format, and your observation that JWT-level crypto is registered with IANA, are true but orthogonal. It's true that the question of what's being *used* in a different scope is expressed in a different scope, and it's true that registration happens elsewhere. But what we're asking and answering here is different. We want to know whether an agent supports that crypto mechanism in the abstract, not whether the mechanism is active in a particular scope. If a given artifact contains an Ed25519Signature2018, we can infer that the agent that built the artifact supports it, and knows where such a mechanism is registered -- but we can't infer anything about mechanisms that are not in evidence already in a particular scope; we have to ask the agent about them.

llorllale: > We want to know whether an agent supports that crypto mechanism in the abstract

That's not really that useful though is it?

Here's an example of a more interesting question: do you support issuing zcaps with the Ed25519Signature2018 suite? The fact they support Ed25519Signature2018 for zcaps does not automatically mean they also issue other forms of credentials with the same suite.. I'm sure you and I agree there...

dhh1128: The "crypto" query is about the abstract question. Shall we add a line or two about how to query for particular zcap mechanisms? Maybe we say that you query by type = zcap; is the value you use sufficiently described by signature type, or do you have to combine signature with format or some other descriptor?

TelegramSam:
in WG 20201104, we discussed this needing to be a new RFC given it's a major version bump and not a minor version bump. Other than the structure, the new version looks good.

dhh1128:
Updated per comment from @TelegramSam above. New v2 protocol is now RFC 0557; old RFC 0031 has note about being superseded, but remains unmodified.

features/0031-discover-features/README.md
TelegramSam: I think the complexities of crypto make discovering crypto support a little more difficult than the rest. For example, if I support crypto suite 'x' for signatures, do I support signing them or verifying them only?
Is there a way to move forward with this RFC and continue the crypto related discussions?

TelegramSam:
I wonder if wildcards need more clarification: they are NOT regular expressions. Do we allow wildcards in the middle of a string, or only at the end?

features/0031-discover-features/README.md
llorllale: > Is there a way to move forward with this RFC and continue the crypto related discussions?

+1 continue continue crypto discussion in another issue. I recommend it be excluded from this PR.

kdenhartog: +1 to scoping crypto in a separate issue to be added later. The crypto is going to be a fairly complex one to resolve and shouldn't block the other great parts of this.

dhh1128:
I have now removed all references to crypto. We can deal with that in another PR.

swcurran (approved):

PR #558, opened 2020-10-16
Clarify detached JWS rules per comments in #556

dhh1128:
See [this comment](https://github.com/hyperledger/aries-rfcs/issues/556#issuecomment-710676317) on issue #556.

Signed-off-by: Daniel Hardman <daniel.hardman@gmail.com>

PR #561, opened 2020-11-05
substantive, controversial change: total base64url consistency in repo

dhh1128:
Recent work on RFC 0017 (see PR #558) made us realize that we have been casual in our references to "base64" -- sometimes meaning true base64, and sometimes meaning base64url (see https://tools.ietf.org/html/rfc4648#section-5). After community discussion, we've concluded that base64url encoding is the preferred target, and that we'd like to be consistent. I updated PR #558 for consistency within the attachments RFC, but I did not update other RFCs because changes in those other places are more controversial. For example, in Aries RFCs 0036 and 0037, we say that Indy-style attachments are base64-encoded by libindy. If libindy is returning base64 instead of base64url, then simply updating the RFCs creates a delta between the RFC and deployed code that we need to manage thoughtfully.

This PR embodies a sweep through all the other RFCs in the repo, updating base64 to base64url everywhere that seemed appropriate. However, each RFC that I touched may have different outcomes with respect to these changes. Therefore, the hope of this PR is to present all possible updates so the community can examine each one and authors of respective RFCs can be tagged to see how they feel about updates to the content they authored.

This PR is framed as an addition to #558, not as an independent update. Therefore, it should be rebased after 558 is merged, and possibly split into separate commits.


concepts/0017-attachments/README.md
```diff
@@ -346,10 +346,7 @@ signing key when the sender is performing key rotation.
 
 Embedded and appended attachments support signatures by the addition of a `data.jws` field
 containing a signature in [JWS (RFC 7515) format](https://tools.ietf.org/html/rfc7515)
-with [Detached Content](https://tools.ietf.org/html/rfc7515#appendix-F).
-The payload of the JWS is the raw data of the attachment, whether externally referenced
-or encoded in base64 format, and is not contained within the signature itself.
-Signatures over inlined JSON attachments are not currently defined as this
+with [Detached Content](https://tools.ietf.org/html/rfc7515#appendix-F). The payload of the JWS is the raw bytes of the attachment, appropriately base64url-encoded per JWS rules. If these raw bytes are incorporated by value in the DIDComm message, they are already base64url-encoded in `data.base64` and are thus directly substitutable for the suppressed `data.jws.payload` field; if they are externally referenced, then the bytes must be fetched via the URI in `data.links` and base64url-encoded before the JWS can be fully reconstituted. Signatures over inlined JSON attachments are not currently defined as this
```
kdenhartog, line 349: This does a good job of highlighting what I was trying to point out in issue #534 as well

kdenhartog (approved):
I'm a +1 to this. It's going to make it much clearer that we're always using the same encoding scheme. We should probably double check that the changes in a few of these (thinking about 0036 and 0037 particularly) won't break AIP 1.0 implementations as well.

Baha-sk:
+1 for this PR

I would add raw (no padding) to the base64url encoding as well.

Baha-sk (approved):
LGTM

adding `raw (no padding)` will clarify that base64url encoding doesn't pad the binary data to be octet size exactly. This clarification ensures implementations interop.

swcurran:
We've have this wording in several RFCs and it has not changed with this PR.  I assume we are good at keeping it in:

> NOTE: In receiving the out-of-band message, the base64url decode implementation used MUST correctly decode padded and unpadded base64URL encoded data.

It's there because it has bitten us in the past.

PR #562, opened 2020-11-05
Update formulation, formatting in did-exchange RFC

Patrik-Stas:
3 points I would like to address here

- Changed order of the words a bit
- When there's multiple options (especially "wordy" choices such as in edited example), I would find using list of points more readable than "plain sentence". This would be applicable to more places in this (and probably other documents as well) though.
- An open question, I don't understand this "or the `service` element of the `handshake_protocols` attribute of the `invitation` message is used to". I looked at out of band invitation example https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md but it's just a list, so I don't understand what's meant by "service` element of the `handshake_protocols`". Can someone clarify?

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -122,7 +122,13 @@ No errors are sent in timeout situations. If the _requester_ or _responder_ wish
 
 ## Implicit and Explicit Invitations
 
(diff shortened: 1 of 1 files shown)


swcurran (approved):
LGTM

swcurran:
Dang!  @Patrik-Stas I should have read your question more carefully first before merging.  The reference to "handshake_protocols" should be removed.  It should just be the service element of the OOB `invitation` message.

PR #566, opened 2020-11-16
New Concept RFC: Issuer Hosted Custodial Agents

TelegramSam:

### DIFF
--- a/concepts/0566-issuer-hosted-custodidal-agents/README.md
+++ b/concepts/0566-issuer-hosted-custodidal-agents/README.md
@@ -66,7 +66,7 @@ The following best practices should be followed to ensure proper operation and c
 Most of these practices depend upon and support one another.
 
 ### Defend the SSI architecture
-When issuers host custodial agents, care must be taken to avoid shortcuts that would violate SSI architecture. Deviations will frequently frequently lead to incompatibilities.
+When issuers host custodial agents, care must be taken to avoid shortcuts that would violate SSI architecture. Deviations will frequently lead to incompatibilities.
 
 ### DIDComm Protocol based Integration
 Communication between hosted agents and credential issuing agent must be based on published DIDComm protocols. Any communication which eliminates the use of a DID must be avoided. Whenever possible, these should be well adopted community protocols. If the case a new protocol is needed for a particular interaction, this must be fully documented and published, to allow other agents to become compatible by adopting the new protocol.
--- a/concepts/0566-issuer-hosted-custodidal-agents/README.md
+++ b/concepts/0566-issuer-hosted-custodidal-agents/README.md
@@ -1,0 +1,98 @@
+# 0566: Issuer-Hosted Custodial Agents
+- Authors: [Sam Curren](telegramsam@gmail.com)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-11-16
+- Status Note: Draft
+- Start Date: 2020-10-27
+- Tags: concept
+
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):
Ahh no... my eyes are burning... this isn't SSI! 

Just kidding 😄 This is a good write up of the practical tradeoffs necessary to get adoption from custodial services hosting agents on behalf of a user. I'm in favor of merging this PR.

swcurran (approved):
No reason to not approve this, but I don't get the "issuer hosted" angle. Does that mean if I have more than one credential, I have more than one wallet, one with each issuer?  It seems like an odd distribution.

I'm also wondering if this is intended to be a basis for an organizational wallet (e.g. a holder wallet for an organization) or would that be a separate concept. I guessing separate, but wondered if that was being considered.

concepts/0566-issuer-hosted-custodidal-agents/README.md
```diff
@@ -0,0 +1,108 @@
...
+### Smartphone App
+While it is common for a smartphone app to be an agent by itself, there are cases where a smartphone app can act as a remote for the hosted agent. In this iteraction, keys, credentials, and other wallet related data is held in the custodial agent. The mobile app acts as a remote viewer and a way for the user to authorize actions taken by the custodial agent.
+
+## Best Practices
+
+The following best practices should be followed to ensure proper operation and continued transition to a fully realized SSI architecture.
+Most of these practices depend upon and support one another.
+
+### Defend the SSI architecture
+When issuers host custodial agents, care must be taken to avoid shortcuts that would violate SSI architecture. Deviations will frequently frequently lead to incompatibilities.
```
TimoGlastra, line 69:
```suggestion
When issuers host custodial agents, care must be taken to avoid shortcuts that would violate SSI architecture. Deviations will frequently lead to incompatibilities.
```

concepts/0566-issuer-hosted-custodidal-agents/README.md
```diff
@@ -0,0 +1,108 @@
+# 0566: Issuer-Hosted Custodial Agents
+- Authors: [Sam Curren](telegramsam@gmail.com)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-11-16
+- Status Note: Draft
+- Start Date: 2020-10-27
+- Tags: concept
+
+In the fully realized world of Self Soverign Identity, credential holders are equipped with capable agents to helpa them manage credentials and other SSI interactions. Before we arrive in that world, systems that facilitate the transition from the old model of centralized systems to the new decentralized models will be necessary and useful.
```
TimoGlastra, line 9:
```suggestion
In the fully realized world of Self Soverign Identity, credential holders are equipped with capable agents to help them manage credentials and other SSI interactions. Before we arrive in that world, systems that facilitate the transition from the old model of centralized systems to the new decentralized models will be necessary and useful.
```

TelegramSam:
> No reason to not approve this, but I don't get the "issuer hosted" angle. Does that mean if I have more than one credential, I have more than one wallet, one with each issuer? It seems like an odd distribution.

It will result in the situation you describe. Users who have an agent hosted for them will almost always not realize what has happened. At first glance, their hosted agent will feel similar to the web interface that they use on the centralized system. While not ideal from a 'pure' SSI option, it offers better options going forward and is no worse than their current system.

> I'm also wondering if this is intended to be a basis for an organizational wallet (e.g. a holder wallet for an organization) or would that be a separate concept. I guessing separate, but wondered if that was being considered.

I think separate, but I'm not sure I really understand what you are asking here. If you mean that the issuer could host a wallet for an org (as opposed to a person) I suppose it would apply.

PR #568, opened 2020-12-03
Update README.md

devin-fisher:
This should not change the semantic meaning of this RFC but I think this provides a clear explanation.

### DIFF
--- a/concepts/0008-message-id-and-threading/README.md
+++ b/concepts/0008-message-id-and-threading/README.md
@@ -89,10 +89,10 @@ A thread object has the following fields discussed below:
 Because multiple interactions can happen simultaneously, it's important to
 differentiate between them. This is done with a Thread ID or `thid`.
 
-The Thread ID is the Message ID (`@id`) of the first message in the thread. The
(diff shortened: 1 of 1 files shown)


concepts/0008-message-id-and-threading/README.md
```diff
@@ -89,10 +89,10 @@ A thread object has the following fields discussed below:
 Because multiple interactions can happen simultaneously, it's important to
 differentiate between them. This is done with a Thread ID or `thid`.
 
-The Thread ID is the Message ID (`@id`) of the first message in the thread. The
-first message may or may not declare the `~thread` attribute block; it
-does not, but carries an
-implicit `thid` of its own `@id`.
+If the Thread object is defined and a `thid` is given, the Thread ID is the value
+given there. But if the Thread object is not defined in a message, the Thread ID should
```
swcurran, line 93: A good change, but I suggest removing the "should" ambiguity.  `...the Thread ID is the message ID...`

devin-fisher: done

swcurran (approved):

PR #569, opened 2020-12-03
441 present proof best practices

sklump:
Best practices for provers in credential selection and verifiers in proof regarding non-revocation intervals and timestamps.


concepts/0441-present-proof-best-practices/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Presentation Requests and Non-Revocation Intervals
+
+This section prescribes norms and best practices in formulating and interpreting non-revocation intervals on proof requests.
+
+#### Semantics of Non-Revocation Interval Presence and Absence
+
+The presence of a non-revocation interval applicable to a requested item (see [below](#non-revocation-interval-applicability-to-requested-items)) in a presentation request signifies that the verifier requires proof of non-revocation status of the credential providing that item.
+
+The absence of any non-revocation interval applicable to a requested item signifies that the verifier has no interest in its credential's non-revocation status.
```
swcurran, line 29: I suggest adding a clarification along the lines of "A proof may be satisfied by a revocable or non-revocable credential regardless of whether or not a revocation interval information is included in the proof request.  The presence of a revocation interval in a proof request conveys that if the prover is going to use a revocable credential in the proof, a proof of non-revocation must be included. It does not convey any sort of restriction on the revocability of the credential to be proven. In many cases the verifier can not know if the credential to be used by the prover is revocable or not."

concepts/0441-present-proof-best-practices/README.md
```diff
@@ -0,0 +1,163 @@
...
+        "to": 1600000000
+    }
+}
+```
+
+the non-revocation interval on 1600000000 is **generally applicable** to the referent `"legalname"` while the non-revocation interval on 1600001000 **specifically applicable** to referent `"regdate"`.
+
+#### Semantics of Non-Revocation Interval Endpoints
+
+A non-revocation interval contains `"from"` and `"to"` (integer) EPOCH times. For historical reasons, any timestamp within this interval is technically acceptable in a non-revocation subproof. However, these semantics allow for ambiguity in cases where revocation occurs within the interval, and in cases where the ledger supports reinstatement. These best practices obviate this equivocation, fostering deterministic outcomes.
```
swcurran, line 81: "obviate" is probably not the best word to use here -- especially considering many non-English speakers will be reading this. Perhaps use 'These best require the `"from"` and `"to"` values be the same, eliminating non-deterministic outcomes.'

sklump: ... but they don't require that they be the same. I am proposing the omission of "from" as the first choice. Point taken in any case.

swcurran (approved):
Looks good to me

PR #571, opened 2020-12-05
update MIME types RFC for inclusion in AIP

dhh1128:
As discussed on a community call, this PR tweaks RFC 0044 to make it clear how to detect DIDComm v1 vs. v2. It also explains what it means to "implement" the RFC (which should be very easy). The goal is to allow this RFC to be included in an AIP to lay the foundation for a transition to DIDComm v2 in the future.


swcurran:
@dhh1128 -- thanks Daniel.  Does this require a community update, or can frameworks proceed with this unilaterally?

dhh1128:
@swcurran : I think frameworks can proceed unilaterally. The only circumstance where that wouldn't be true is if code has been using a different MIME type, consistently -- and if other agents have come to rely on that. I don't think such a case exists, but let's ask at the next Aries WG calls.

TelegramSam (approved):

llorllale (approved):

PR #572, opened 2020-12-09
[Out-of-band] Adjust URL shortening guidelines; fix up JSON formatting and syntax

andrewwhitehead:
This adjusts the guidelines for URL shortening such that the responder must return the invitation in JSON format - not the encoded base64 content. There is no benefit to returning the response in base64 format, and this makes discoverability of the format much more difficult when a human resolves the shortened URL. Specifying the MIME types involved would help to support alternative formats in future (such as DIDComm message or CBOR for instance).

@swcurran @TelegramSam @llorllale @tmarkovski

### DIFF
--- a/features/0434-outofband/README.md
+++ b/features/0434-outofband/README.md
@@ -43,7 +43,7 @@ We'd like to have the specification of QR handling harmonized into a single RFC 
 
 ### Key Concepts
 
-The Out-of-band protocol are used when an agent doesn't know if it has a connection with another agent. This could be because you are trying to establish a new connection with that agent, you have connections but don't know who the other party is, or if you want to have a connection-less interaction. Since there is no DIDComm connection to use for the messages of this protocol, the messages are plaintext and sent out-of-band, such as via a QR code, in an email message or any other available channel. Since the delivery of out-of-band messages will often be via QR codes, this RFC also covers the use of QR codes.
+The Out-of-band protocol is used when an agent doesn't know if it has a connection with another agent. This could be because you are trying to establish a new connection with that agent, you have connections but don't know who the other party is, or if you want to have a connection-less interaction. Since there is no DIDComm connection to use for the messages of this protocol, the messages are plaintext and sent out-of-band, such as via a QR code, in an email message or any other available channel. Since the delivery of out-of-band messages will often be via QR codes, this RFC also covers the use of QR codes.
 
 Two well known use cases for using an out-of-band protocol are:
 
--- a/features/0434-outofband/README.md
+++ b/features/0434-outofband/README.md
@@ -33,7 +33,7 @@ In the existing Connections and DID Exchange `invitation` handling, the _inviter
 
 ### Handling of all Out-of-Band Messages
 
-We currently have two sets of out-of-band messages that cannot be delivered via DIDComm because there is no channel. We'd like to align those messages into a single "out of band" protocol so that their handling can be harmonized inside an agent, and a common QR code handling mechanism can be used.
+We currently have two sets of out-of-band messages that cannot be delivered via DIDComm because there is no channel. We'd like to align those messages into a single "out-of-band" protocol so that their handling can be harmonized inside an agent, and a common QR code handling mechanism can be used.
 
 ### URLs and QR Code Handling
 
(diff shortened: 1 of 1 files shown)


features/0434-outofband/README.md
```diff
@@ -43,7 +43,7 @@ We'd like to have the specification of QR handling harmonized into a single RFC
...
+The Out-of-band protocol are used when an agent doesn't know if it has a connection with another agent. This could be because you are trying to establish a new connection with that agent, you have connections but don't know who the other party is, or if you want to have a connection-less interaction. Since there is no DIDComm connection to use for the messages of this protocol, the messages are plaintext and sent out-of-band, such as via a QR code, in an email message or any other available channel. Since the delivery of out-of-band messages will often be via QR codes, this RFC also covers the use of QR codes.
```
swcurran, line 46: Please fix the first "are" in that sentence. Should be: `The Out-of-band protocol is used when an agent...`

features/0434-outofband/README.md
```diff
@@ -427,11 +427,11 @@ Knowledge is Good
...
 
 - The sender should generate and track a GUID for the out-of-band message URL.
-- The shortened version should be:
+- The shortened version should be of the form:
   - `https://example.com/ssi?id=5f0e3ffb-3f92-4648-9868-0d6f8889e6f3`
   - Note the replacement of the query parameter `oob` with `id` when using shortened URL.
-- On receipt of this form of message, the agent must do an HTTP GET to retrieve the associated, encoded out-of-band message.
-  - A sender may want to wait to generate the full invitation until the redirection event of the shortened URL to the full length form dynamic, so a single QR code can be used for distinct out of band messages.
+- On receipt of this form of message, the agent must do an HTTP GET to retrieve the associated out-of-band message. The agent should include an `Accept` header requesting the `application/json` MIME type, and the sender must include a `Content-Type` header specifying `application/json; charset=utf-8`. The sender must return the invitation in JSON format.
```
m00sey, line 433: as `must` is a keyword for rfcs can we make it all caps?

I really like the update.

features/0434-outofband/README.md
```diff
@@ -426,12 +426,12 @@ Knowledge is Good
...
 
-- The sender should generate and track a GUID for the out-of-band message URL.
-- The shortened version should be:
+- The sender generates and tracks a GUID for each out-of-band message URL.
+- The shortened URL **MUST** include an `id` query parameter, for example:
   - `https://example.com/ssi?id=5f0e3ffb-3f92-4648-9868-0d6f8889e6f3`
   - Note the replacement of the query parameter `oob` with `id` when using shortened URL.
-- On receipt of this form of message, the agent must do an HTTP GET to retrieve the associated, encoded out-of-band message.
-  - A sender may want to wait to generate the full invitation until the redirection event of the shortened URL to the full length form dynamic, so a single QR code can be used for distinct out of band messages.
+- On receipt of this form of message, the agent **MUST** perform an HTTP GET to retrieve the associated out-of-band message. The agent **SHOULD** include an `Accept` header requesting the `application/json` MIME type, and the sender **MUST** include a `Content-Type` header specifying `application/json; charset=utf-8`. The sender **MUST** return the invitation in JSON format.
```
llorllale, line 433: I believe an OOB invitation - with its `@id` and `@type` - is a JSON-LD document. If I'm reading RFC 0047 correctly, DIDComm protocols comprise of messages in JSON-LD format.

Therefore I think the `Accept` header should be set to `application/ld+json`.

Invoking @dhh1128 for cross-checking.

dhh1128: The canonical narrative about how to lightly integrate JSON-LD with plain JSON is: "just add a `@context` and then interpret the JSON as plain JSON unless/until you need more." I think folks like Manu and Dave would say that's a fair high-level summary. Under this model, every JSON doc *is* a JSON-LD doc; you just don't need to know that if all you care about is JSON.

DIDComm's approach diverges slightly from this, though its intention is similar. We wanted to allow for protocols to use JSON-LD as an extension mechanism, but instead of claiming that every message actually *is* a JSON-LD doc, we claim that every message *could be construed as* a JSON-LD doc. The difference is the presence or (in DIDComm's case) the absence of a `@context`.

Section 6 of the JSON-LD spec says that `@context` can be communicated by a backchannel. When you do that, it's not technically correct to say that the data you're looking at "is" JSON-LD; rather, when you combine the data you're looking at with the context you got from a backchannel, the combined result "can be viewed as" JSON-LD.

RFC 0047 says that the `@` character is reserved for JSON-LD'isms, and that `@id` and `@type` are JSON-LD at work. What that means is that we've deliberately named some DIDComm fields such that if you view messages through a JSON-LD lens, they have exactly the right semantics. By reserving things in this way, we guarantee that there can never be any conflicts between the JSON and JSON-LD worldviews. We then go on to say in the RFC that whether JSON-LD'isms are actually intended to invoke JSON-LD functionality is dependent on a protocol; the decision to use JSON-LD as an extension mechanism is made on a message family (protocol) basis. When a protocol says it wants JSON-LD to be its lens, it's supposed to actually define a `@context` that defines its fields properly, and (presumably; this is ambiguous in the RFC) require that `@context` be present in the message fields. Until then, the JSON-LD interpretation of the data is just potential, not actual.

I don't think the Out of Band protocol has claimed that it wants to have its messages (invitation or otherwise) viewed through a JSON-LD lens, and I don't think it has defined any `@context` or added `@context` as a required field to its messages. Therefore, I think it would be incorrect to interpret OOB messages as JSON-LD.

Per [RFC 0044](https://github.com/hyperledger/aries-rfcs/tree/master/features/0044-didcomm-file-and-mime-types#detecting-didcomm-versions), the correct MIME type for a plaintext DIDComm v1 message (of which OOB invitations are an example) is `application/json`, so I think the wording proposed in the PR is correct.

llorllale: Alright - I'll take it.

Although it worries me that it takes so many words to explain this - and I'm still not 100% sure I followed through every word.

My takeaway was: we have set things up to enable transitions JSON -> JSONLD; however all messages are JSON by default unless the protocol/RFC specifies otherwise.

llorllale (approved):

PR #576, opened 2020-12-14
RFC0434: add jwe_suite

llorllale:

### DIFF
--- a/features/0434-outofband/README.md
+++ b/features/0434-outofband/README.md
@@ -31,6 +31,13 @@ We need the ability to reuse a connection.
 
 In the existing Connections and DID Exchange `invitation` handling, the _inviter_ dictates what connection establishment protocol all _invitee_'s will use. A more sustainable approach is for the _inviter_ to offer the _invitee_ a list of supported protocols and allow the _invitee_ to use one that it supports.
 
+### JWE Envelope Parameters
+
+With two JWE envelope formats at the time of this writing ([RFC0019](../0019-encryption-envelope/README.md) and
+[RFC0334](../0334-jwe-envelope/README.md)), a mechanism that allows agents to transition from one to the other has become
+necessary. Having the sender of an invitation include the JWE parameters that it supports allows the recipient to
+select the preferred set to then use with the handshake protocol.
+
 ### Handling of all Out-of-Band Messages
 
 We currently have two sets of out-of-band messages that cannot be delivered via DIDComm because there is no channel. We'd like to align those messages into a single "out of band" protocol so that their handling can be harmonized inside an agent, and a common QR code handling mechanism can be used.
@@ -98,7 +105,11 @@ The out-of-band protocol a single message that is sent by the *sender*.
   "handshake_protocols": [
       "https://didcomm.org/didexchange/1.0",
       "https://didcomm.org/connections/1.0"
-      ],
+  ],
+  "jwe_suite": [
+      "JWM/1.0_Authcrypt",
+      "didcomm-envelope-enc_X25519_ECDH-1PU+A256KW_XC20P"
+  ],
   "request~attach": [
     {
         "@id": "request-0",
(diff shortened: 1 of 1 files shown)


Baha-sk (approved):

kdenhartog (commented):
We're not currently using the OOB protocol for initial setup and only are using DIDCommV2. How would we be expected to convey this?

llorllale:
@kdenhartog 

> We're not currently using the OOB protocol for initial setup and only are using DIDCommV2. How would we be expected to convey this?

I understand we are punting the switch from DIDComm V1 to V2.

cc @swcurran to see if Stephen has any thoughts on this.

swcurran:
@kdenhartog -- how are you doing the initial setup?

kdenhartog:
> @kdenhartog -- how are you doing the initial setup?

Given right now we're only doing issuance/verification of VCs and DID Auth through OIDC provider we are able to use the 0-RTT functionality to build a DIDComm connection while issuing a VC / Verifying a VC during an OIDC interaction

llorllale:
@kdenhartog 

> > @kdenhartog -- how are you doing the initial setup?
> 
> Given right now we're only doing issuance/verification of VCs and DID Auth through OIDC provider we are able to use the 0-RTT functionality to build a DIDComm connection while issuing a VC / Verifying a VC during an OIDC interaction

Let me re-phrase the question: how are the two parties negotiating (JWE?) encryption parameters?

Should we assume you're using [published OIDC provider metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)?

kdenhartog:
> @kdenhartog
> 
> > > @kdenhartog -- how are you doing the initial setup?
> > 
> > 
> > Given right now we're only doing issuance/verification of VCs and DID Auth through OIDC provider we are able to use the 0-RTT functionality to build a DIDComm connection while issuing a VC / Verifying a VC during an OIDC interaction
> 
> Let me re-phrase the question: how are the two parties negotiating (JWE?) encryption parameters?
> 
> Should we assume you're using [published OIDC provider metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)?

We support all types defined in DIDCommV2 whether optional or required so don't need to negotiate. This is a luxury afforded to us because we're only inter-oping with ourselves at the moment.

In talking with my team we see 3 potential options that we're weighing at the moment.

1. Include the discovery information in the DID Document.
  - this works well for did peer method, but presents challenges for basically every other method type so it probably won't be a good option.

2. Include the discovery information in a public metadata point similar to OIDC
  - This offers a layer of indirection that adds some form of privacy, but because the metadata point still would need to be published in some way (requires effectively the same amount of disclosure as the DID Document - e.g. public ledger DID means public metadata endpoint) we're not certain that the privacy would hold. Furthermore, it offers another point that has to be supported for highly available agents / architected around for semi available agents in order to make discovery work. Additionally, we're not confident in the willingness of others to adopt this method as well at this point so likely won't go this path.
  
3. Support an optional discovery JWM message similar to OOB
 - This is the path that makes the most sense to me at this point because it maps well to the same processes of how you'd discover the did document data with a 0-RTT system while also allowing the ability for an explicit discovery of supported protocols and other data related such that it could be passed at any time rather than only at the beginning. This is important if you want to update your support for data without having to restart the connection. Importantly though, it's not the required first message, but can act as a fall back message if a 0-RTT message is not understood. Particularly what I've been thinking about with this is that the majority of enterprise cases don't allow for any inherent negotiation and instead it's more of a "take it or leave it approach" because of the need to operate at scale. However, when you're interacting in a peer-to-peer environment the assumption that a connection will be set and then doing some form of logic beyond a connection later on seems far more logical and so if you wanted to you could send the discovery message as the first message, but it shouldn't be required in our view. This is the thing that makes a fair amount of sense to us at the moment, but we're open to considering options 1 or 2 / other options if there's strong reasons for doing so.
 
 In summary, right now we don't need to do anything like this because all of our flows operate in a 0-RTT flow and we're supporting all forms of DIDCommV2 encryption. We also recognize the need for this in a subset of the use case we're considering in the future, but have deferred on this to this point because we see it as something that we can upgrade to overtime. Additionally, because we've opted to stick with DIDCommV2 only, we've not spent much time considering the usage of the OOB protocol to date, but want. to make sure that what we consider will align with everyone when Aries begins to converge on DIDCommV2.

llorllale:
@kdenhartog 

- Option 1 does not sound appealing to me because it would complicate even further the encoding of ledger-less DIDs. We are already struggling to come up with ways to "just" encode more than one key in a way that makes sense.
- Option 2 puts an onerous burden on agents that are not always online - just like you said

As for Option 3 - agree it should not be required. In fact, this proposal pretty much enables this use case while still not invalidating the use of the Features Discovery protocol after connection establishment.

Perhaps we can delay discussion about the exact details of how the JWE parameters will be conveyed and instead come to agree on **where** in the protocols this should happen? My proposal is:

* allow the sender of an OOB message to declare their supported JWE parameters _somehow_ (precedent: already disclosing supported handshake protocols)
* also allow an agent to discover the JWE parameters via the discovery protocol

I think the first point is more pressing - we're already discussing adoption of RFC0334 which would smooth the transition to DIDComm V2.

@swcurran @TelegramSam @dhh1128 thoughts?

PR #577, opened 2020-12-18
PROPOSE: RFC 0577 Use self-encoded keys

llorllale:

### DIFF
--- a/features/0577-self-encoded-keys/README.md
+++ b/features/0577-self-encoded-keys/README.md
@@ -1,0 +1,296 @@
+# Aries RFC 0577: Use self-encoded keys
+
+- Authors: [George Aristy](mailto:george.aristy@securekey.com) (SecureKey)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2020-12-18
+- Status Note: N/A
+- Supersedes: [RFC0360](../0360-use-did-key/README.md)
+- Start Date: 2020-12-09
+- Tags: [feature](/tags.md#feature)
+
+## Summary
+
+A simple, lightweight format for a sender to encode cryptographic keys is introduced that unambiguously conveys the components required for their use: the key's curve, the key's value, and the associated algorithm. Further changes to RFCs with "naked" keys are proposed to disambiguate their purpose.
+
+## Scope
+
+This format is intended to replace the existing usage of `did:key` identifiers in certain Aries RFC protocols until a different or evolved solution is proposed that relies on standards and meets the community's requirements.
+
+## Background
+
+The Hyperledger Aries community standards for encoding of cryptographic keys between a sending a receiving party has gone through two iterations.
+
+### First iteration
+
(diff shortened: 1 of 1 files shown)


dhh1128 (commented):
I've read this proposal. It makes a lot of sense to me. I'm not sure I understand all the implications, however. I'd like an interactive discussion to explore; I think it will take some cross-referencing to the thinking of others before I am fully comfortable.

Baha-sk (approved):
LGTM, this will allow separating JWE encryption keys from the main DID keys (ie no need to depend on Ed25519 key conversion to do didComm)

llorllale:
From yesterday's call:

* objections were raised against moving away from using DIDs to express "naked keys"
* @swcurran will reach out to the editors of the `did:key` method spec to inquire as to how disjoint types of keys could be encoded using `did:key`
* @TelegramSam will draft a new DID method spec similar to `did:key` that will allow a sender to encode several keys plus additional metadata that would indicate the key's purpose and algorithms

features/0577-self-encoded-keys/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+The above conventions were captured in [RFC160-connection-protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0160-connection-protocol/README.md) and [RFC0019-encryption-envelop](https://github.com/hyperledger/aries-rfcs/blob/master/features/0019-encryption-envelope/README.md).
+
+### Second iteration
+
+There was a growing awareness of the drawbacks to the convention implemented in the previous iteration. In particular, there was no support for other types of keys (see [cryptographic agility](https://en.wikipedia.org/wiki/Crypto-agility)), and the sender's purpose for them was not explicit.
```
swcurran, line 33: The "no support" line is excessive.  Simply by using multicodec, other keys are supported by definition.  However, I agree with the sentiment that any key that does not support birational equivalence cannot be used in places where both of a pair of keys are needed.

llorllale: @swcurran this paragraph is describing the situation in the first iteration where all naked keys were base58-encoded and were always of type Ed25519, hence "no support for other types of keys".

Regardless, this section is a mere description of past events intended to guide the reader to an understanding of the motivations behind the proposal. It does not form part of the solution's description.

swcurran: Got it.  Thanks.

features/0577-self-encoded-keys/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+```diff
+{
+  "service": [{
+     "id": "did:example:123456789abcdefghi#did-communication",
+     "type": "did-communication",
+     "priority" : 0,
+-    "recipientKeys" : [ "did:example:123456789abcdefghi#1" ],
+-    "routingKeys" : [ "did:example:123456789abcdefghi#1" ],
++    "authnKeys": ["ES384|z2bHnTFkhn3NLWvF4qei6cLBiL5xtvuY8wt74jvi7w9rA3W29i6st9YkmL8EYBsbARTrrN"],
```
swcurran, line 180: Does this need to be ordered list that matches recipientKeys? E.g does the nth authnKeys entry correspond to the nth recipientKeys entry? If so, would it be better to make the ordering more explicit -- e.g. one array, with two items in each?

llorllale: `authnKeys` exists for a signature required in the did-exchange protocol. Ordering does not matter so long as the signature is from one of these keys.

swcurran: Cool - thanks.

PR #580, opened 2021-01-08
pluralize lists, call preview a preview

sklump:
... also tweak json listings to avoid red blotches on display


sklump:
Also, note that the preview is called a preview in the credential offer - it should be called a preview in the proposal

swcurran:
This looks good to me.  @llorllale does the Aries Framework Go team have any concerns with the tweaks here?  They seem correct.

Has anyone else already implemented this protocol?

sklump:
Also: in credential request, should there be text about the replacement id must match any received in the offer, if in response to an offer with a replacement id? Otherwise, the holder must know an issuer-unique identifier, which is awkward.

sklump:
OK, good to go I believe

swcurran:
@llorllale can you please review and approve if OK?  Note that since the Wednesday meeting another field was dropped -- "replacement_id" in the "credential request" message.  It should not be there is probably a cut'n'paste typo from somewhere.

llorllale (approved):

swcurran (approved):
LGTM

PR #582, opened 2021-02-01
Note iOS restriction on custom scheme registration

jmandel:
https://stackoverflow.com/questions/13130442/multiple-apps-with-the-same-url-scheme-ios and https://blog.trendmicro.com/trendlabs-security-intelligence/ios-url-scheme-susceptible-to-hijacking/ have background on this limitation.

### DIFF
--- a/concepts/0268-unified-didcomm-agent-deeplinking/README.md
+++ b/concepts/0268-unified-didcomm-agent-deeplinking/README.md
@@ -69,7 +69,7 @@ Furthermore, messages must also be base64 encoded serialized jsons, stripped of 
 
 This puts extra work on wallet developers to ensure a good experience. 
 
-On iOS it only one app can be registered to handle `didcomm://` at a time; the first one to be installed will prevent others from using this custom scheme.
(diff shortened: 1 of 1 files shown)


concepts/0268-unified-didcomm-agent-deeplinking/README.md
```diff
@@ -69,6 +69,7 @@ Furthermore, messages must also be base64 encoded serialized jsons, stripped of
 
 This puts extra work on wallet developers to ensure a good experience. 
 
+On iOS it only one app can be registered to handle `didcomm://` at a time; the first one to be installed will prevent others from using this custom scheme.
```
TimoGlastra, line 72:
```suggestion
On iOS only one app can be registered to handle `didcomm://` at a time; the first one to be installed will prevent others from using this custom scheme.
```

TimoGlastra:
Thanks @jmandel! Commits need to be DCO signed, sadly this is not possible to do from within github (AFAIK), so you'd need to do this in a local clone.

See https://github.com/hyperledger/aries-rfcs/blob/master/contributing.md#signing-off-commits-dco for more information

TimoGlastra:
Closing in favour of #730

PR #583, opened 2021-02-02
Fix typos where the message type prefix should be https not http

swcurran:


dhh1128 (approved):

dhh1128:
Completely uncontroversial. Editorial fix.

PR #586, opened 2021-02-09
Sign Attachment Protocol

TelegramSam:
This protocol is based off earlier work, and simplifies several things and adds a mechanism to present gathered signatures.


features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+```plantuml
+Coordinater --> Coordinater: Prepare Attachment
+Coordinater -> Signer1: Signature Request
+Signer1 -> Coordinater: Signature Response
+group Additional Signers
+    Coordinater -> SignerN: Signature Request
+    SignerN -> Coordinater: Signature Response
+end
+```
```
andrewwhitehead, line 48: *Coordinator

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+hide empty description
+[*] -> Requested: Request
+Requested -> [*] : Response (Signed)
+Requested -> [*] : Response (Declined)
+```
+## Messages
+
+### Signature Request
+
+This message represents a request from a "coordinator" to a "signer" to provide a signature for the attachment(s) included in the request.  The request will indicate the attached message type as well as the requested signature type, and "goal codes" will be included to indicate the "goal" of the coordinator, and stated goal of the signer.
```
andrewwhitehead, line 70: "the stated goal of the coordinator" for balance maybe?

I feel like maybe there should be only one goal - ie. the goal of the coordinator. The goal (or action) of the signer seems like it would always be implied by that goal.

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+The message attachment will follow the structure defined in Aries RFC 0017 and *may* be signed (using a JWS signature, outside of the attachment itself).
+
+```jsonc
+{
+  "@type": "http://didcomm.org/sign-attachment/%VER/signature-request",
+  "@id": "fce30ed1-96f8-44c9-95cf-b274288009dc",
+  "comment": "some comment",
+  "signature_request": [{
+      "attachment": "#1",
+      "context": "did:sov", //why context?
```
andrewwhitehead, line 81: Context could probably use further explanation, but not here :)

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+      "method": "add-signature",
+      "signature_type": "<requested signature type>",
+      "signer_goal_code": "transaction.endorse",
+      "author_goal_code": "ledger.transaction.write"
+  }],
+  "~attach": [{
+    "@id": "1",
```
andrewwhitehead, line 88: `@id` should match the referenced one, ie. `"#1"`

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+  "@type": "http://didcomm.org/sign-attachment/%VER/signature-response",
+  "@id": "ece30ed1-96f8-44c9-95cf-b274288009dc",
+  "~thread": {
+    "thid": "fce30ed1-96f8-44c9-95cf-b274288009dc"
+  },
+  "signature_response": [{
+      "message_id": "143c458d-1b1c-40c7-ab85-4d16808ddf0a",
+      "context": "did:sov",
+      "method": "add-signature",
+      "signer_goal_code": "transaction.endorse" # or "transaction.refuse",
```
andrewwhitehead, line 130: // for comment

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+      "signature": { ... structure determined by signature type ... }
+  }]
+}
+```
+**signature_response** - An array of signature responses. This should correlate to the number of requested signatures.
+- **message_id**: Corresponds to the `@id` of the message 
+- **context**: same as in request
+- **method**: same as in request
+- **signer_goal_code**: the same as the `signer_goal_code` of the request. If the signature is refused, a different suitable goal code should be used. 
+- **signature_type"**: requested signature type
```
andrewwhitehead, line 141: extra "

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+  },
+  "signatures": [{
+      "message_id": "143c458d-1b1c-40c7-ab85-4d16808ddf0a",
+      "context": "did:sov",
+      "method": "add-signature",
+      "signer_goal_code": "transaction.endorse" # or "transaction.refuse",
+      "signature_type": "<requested signature type>",
+      "signature": { ... structure determined by signature type ... }
+  }],
+  "messages~attach": [{
```
andrewwhitehead, line 166: Shouldn't this be `~attach` to match the request? Although I would prefer just calling it `messages`.

features/0586-sign-attachment/README.md
```diff
@@ -0,0 +1,219 @@
...
+    }
+  }]
+}
+```
+**signatures** - An array of signature responses. This should correlate to the number of requested signatures.
+- **message_id**: Corresponds to the `@id` of the message 
+- **context**: same as in request
+- **method**: same as in request
+- **signer_goal_code**: the same as the `signer_goal_code` of the request. If the signature is refused, a different suitable goal code should be used. 
+- **signature_type"**: requested signature type
```
andrewwhitehead, line 182: extra "

features/0586-sign-attachment/README.md
TelegramSam: fixed

features/0586-sign-attachment/README.md
TelegramSam: fixed

features/0586-sign-attachment/README.md
TelegramSam: fixed.

features/0586-sign-attachment/README.md
TelegramSam: fixed.

features/0586-sign-attachment/README.md
TelegramSam: I've adjusted the language to reflect that this is from the opinion of the coordinator. The one goal discussion is a good one, but I didn't want to hold up the other changes for that discussion.

features/0586-sign-attachment/README.md
TelegramSam: I moved to the attachment append method (as opposed to inline) partially as an experiment to see what it would look like. We are considering only appended attachments in DIDComm V2.

TelegramSam: But - I did make it consistent.

TelegramSam:
@andrewwhitehead small adjustments made. Larger points are good ones, and I've left those open.

dhh1128 (approved):
LGTM

I would note, however, that I think "Please Sign This" is a better name than "Sign Attachment." My reasoning is that "Sign Attachment" makes me want to assume that the signed artifact is intended to be an attachment (a false assumption). On the other hand, "Please Sign This" makes it clear that what's being signed is an attachment to the message in this protocol -- and expresses no opinion whatsoever about how whether the signed artifact will eventually become an attachment on another message.

Change the name or leave it alone; I'm okay with either choice.

Baha-sk:
is this change related to signed attachment in DIDComm?

I'm asking because this [PR](https://github.com/hyperledger/aries-rfcs/pull/626) is about to remove the signing wording of attachments in DIDComm (ie recipientKeys only used for envelope encryption). If they're related, then we have a conflict.

Or if they're related and both needed, then this PR must specify where the signing key comes from. It cannot be stored in DIDComm's `recipientKeys` section. It should be resolved from the signer's DID doc's `Authentication` section.

Exulansis:
Can someone point me to where I can find some further info / guidance on the purpose and allowed values for the `context`, `method`, and `signature_type` fields defined in this RFC?

For a bit of context, I am trying to figure out if the protocol defined in this RFC can be used (perhaps in combination with the attachment formats defined in [Aries RFC 0593](https://github.com/hyperledger/aries-rfcs/tree/master/features/0593-json-ld-cred-attach)) to request / produce [Linked Data Proof](https://w3c-ccg.github.io/ld-proofs/) signatures.

Thank you!

pstuermlinger:
Hello,
I want to take the opportunity to propose a change in the allowed values of the `signer_goal_code` field and associated behavior.

**Motivation**
Currently, it seems like the goal code `ledger.transaction.write` is a valid goal only for the author. However, once a transaction/attachment has enough signatures, there's no technical reason why the endorser shouldn't write the transaction to the ledger by itself. If the RFC would allow such a behavior, it would enable author-agents within highly restricted enterprise or government networks to utilize a priviledged endorser not only as signer but also as a messenger.

**Proposed solution**
Change the type of `signer_goal_code` from  string to array. This allows an author to set the signer goal to either `["transaction.endorse"]`, `["transaction.endorse", "ledger.transaction.write"]` or only `["ledger.transaction.write"]`.
For the sake of accuracy and possible future developments, I suggest following the order.

In case of `ledger.transaction.write` as (last) goal, the Signature Response should provide the ledger write result (e.g. the created schema definition) as attachment.

**Related discussions / issues**
- https://github.com/hyperledger/aries-cloudagent-python/issues/1031
- https://github.com/hyperledger/aries-cloudagent-python/issues/736

If my proposal is accepted, I'm happy to make the necessary changes in the document.

Exulansis:
> Can someone point me to where I can find some further info / guidance on the purpose and allowed values for the `context`, `method`, and `signature_type` fields defined in this RFC?
> 
> For a bit of context, I am trying to figure out if the protocol defined in this RFC can be used (perhaps in combination with the attachment formats defined in [Aries RFC 0593](https://github.com/hyperledger/aries-rfcs/tree/master/features/0593-json-ld-cred-attach)) to request / produce [Linked Data Proof](https://w3c-ccg.github.io/ld-proofs/) signatures.

I think this question would be resolved with some guidance on how to define / register application specific protocols based on this RFC document (e.g. for usage with Linked Data Documents / [Linked Data Proofs)](https://w3c-ccg.github.io/ld-proofs/).

The RFC currently states:

> The result of this is that any application of this protocol requires another spec for the application of this protocol. This presents an issue for compatibility detection and increases the complexity of handling these messages for agents.

I am curious if examples of such application specific specs already exist (I was so far only able to find [this document](https://hackmd.io/5LzMhfsMQBevB5V2tKz4hA?view)), or if any are being developed.

Thank you.

dhh1128:
I am sensitive to the pending questions in this conversation -- but I don't feel like I can answer them. I think @TelegramSam or one of the folks on the BCGov team needs to speak to the intent and to whether the problem domain of the protocol overlaps with collecting Linked Data Proof signatures enough to justify solving that problem this way.

PR #587, opened 2021-02-10
RFC 0587 Encryption Envelope v2

troyronda:

### DIFF
--- a/features/0587-encryption-envelope-v2/README.md
+++ b/features/0587-encryption-envelope-v2/README.md
@@ -1,0 +1,95 @@
+# Aries RFC 0587: Encryption Envelope v2
+
+- Authors: [Baha A. Shaaban](mailto:baha.shaaban@securekey.com) (SecureKey Technologies Inc.), [Troy Ronda](mailto:troy.ronda@securekey.com) (SecureKey Technologies Inc.)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2021-02-10
+- Status Note:
+- Supersedes:
+- Start Date: 2021-02-10
+- Tags: [feature](/tags.md#feature)
+
+## Summary
+
+This RFC proposes that we support the definition of envelopes from [DIDComm Messaging](https://identity.foundation/didcomm-messaging/spec).
+
+## Motivation
+
+This RFC defines ciphersuites for envelopes such that we can achieve better compatability with DIDComm Messaging being specified at DIF.
+The ciphersuites defined in this RFC are a subset of the definitions in [Aries RFC 0334-jwe-envelope](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope).
+
+## Encryption Algorithms
+
+[ECDH-1PU draft 03](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) defines the JWE structure. The following sections summarize the supported algorithms.
+
+### Curves
+
+DIDComm Messaging (and this RFC) requires support for `X25519`, `P-256`, and `P-384`.
+
+- P-256 (reference in [RFC4492](https://tools.ietf.org/search/rfc4492#appendix-A))
+- P-384 (reference in [RFC4492](https://tools.ietf.org/search/rfc4492#appendix-A))
+- P-521 (reference in [RFC4492](https://tools.ietf.org/search/rfc4492#appendix-A))
+- X25519 (reference in [RFC7748](https://tools.ietf.org/html/rfc7748#section-5))
+
+### Content Encryption Algorithms
+
+DIDComm Messaging (and this RFC) requires support for both `XC20P` and `A256GCM`.
+
+- XC20P (XChaCha20Poly1305 - reference in [xchacha draft 03](https://tools.ietf.org/html/draft-irtf-cfrg-xchacha-03))
+- A256GCM (AES-GCM with a 256 bit key - reference in [RFC7518](https://tools.ietf.org/html/rfc7518#section-5.1))
+
+### Key Wrapping Algorithms
+
+DIDComm Messaing (and this RFC) requires support for `ECDH-1PU+A256KW`.
+
+- ECDH-1PU+A256KW (defined in [ECDH-1PU draft 03](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03#section-2.2))
+
+## Anoncrypt equivalent
+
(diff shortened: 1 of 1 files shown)


llorllale (approved):

kdenhartog (approved):
Using this for AIP 2.0 is going to make the transition much smoother

Baha-sk (approved):

swcurran (changes requested):
Could you add how we would move to this in existing frameworks?   Will we need a community update approach?  I imagine that same question goes for RFC 0334.  This one talks about what we should do, but not about how we get there.

Dummy question: It also looks like we are saying we all MUST include multiple crypto suites vs. negotiating what we support.  Is that right?

Thanks

llorllale:
> negotiating what we support

I proposed #576 a while back. I'm not fond of its details, but the key point is in communicating the supported suites in the OOB invitation when bootstrapping a connection. This can be complemented with the Features Discovery protocol as well (as per the discussion over there).

llorllale:
> Will we need a community update approach?

That depends on the community.

SecureKey is comfortable with jumping into this without intermediate steps.

swcurran:
To clarify -- will we need a community update so that other frameworks that are deployed in the wild can move to this?  We need to keep interop going amongst the components for which interop is already working.

@andrewwhitehead - can you please take a look and assess how we would move to this?

troyronda:
@swcurran Here's an example community update:

-  Implement new envelope format and algorithms.
- Add property to OOB to indicate envelope "version" (easier to define) or supported individual "suites" (harder to define).
- Frameworks that support the newer version use the newer version when the OOB property indicates the other side does as well.
- Transition out of supporting old format. (hopefully not too long of a timeline).

troyronda:
@swcurran I have created a sibling PR: https://github.com/hyperledger/aries-rfcs/pull/595 that adds an `accept` property to the OOB message. Using the `accept` property in the OOB message, the acceptable envelope type(s) can be determined by the receiver of an OOB message.

troyronda:
Updated with:

- renamed to encryption-envelope-v2.
- mime type clarification.
- commentary about the DIDComm v2 transition.

features/0587-encryption-envelope-v2/README.md
```diff
@@ -0,0 +1,87 @@
...
+## Motivation
+
+This RFC defines ciphersuites for envelopes such that we can achieve better compatability with DIDComm Messaging being specified at DIF.
+The ciphersuites defined in this RFC are a subset of the definitions in [Aries RFC 0334-jwe-envelope](https://github.com/hyperledger/aries-rfcs/tree/master/features/0334-jwe-envelope).
+
+## Encryption Algorithms
+
+[ECDH-1PU draft 03](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) defines the JWE structure. The following sections summarize the supported algorithms.
+
+### Curves
```
troyronda, line 24: DIDComm v2 requires at least X25519 and P-256 and also allows for P-384 and P-521.

Note - It’s worthwhile to also support P-384 and would be preferable for some implementers as compared to P-256.

troyronda:
Added clarification on the suites required by this RFC and context on the DIDComm Messaging requirements.

features/0587-encryption-envelope-v2/README.md
troyronda: Added clarification text to RFC.

features/0587-encryption-envelope-v2/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Anoncrypt equivalent
+
+For situations where the envelopes should be anonymous, we use a newly minted DID rather than a different Anoncrypt mechanism.
+The newly minted DID is used once and then discarded.
+This approach matches the DIDComm Messaging mechanism.
+
+## Mime Type
+
+The mime type associated to this envelope is `application/didcomm-encrypted+json`.
```
troyronda, line 56: The DIDComm Messaging spec currently states: `application/didcomm-enc-env`
https://identity.foundation/didcomm-messaging/spec/#https

RFC 0044 currently states (for v2): `application/didcomm-encrypted+json`

features/0587-encryption-envelope-v2/README.md
```diff
@@ -0,0 +1,95 @@
...
+[RFC 0044](../0044-didcomm-file-and-mime-types/README.md) provides a general discussion of mime types.
+
+## DIDComm v2 Transition
+
+As this RFC specifies the same envelope format as will be used in DIDComm v2, an implementor should detect if the payload contains DIDComm v1 content or the JWM from DIDComm v2.
```
llorllale, line 61: https://github.com/decentralized-identity/didcomm-messaging/issues/165

features/0587-encryption-envelope-v2/README.md
kdenhartog: Looks good to me based on what we have in the DIDCommV2 spec at the moment.

features/0587-encryption-envelope-v2/README.md
troyronda: Resolved in https://github.com/decentralized-identity/didcomm-messaging/pull/168.

andrewwhitehead:
It's a bit verbose, but I think another option would be to wrap the current agent message format in a JWM, to avoid having to do content type detection. For example:

```jsonc
{
    "id": "518be002-de8e-456e-b3d5-8fe472477a86",
    "type": "https://didcomm.org/compat/1.0/agent_message",
    "from": "did:example:alice",
    "to": ["did:example:bob"],
    "created_time": 1516269022,
    "expires_time": 1516385931,
    "body": {
        "@type": "https://didcomm.org/trust_ping/1.0/ping",
        "@id": "518be002-de8e-456e-b3d5-8fe472477a86",
        "~timing": {
            "out_time": "2018-12-15 04:29:23Z",
            "expires_time": "2018-12-15 05:29:23Z",
            "delay_milli": 0
        },
        "comment": "Hi. Are you listening?",
        "response_requested": true
    }
}
```

troyronda:
@swcurran @TelegramSam @andrewwhitehead @llorllale I have updated this PR to state that the `cty` property of the JWE will distinguish DIDComm v1 vs DIDComm v2.

dhh1128 (approved):

dhh1128:
LGTM.

kdenhartog (approved):
I like the new `CTY` answer for handling DIDCommV1 vs V2. It's such an obvious and simple solution. Great idea!

swcurran (approved):

PR #588, opened 2021-02-11
RFC0360: add RFC0211 and 0434 to the list

llorllale:

### DIFF
--- a/features/0360-use-did-key/README.md
+++ b/features/0360-use-did-key/README.md
@@ -60,6 +60,8 @@ The following currently implemented RFCs would be affected by acceptance of this
 - [0028-introduce](https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce)
 - [0056-service-decorator](https://github.com/hyperledger/aries-rfcs/tree/master/features/0056-service-decorator)
 - [0160-connection-protocol](https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol)
+- [0434-out-of-band-protocols](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md)
+- [0211-mediator-coordination-protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md)
 
 Service entries in `did:peer` DIDDocs (such as in RFCs
 [0094-cross-domain-messaging](https://github.com/hyperledger/aries-rfcs/tree/master/concepts/0094-cross-domain-messaging)


features/0360-use-did-key/README.md
```diff
@@ -60,6 +60,8 @@ The following currently implemented RFCs would be affected by acceptance of this
 - [0028-introduce](https://github.com/hyperledger/aries-rfcs/tree/master/features/0028-introduce)
 - [0056-service-decorator](https://github.com/hyperledger/aries-rfcs/tree/master/features/0056-service-decorator)
 - [0160-connection-protocol](https://github.com/hyperledger/aries-rfcs/tree/master/features/0160-connection-protocol)
+- [0434-out-of-band-protocols](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md)
+- [0211-mediator-coordination-protocol](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md)
```
swcurran, line 64: Looks like 0211 needs an update to reference where we're using DID:key.

@dbluhm @burdettadam  -- can you confirm that in the implementation you have done that did:key is being used?

dbluhm: The current implementation is naively assuming the values in recipient keys and routing keys are base58 encoded ed25519 verkeys so we'll need an update to fix that. We adhered pretty strictly to the RFC as it stood lol.

swcurran: So for now, this should leave off this from the list of did:key until the 0211 gets updated.

@dbluhm @TelegramSam -- thoughts?  Are you willing to update the RFC / should it be?

dbluhm: I think it should be updated. The keys passed around in the coordinate mediation protocol are used for two purposes: as an identifier for a connection that doesn't require additional information to be passed between agents during did exchange/connection (recipient keys) and as the key used to encrypt the forward messages (routing keys). In both cases, I think having the key values be something unambiguous for use in messaging with agents that are not strictly ed25519 based will be important for interop, even though in the identifier case it's not actually used for any encryption, decryption etc.

swcurran: Do we have a volunteer to update 0211?  :-)

llorllale: See https://github.com/hyperledger/aries-rfcs/pull/596

llorllale:
@TelegramSam @swcurran this PR is now up to date with `master`

dhh1128:
approved on community B call 24 Mar 2021

Issue #585, opened 2021-02-03
Issue Credential v2 Credential Preview doesn't allow nested attributes


TimoGlastra:
The [Credential Preview](https://github.com/hyperledger/aries-rfcs/tree/master/features/0453-issue-credential-v2#preview-credential) inner message used in the `propose-credential` and `offer-credential` messages only supports one level of attribute nesting. 

This works fine for indy credentials, as they only support one level of attribute nesting. However the VC Data model allows the credential subject to be any object. The credential preview can't be used to convey a nested attribute structure.

You could use mime-type to show the attribute is a json object, however the value should be a base64 encoded string in this case and suggests to me that you would also include it as a base64 encoded string in the credential itself.

```json
{
    "@type": "https://didcomm.org/issue-credential/%VER/credential-preview",
    "attributes": [
        {
            "name": "person",
            "mime-type": "application/json",
            "value": "<base64 encoded person object>"
        },
    ]
}
```

Looking at present proof v2 the preview message is removed and everything is handled using the attachment format. Is there a specific reason issue credential v2 didn't take this approach?

I think we should either allow the credential preview to support nested attribute structures or remove it and require the attachment format to handle this. 

Thoughts?

swcurran:
I agree this should have been done.  It's a bit tricky what the format should be on an initial message, as I'm assuming the holder should be able to simply propose a credential without proposing the data for the credential, since they don't hold it.  If they are responding to a credential offer with a counter-proposal, they can modify the credential data in the credential offer.

TimoGlastra:
Well I think that is already possible by using the attachment formats. The credential preview is optional, so you just send a credential proposal with only the attachment format. For indy that'll allow you to specify a  schema / cred def / etc.. for the DIF format you can attach a credential manifest.

So I think it should be just about the attributes, conveying everything else in the attachment format.

Easiest format I can think of is to allow for nested `attributes`. This way the old format is still valid and we can update to a `2.1` version. 

```json
{
    "@type": "https://didcomm.org/issue-credential/%VER/credential-preview",
    "attributes": [
        {
            "name": "person",
            "attributes": [
                {
                     "name": "first_name",
                     "value": "John"
                },
                {
                     "name": "last_name",
                     "value": "Smith"
                }
            ]
        },
        {
            "name": "country",
            "value": "NL"
        }
    ]
}
```

As an addition I think we should also the preview to be included in a credential request, as it is allowed for the holder to start from a credential request.

[...]

PR #591, opened 2021-02-23
extend issue credential v2 credential preview

TimoGlastra:
Extends the issue credential v2 credential preview as described in #585.

1. Allow for nested attributes in credential preview to better support W3C VC data structures
2. Allow for credential-preview in credential-request to allow holder initiate from credential-request with credential-preview
3. Allow to embed mime-type in attribute value of credential-preview using data URL. 

I wasn't really sure how the versioning works. In v1 of the protocol it seems like messages are versioned separately instead of moving the whole protocol to 1.1.

Regarding 3. this is added to allow more flexibility for specifying the mime-type of an attribute. Currently the value could only be a base64 encoded BLOB and the mime-type needed to be specified in the `mime-type` field. Two problems I see:
1. I see a lot usages of VCs that embed images as data URLs. If that's what people use, we should allow Aries agents to issue such credentials. (example: https://w3c-ccg.github.io/vc-examples/cmtr/examples/v0.2/cmtr-verifiable-credential-v0.2.json)
2. Embedding the mime-type allows the mime-type to be discoverable outside of Aries because the credential-preview metadata is not required anymore.

### DIFF
--- a/features/0453-issue-credential-v2/README.md
+++ b/features/0453-issue-credential-v2/README.md
@@ -1,6 +1,6 @@
 # Aries RFC 0453: Issue Credential Protocol 2.0
 
-- Authors: Nikita Khateev, Stephen Klump, Stephen Curran
+- Authors: Nikita Khateev, Stephen Klump, Stephen Curran, Timo Glastra
 - Status: [PROPOSED](/README.md#proposed)
 - Since: 2020-03-23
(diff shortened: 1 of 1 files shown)


dhh1128 (approved):
LGTM

TimoGlastra:
Closing as json-ld credentials will use the attachment format to convey credential data. See #593

PR #593, opened 2021-02-24
RFC 0593: JSON-LD Credential Attachment format for requesting and presenting credentials

TimoGlastra:
Still needs some touch-ups, but the general flow is worked out. 

I think we also need to add an `subjectId` property, but still thinking of the implications of how it should be authentically bound to the subject / the didcomm connection

### DIFF
--- a/features/0593-json-ld-cred-attach/README.md
+++ b/features/0593-json-ld-cred-attach/README.md
@@ -1,6 +1,6 @@
 # Aries RFC 0593: JSON-LD Credential Attachment format for requesting and presenting credentials
 
-- Authors: Timo Glastra (Animo Solutions)
+- Authors: Timo Glastra (Animo Solutions), George Aristy (SecureKey Technologies)
 - Status: [PROPOSED](/README.md#proposed)
 - Since: 2021-02-24
 - Status Note:
@@ -30,7 +30,7 @@ Format identifier: `aries/ld-proof-vc-proposal@v1.0`
 
 The credential proposal allows the holder to initiate a credential exchange without fully specifying the contents of the credential. If all properties of the credential are already known, the protocol can also be initiated with a credential request. While the credential detail requires all fields to be present, the proposal allows for fields to be omitted.
 
-The below `propose-credential` message shows an example of a credential proposal with the `@context`, `type`, `credentialSubject` and `credentialSubject` present. The issuer can then send a credential detail in a credential offer message adding extra properties required for issuance such as `issuer`, `credentialStatus.type`, `proofType`, etc..
+The below `propose-credential` message shows an example of an credential proposal with the `@context`, `type`, `credentialSubject` and `options.credentialStatus` present. The issuer can then send a credential detail in an `offer-credential` message adding extra properties required for issuance such as `issuer` and `options.proofType`.
 
 ```json
 {
@@ -49,21 +49,24 @@ The below `propose-credential` message shows an example of a credential proposal
       "mime-type": "application/ld+json",
       "data": {
         "json": {
-          "@context": [
-            "https://www.w3.org/2018/credentials/v1",
-            "https://www.w3.org/2018/credentials/examples/v1"
-          ],
-          "type": ["VerifiableCredential", "UniversityDegreeCredential"],
-          "credentialSubject": {
-            "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-            "degree": {
-              "type": "BachelorDegree",
-              "name": "Bachelor of Science and Arts"
+          "credential": {
+            "@context": [
+              "https://www.w3.org/2018/credentials/v1",
+              "https://www.w3.org/2018/credentials/examples/v1"
+            ],
+            "type": ["VerifiableCredential", "UniversityDegreeCredential"],
+            "credentialSubject": {
+              "id": "did:key:z6MkodKV3mnjQQMB9jhMZtKD9Sm75ajiYq51JDLuRSPZTXrr",
+              "degree": {
+                "type": "BachelorDegree",
+                "name": "Bachelor of Science and Arts"
+              }
             }
           },
-          "credentialSchema": {
-            "id": "https://example.org/examples/degree.json",
-            "type": "JsonSchemaValidator2018"
+          "options": {
+            "credentialStatus": {
+              "type": "CredentialStatusList2017"
+            }
           }
         }
       }
@@ -72,8 +75,22 @@ The below `propose-credential` message shows an example of a credential proposal
 }
 ```
 
-- `credentialStatus.id` -- Credential Status may be omitted completely, but if present should contain the `type` property to indicate which mechanism to use. As the id can be closely tied to the issuance process the property is not required for the credential proposal.
-- `proofType` -- This is not a property defined by the VC data model, but indicates the proof types the holder proposes for the credential. Entries match the `type` property of the [Proofs](https://www.w3.org/TR/vc-data-model/#proofs-signatures) object. Multiple entries indicate the holder proposes multiple proofs for the credential.
+- `credential` - Required. JSON-LD Credential as proposed by the holder. Properties should align with the criteria described in the [Verifiable Credential Data Model](https://www.w3.org/TR/vc-data-model). The properties listed below are formally supported, but additional properties MAY be included if it conforms with the data model. A note indicates the property deviates from the criteria as defined by the data model.
+
+  - `@context`
+  - `type`
+  - `id`
+  - `issuer` - Optional
+  - `issuanceDate` - Optional
+  - `expirationDate`
+  - `credentialSubject` - Optional. Allows the holder to propose a credential without specifying the claims.
+  - `credentialSchema`
+
+- `options` - Additional options for the credential.
+
+  - `proofType` - Optional list of proof types proposed by the holder to use for the linked data proof. Entries should match suites registered in the [Linked Data Cryptographic Suite Registry](https://w3c-ccg.github.io/ld-cryptosuite-registry/).
+  - `credentialStatus` - Optional object defining the credential status mechanism proposed by the holder to use for the credential. MAY contain additional properties if required for the status method.
+    - `type` - Credential status type proposed by the holder to use for the credential. Required if `credentialStatus` object is present.
 
 ### `ld-proof-vc-detail` attachment format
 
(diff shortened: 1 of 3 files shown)


features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,249 @@
...
+  "formats": [
+    {
+      "attach_id": "d60c05ac-35c2-484c-af3e-8ef7f21baf0a",
+      "format": "aries/vc-ld-proof@v1.0"
+    }
+  ],
+  "offers~attach": [
+    {
+      "@id": "d60c05ac-35c2-484c-af3e-8ef7f21baf0a",
+      "mime-type": "application/json",
```
kdenhartog, line 98: Same thing should this be updated to "application/ld+json"?

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,249 @@
...
+  "formats": [
+    {
+      "attach_id": "13a3f100-38ce-4e96-96b4-ea8f30250df9",
+      "format": "aries/vc-ld-proof@v1.0"
+    }
+  ],
+  "requests~attach": [
+    {
+      "@id": "13a3f100-38ce-4e96-96b4-ea8f30250df9",
+      "mime-type": "application/json",
```
kdenhartog, line 147: same here as above

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,249 @@
...
+  "formats": [
+    {
+      "attach_id": "64a11985-79a9-4daa-b0e3-0ae0531e7a14",
+      "format": "aries/vc-ld-proof@v1.0"
+    }
+  ],
+  "filter~attach": [
+    {
+      "@id": "64a11985-79a9-4daa-b0e3-0ae0531e7a14",
+      "mime-type": "application/json",
```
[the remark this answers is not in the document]
TimoGlastra: Thanks, updated!

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Reference
+
+### `ld-proof-vc-proposal` attachment format
+
+Format identifier: `aries/ld-proof-vc-proposal@v1.0`
+
+The credential proposal allows the holder to initiate a credential exchange without fully specifying the contents of the credential. If all properties of the credential are already known, the protocol can also be initiated with a credential request. While the credential detail requires all fields to be present, the proposal allows for fields to be omitted.
+
+The below `propose-credential` message shows an example of a credential proposal with the `@context`, `type`, `credentialSubject` and `credentialSubject` present. The issuer can then send a credential detail in a credential offer message adding extra properties required for issuance such as `issuer`, `credentialStatus.type`, `proofType`, etc..
```
llorllale, line 33:
```suggestion
The below `propose-credential` message shows an example of a credential proposal with the `@context`, `type`, `credentialSubject` and `credentialSchema` present. The issuer can then send a credential detail in a credential offer message adding extra properties required for issuance such as `issuer`, `credentialStatus.type`, `proofType`, etc..
```

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }
+      }
+    }
+  ]
+}
+```
+
+- `credentialStatus.id` -- Credential Status may be omitted completely, but if present should contain the `type` property to indicate which mechanism to use. As the id can be closely tied to the issuance process the property is not required for the credential proposal.
+- `proofType` -- This is not a property defined by the VC data model, but indicates the proof types the holder proposes for the credential. Entries match the `type` property of the [Proofs](https://www.w3.org/TR/vc-data-model/#proofs-signatures) object. Multiple entries indicate the holder proposes multiple proofs for the credential.
```
llorllale, line 76: Registry for LD proof types: https://w3c-ccg.github.io/ld-cryptosuite-registry/

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,206 @@
...
+            "type": "JsonSchemaValidator2018"
+          }
+        }
+      }
+    }
+  ]
+}
+```
+
+- `credentialStatus.id` -- Credential Status may be omitted completely, but if present should contain the `type` property to indicate which mechanism to use. As the id can be closely tied to the issuance process the property is not required for the credential proposal.
```
llorllale, line 75:
```suggestion
- `credentialStatus` -- Credential Status may be omitted completely, but if present should contain the `type` property to indicate which mechanism to use. As the id can be closely tied to the issuance process the property is not required for the credential proposal.
```

llorllale: Also, we should be careful with this. An existing VC parser with a disabled proof check should be able to parse this object and not raise an error. This means `credentialStatus.id` [MUST be present and MUST be a URL](https://www.w3.org/TR/vc-data-model/#status).

llorllale: Might be better to put this in an "options" object, much like the issuer vc-http-api: https://w3c-ccg.github.io/vc-http-api/#tag/Issuer

features/0593-json-ld-cred-attach/README.md
llorllale: This also seems like a good candidate for an `options` object.

llorllale (changes requested):
Concerns around including properties not defined in the VC data model into the unsigned VC. These should be moved into a sibling "options" object because: a) they are not signed over, and b) they are not part of the data model.

features/0593-json-ld-cred-attach/README.md
TimoGlastra: There are more fields that have this limitation. `A verifiable credential MUST have an issuer property.`. A holder should be able to propose a credential without the `issuer` IMO.

But I totally agree on the `options` object. Seems like a better place

features/0593-json-ld-cred-attach/README.md
llorllale: I am curious about the inclusion of the `credentialSchema` property.

Since the VC data model is heavily biased towards JSONLD, we have been using the top-level JSONLD `@context` to determine the vocabulary (aka. "schema") of the credential. Is the intent here to use `credentialSchema` for just that?

features/0593-json-ld-cred-attach/README.md
llorllale: I propose the `issuer` field MUST be provided in order to satisfy parsers.

However, for the purposes of this RFC, I propose the issuer MUST ignore the value provided by the Holder and instead use their own. Otherwise the entire message would need to be made non-repudiable, which is out of scope of this RFC.

features/0593-json-ld-cred-attach/README.md
TimoGlastra: I think the `schema.uri` from PE is a good reason for `credentialSchema` to be included. This allows to filter credentials by their `credentialsSchema` in PE.

Looking at your implementation that's what you're doing already:

https://github.com/hyperledger/aries-framework-go/blob/cb1ce0da0c3b5d76db78444719268b2d00a4d2f9/pkg/doc/presexch/definition.go#L358

features/0593-json-ld-cred-attach/README.md
TimoGlastra: Moved `proofType` to options and added link to registry

TimoGlastra:
Updated the PR according to feedback and discussion. I've left the optional fields optional, but should be a small update to make them required if we decide to follow the vc-data-model

swcurran (approved):
LGTM

features/0593-json-ld-cred-attach/README.md
llorllale: We have inconsistent behaviour in AFGO that we need to fix:

* Match(): used by the Verifier. Matches `schema.uri` against `@context`
* CreateVP(): used by the Holder. Matches `schema.uri` against `credentialSchema.id`

features/0593-json-ld-cred-attach/README.md
llorllale: Note: although its presence is required by the VC data model, this property definitely makes sense to NOT be included in the credential "template" in our use case for obvious reasons

features/0593-json-ld-cred-attach/README.md
llorllale: > Might be better to put this in an "options" object, much like the issuer vc-http-api: https://w3c-ccg.github.io/vc-http-api/#tag/Issuer

Note that the optionality of `credentialStatus` is not clear at the moment - I posted the question: https://github.com/w3c/vc-data-model/issues/768

llorllale: @TimoGlastra I think it makes sense to have the Holder include all fields required of a verifiable credential in the credential object (with the obvious exception of `proof`).

Allowing the Holder to define who the issuer is opens this up for more possibilities and a wider variety of use cases. Same thing goes for `issuanceDate`. Basically, I think both these fields should be REQUIRED. And again - in general, I think all fields required by the VC data model MUST be present except for `proof`

features/0593-json-ld-cred-attach/README.md
llorllale: We agreed to use the full JSONLD type URI as shown in an example [here](https://github.com/decentralized-identity/presentation-exchange/issues/134#issuecomment-721624167)

TimoGlastra:
@llorllale and I decided to require that all properties must match the vc-data-model. Therefore the proposal format is removed, as there is no distinction between the proposal and request/offer anymore

features/0593-json-ld-cred-attach/README.md
```diff
@@ -0,0 +1,202 @@
...
+  "formats": [
+    {
+      "attach_id": "5b38af88-d36f-4f77-bb7a-2f04ab806eb8",
+      "format": "aries/ld-proof-vc@v1.0"
+    }
+  ],
+  "credentials~attach": [
+    {
+      "@id": "5b38af88-d36f-4f77-bb7a-2f04ab806eb8",
+      "mime-type": "application/ld+json",
```
llorllale, line 171: Note: I think mime-type should be `application/json` because the enclosing object is JSON

llorllale (approved):

features/0593-json-ld-cred-attach/README.md
TimoGlastra: This is the case for the cred-detail, but the issued credential is fully json-ld

llorllale: Ah sorry - my comment was misplaced. It applies to line 84 of this file.

TimoGlastra: updated

PR #594, opened 2021-02-24
Indy attachments RFC

dhh1128:


TimoGlastra (approved):
LGTM. 

Only note is what I already mentioned on the WG call yesterday about the change in the proposal messages. But that's why it's good there is an RFC now :)

dhh1128:
As far as I know, this is now ready to merge.

TelegramSam (approved):
Discussed in WG B 20200303

PR #595, opened 2021-02-26
RFC0434: add accept property to indicate media type

troyronda:
adds the accept property to allow the sender to indicate its
preference for acceptable mime types.

### DIFF
--- a/features/0434-outofband/README.md
+++ b/features/0434-outofband/README.md
@@ -1,4 +1,4 @@
-# Aries RFC 0434: Out-of-Band Protocols
+# Aries RFC 0434: Out-of-Band Protocol 1.1
 
 - Authors: [Ryan West](ryan.west@sovrin.org), [Daniel Bluhm](daniel.bluhm@sovrin.org), Matthew Hailstone, [Stephen Curran](swcurran@cloudcompass.ca), [Sam Curren](sam@sovrin.org), [George Aristy](george.aristy@securekey.com)
 - Status: [PROPOSED](/README.md#proposed)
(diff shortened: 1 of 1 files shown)


features/0434-outofband/README.md
```diff
@@ -119,6 +123,9 @@ The items in the message are:
 - `label` - [optional] a self-attested string that the receiver may want to display to the user, likely about who sent the out-of-band message.
 - `goal_code` - [optional] a self-attested code the receiver may want to display to the user or use in automatically deciding what to do with the out-of-band message.
 - `goal` - [optional] a self-attested string that the receiver may want to display to the user about the context-specific goal of the out-of-band message.
+- `accept` - [optional] an array of mime types in the order of preference of the sender that the receiver can use in responding to the message. 
+If `accept` is not specified, the receiver uses its preferred choice to respond to the message (sender accepts any).
```
kdenhartog, line 127: I like this idea of this when an OOB message is being used. How would we use this with 0-RTT when only one type is supported though?

For example, we're directly sending a plaintext VC request template for our first message (over a TLS channel) and only supporting didcommV2. I think this wouldn't end up solving interop concerns with our implementation though because we opt to lead with a 0-RTT message rather than an OOB message which would be used to specify this.

troyronda: @kdenhartog I think this concept carries over to other similar scenarios. When the first message is not DIDComm, a header should be used to indicate what the sender accepts. This PR generally enables the `accept` property (regardless of transport) for those using OOB. If you are not using OOB, you could use an equivalent for your situation (e.g., the `accept` HTTP header).

troyronda: Also note - I used the `accept` property name in this PR based on the `accept` HTTP header.

llorllale:
We should discuss this further. Mime types should not be used to negotiate or announce encryption parameters. This holds true for V1 and V2.

troyronda:
> Mime types should not be used to negotiate or announce encryption parameters.

@llorllale This parameter isn't explicitly negotiating encryption parameters but rather the content type of the envelope. Since Aries/DIDComm have multiple content types defined, there should be a property for stating what's supported regardless.

For situations where there are multiple implementation options for encryption parameters, then an additional property to indicate supported (and preferred) options would also be advisable.

Baha-sk (approved):

kdenhartog (approved):

troyronda:
The sibling PR in DIDComm v2 has been merged: https://github.com/decentralized-identity/didcomm-messaging/pull/167

dhh1128:
Approved on community B call, 24 Mar 2021

dhh1128:
Approved on community B call, 24 Mar 2021

PR #596, opened 2021-03-03
RFC0211: encode keys using did:key

llorllale:
As per RFC0360.


TelegramSam (approved):

llorllale:
@TelegramSam branch updated

swcurran (approved):
thanks!

Issue #599, opened 2021-03-04
Multiple format identifiers for attachment registries adds unnecessary complexity


TimoGlastra:
Issue credential and Present proof v2 add a registry to each message for different attachment formats. Each message now uses different format identifiers and mime-types. 

Now looking at it from an implementation perspective this adds complexity without adding more information. Having a unique format identifier for each message provides no value over having the same identifier for each message as the expected content can be identified from the message + format identifier combination. 

What's the difference in context between the two tables below?

| `@type`                                   | format                      | mime-type                        |
| ----------------------------------------- | --------------------------- | -------------------------------- |
| `/issue-credential/v2/propose-credential` | `hlindy/cred-abstract@v2.0` | `application/indy-cred-abstract` |
| `/issue-credential/v2/request-credential` | `hlindy/cred-req@v2.0`      | `application/indy-cred-req`      |
| `/issue-credential/v2/issue-credential`   | `hlindy/cred@v2.0`          | `application/indy-cl-anoncred2`  |

| `@type`                                   | format             | mime-type          |
| ----------------------------------------- | ------------------ | ------------------ |
| `/issue-credential/v2/propose-credential` | `hlindy-cred@v2.0` | `application/json` |
| `/issue-credential/v2/request-credential` | `hlindy-cred@v2.0` | `application/json` |
| `/issue-credential/v2/issue-credential`   | `hlindy-cred@v2.0` | `application/json` |

Using a single format identifier (+ simpler mime-types) for all messages will require a lot less checks and still carry the same meaning.

I also used multiple format identifiers in #593, but now it's still possible to change all formats to use a single identifier.

Thoughts?

sklump:
I agree 100% - thanks Timo

sklump:
Similarly for RFC 454: all types should be something like `hlindy-pres@v2.0`.

sklump:
Note that https://github.com/hyperledger/aries-rfcs/pull/607 incorporates the above, with hlindy@v2.0 and dif@cred-manifest@v1.0, dif@pres-exch@v1.0.

PR #607, opened 2021-03-11
Rfc592 cred proposal filter

sklump:
Also set MIME types to application/json consistently, touch up old references to hlindy-zkp etc


features/0453-issue-credential-v2/README.md
```diff
@@ -76,7 +76,7 @@ This protocol is about the messages that must be exchanged to issue verifiable c
...
-A registry of attachment formats is provided in this RFC within the message type sections. A sub-section should be added for each attachment format type (and optionally, each version). Updates to the attachment type formats does **NOT** impact the versioning of the Issue Credential protocol. Formats are flexibly defined. For example, the first definitions are for `hlindy-zkp-v1.0`, assuming that all Hyperledger Indy implementations and ledgers will use a common format. However, if a specific instance of Indy uses a different format, another format value can be documented as a new registry entry.
+A registry of attachment formats is provided in this RFC within the message type sections. A sub-section should be added for each attachment format type (and optionally, each version). Updates to the attachment type formats does **NOT** impact the versioning of the Issue Credential protocol. Formats are flexibly defined. For example, the first definitions are for `hlind/`, assuming that all Hyperledger Indy implementations and ledgers will use a common format. However, if a specific instance of Indy uses a different format, another format value can be documented as a new registry entry.
```
TimoGlastra, line 79:
```suggestion
A registry of attachment formats is provided in this RFC within the message type sections. A sub-section should be added for each attachment format type (and optionally, each version). Updates to the attachment type formats does **NOT** impact the versioning of the Issue Credential protocol. Formats are flexibly defined. For example, the first definitions are for `hlindy`, assuming that all Hyperledger Indy implementations and ledgers will use a common format. However, if a specific instance of Indy uses a different format, another format value can be documented as a new registry entry.
```

TimoGlastra: Maybe also add a sentence that indicates a format cannot use different identifiers for different messages?

TimoGlastra (approved):
I agree with this change.

swcurran (approved):
LGTM me -- good clarifications and alignment.

swcurran:
Good to have @dhh1128 review this as well.

features/0592-indy-attachments/README.md
```diff
@@ -20,51 +20,107 @@ Allows Indy-style credentials to be used with credential-related protocols that
...
 The same structure can be embedded at `/offers~attach/data/base64` in an [`offer-credential` message](../0453-issue-credential-v2/README.md#offer-credential). 
 
 ### cred request format
 
 This format is used to formally request a credential. It differs from the credential abstract above in that it contains a cryptographic commitment to a link secret; an issuer can therefore use it to bind a concrete instance of an issued credential to the appropriate holder. (In contrast, the credential abstract describes the schema and cred def, but not enough information to actually issue to a specific holder.)
 
-The identifier for this format is: `hlindy/cred-req@v2.0` It is a base64-encoded version of the data returned from [indy_prover_create_credential_req()](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L658).
+The identifier for this format is `hlindy@v2.0` in the context of a Request Credential message. It is a base64-encoded version of the data returned from [indy_prover_create_credential_req()](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L658).
```
dhh1128, line 123: This phrase, "in the context of a Request Credential message", manifests the difference in perspectives between this PR and the last update to 0592. I expected the identifier to be for the specific attachment format: `hlindy/cred-req@v2` rather than for the ecosystem. Thus "in the context" would be unnecessary.

I think attachment formats can be used in multiple messages (including messages in protocols that haven't been written yet), so we should not hard-contextualize them by a specific message in a specific protocol.

features/0592-indy-attachments/README.md
```diff
@@ -102,47 +158,48 @@ A complete [`request-credential` message from the Issue Credential protocol 2.0]
 
 ### credential format
 
-A concrete, issued Indy credential may be transmitted over many protocols, but is specifically expected as the final message in [Issuance Protocol 2.0](../0453-issue-credential-v2/README.md). The identifier for its format is `hlindy/cred@v2.0`, and its media (MIME) type is `application/indy-cl-anoncred2`.
+A concrete, issued Indy credential may be transmitted over many protocols, but is specifically expected as the final message in [Issuance Protocol 2.0](../0453-issue-credential-v2/README.md). The identifier for its format is `hlindy@v2.0` in the context of an Issue Credential message.
```
dhh1128, line 161: Indy credentials definitely need their own format that distinguishes them from the format of a credential request or proposal.

features/0592-indy-attachments/README.md
```diff
@@ -154,7 +211,7 @@ Here is a sample proof request that embodies the following: "Using a government-
 
 This is the format of an Indy-style ZKP. It plays the same role as a W3C-style verifiable presentation (VP) and can be [mapped to one](https://docs.google.com/document/d/1ntLZGMah8iJ_TWQdbrNNW9OVwPbIWkkCMiid7Be1PrA/edit#heading=h.vw0mboesl528).
 
-This is the format emitted by libindy's [indy_prover_create_proof()](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L1404)) function. A proof that responds to the [previous proof request sample](#proof-request-format) looks like this:
+The identifier for this format is `hlindy@v2.0` in the context of a Presentation message. It is the base64-encoded version of the data emitted by libindy's [indy_prover_create_proof()](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L1404)) function. A (non-base64-encoded) proof that responds to the [previous proof request sample](#proof-request-format) looks like this:
```
dhh1128, line 214: indy_prover_create_proof already emits data in base64-encoded data. The verbiage here implies that the encoding happens after the fact. I suggest rewording to "It is the data emitted by... function (which emits the data base64-encoded). A (non-base64-encoded)...

dhh1128 (changes requested):
See inline comments. I didn't tag every place where my comments apply -- only the first few.

features/0453-issue-credential-v2/README.md
```diff
@@ -146,8 +146,8 @@ Description of attributes:
 
 Credential Format | Format Value | Link to Attachment Format | Comment |
 --- | --- | --- | --- | 
-DIF Credential Manifest | `dif/credential-manifest@v1.0` | [`propose-credential` attachment format](../0511-dif-cred-manifest-attach/README.md#propose-credential-attachment-format) | 
-Hyperledger Indy Credential Abstract | `hlindy/cred-abstract@v2.0` | [`cred abstract` format](../0592-indy-attachments/README.md#cred-abstract-format)|
+DIF Credential Manifest | `dif/cred-manifest@v1.0` | [`propose-credential` attachment format](../0511-dif-cred-manifest-attach/README.md#propose-credential-attachment-format) |
```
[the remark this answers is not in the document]
llorllale: I agree. I don't know what this change is trying to fix.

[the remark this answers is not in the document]
llorllale: The identifier in `Format Value` is meant to match the one in the respective attachment format RFC. Changing this here to `dif/cred-manifest@v1.0` without an equal change in RFC0510 creates misalignment.

But still - what are we trying to solve with this change?

llorllale (dismissed):
Must explain the change in format identifiers.

features/0453-issue-credential-v2/README.md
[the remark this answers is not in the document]
sklump: I originally implemented the standard with its hlindy-zkp@v1.0 everywhere [https://github.com/hyperledger/aries-rfcs/tree/de0092a41b7d54c4fe58a69e0dba24ae8bec9360/features/0454-present-proof-v2]. @TimoGlastra  branched off that implementation for DIF credentials. Then the practice of the standard changed and I missed it - so I would ask the same question against that change in the first place: it requires a redesign of the format abstraction which encapsulates a bunch of logic, and then that will make Timo's merge into their DIF implementation all the more laborious.

PR 607 suggests that a format identifier should identify a format (as it did in the first place), not a protocol and attachment type - by construction it is already embedded in the protocol and message, allowing for only one attachment type per format. However, if the consensus is otherwise, I can go back and change things accordingly.

In any case, the effort level is likely to be man-days at most, not weeks or months.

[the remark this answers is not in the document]
TimoGlastra: @llorllale @dhh1128 What we're trying to solve here is what I described in this issue: #599. We didn't get to it during the WG call. 

The specific format identifiers feels like it adds unnecessary complexity. However, I think Daniel raises a good point that the formats could be used outside of the issue credential / present proof protocol. I had not thought about it that way.

features/0592-indy-attachments/README.md
sklump: So those future things can define their own format markers. It's the attachment, not the format specifier that carries data. The standard defines what to expect in the attachment by format specifier. For example, in the context of a credential request, for indy one ought to expect an (encoded) indy proof-request. I thought that was why the standard defined a registry per-message in the first place.

But of course I'm happy to describe (aries-rfc) and implement (aca-py) whatever the consensus is here.

sklump:
I will go back to the text and propose changes here in a couple of hours.

features/0592-indy-attachments/README.md
sklump: indy_prover_create_proof emits JSON-encoded, not base64-encoded data.

sklump:
How about now? Is this OK? I want it to work for everyone. Thanks.

features/0592-indy-attachments/README.md
dhh1128: Sorry, I was confusing this with a different function that emits base64-encoded data. You are correct and I am not; the data is JSON as emitted from the function.

The rationale for special-casing JSON encoding for an attachment was to make data easier to read for developers in the body of a message. If we're not using *embedded* attachments, I think base64 should be used, as this argument is weaker. Base64 is also required if the attachment will be signed (to avoid the JSON canonicalization problem). And DIDComm v2 always uses base64. None of these observations means that you are incorrect, though; any form of attachment ought to work.

features/0592-indy-attachments/README.md
dhh1128: I don't feel like I understand your reasoning here. As you're proposing to use it, it seems to me that "hlindy@v2.0" is not a format; it is a collection of formats that are expected to work together. Evidence for this assertion is that you're using at as an identifier for data structures with significantly different structures from one another; the only thing they have in common is that they form a single coherent ecosystem (right now). What I am asking for is a format identifier that would allow a new snapshot of the ecosystem where the 2.0 version of the cred offer could be combined with the 3.0 version of the cred request. In other words, I want "2.0" to modify something "proof-req" or "cred-offer", not just "hlindy".

This is a pretty significant difference of worldview; I don't think we can merge this PR until we align one way or another.

dhh1128:
This looks quite good now, and I think I'm aligned with the intent. However, I think I'm seeing a couple typos. I'll review more carefully in an hour or so.

features/0453-issue-credential-v2/README.md
```diff
@@ -147,7 +147,7 @@ Description of attributes:
 Credential Format | Format Value | Link to Attachment Format | Comment |
 --- | --- | --- | --- | 
 DIF Credential Manifest | `dif/credential-manifest@v1.0` | [`propose-credential` attachment format](../0511-dif-cred-manifest-attach/README.md#propose-credential-attachment-format) | 
-Hyperledger Indy Credential Abstract | `hlindy/cred-abstract@v2.0` | [`cred abstract` format](../0592-indy-attachments/README.md#cred-abstract-format)|
+Hyperledger Indy Credential Abstract | `hlindy@cred-filter@v2.0` | [`cred filter` format](../0592-indy-attachments/README.md#cred-filter-format)|
```
dhh1128, line 150: I think this identifier should have a / instead of a @ in the first slot: `hlindy/cred-filter@v2.0`

features/0592-indy-attachments/README.md
```diff
@@ -20,51 +20,107 @@ Allows Indy-style credentials to be used with credential-related protocols that
...
 ## Reference
 
-### cred abstract format
+### cred filter format
 
-This format is used to clarify the structure and semantics (but not the concrete data values) of a potential credential -- both in proposals sent from potential holder to issuer, and in offers sent from issuer to potential holder.
+The potential holder uses this format to propose criteria for a potential credential for the issuer to offer.
 
-The identifier for this format is: `hlindy/cred-abstract@v2.0` It is a base64-encoded version of the data returned from [`indy_issuer_create_credential_offer()`](https://github.com/hyperledger/indy-sdk/blob/57dcdae74164d1c7aa06f2cccecaae121cefac25/libindy/src/api/anoncreds.rs#L280).
+The identifier for this format is `hlindy/cred-filter@v2.0`. It is a base64-encoded version of the data structure specifying zero or more criteria from the following (non-base64-encoded) structure:
```
dhh1128, line 27: It seems odd to me that we'd allow a message to propose criteria, but allow zero criteria. Shouldn't we say "one or more criteria" -- on the theory that if they don't have any criteria to propose, they simply don't propose anything? I guess it depends on whether we think it makes sense for a holder to say, "I propose that you give me a credential, but I'm not going to tell you what schema or cred def or anything else; you can just give me whatever credential you feel like."

This comment is not intended to request a change to the text, so much as just to ask a question. I'm okay merging the PR without it being addressed.

dhh1128 (approved):
See my comment about one minor typo. Otherwise, LGTM.

features/0592-indy-attachments/README.md
sklump: Zero or more is just what I implemented - otherwise it needs a check in marshmallow de/serialization to ensure that at least one of these mandatory fields are in place.

dhh1128: Okay, thanks for the clarification.

features/0453-issue-credential-v2/README.md
sklump: oh, you're right

sklump: Fixed in PR 618.

PR #615, opened 2021-03-15
STATUS CHANGE: Request to move RFC-0211-route-coordination to approved

swcurran:
Request for the community to vote on changing the status of the RFC to approved.

Note that the RFC still contains some unanswered questions, which should be answerable by the developers of the implementation
done in ACA-Py.  Perhaps @dbluhm might do a PR to my PR to address/remove those questions.


dbluhm:
Opened up a PR addressing those questions with the changes from this PR included so we can just worry about the one PR; happy to open #617 as a PR against your branch if that is preferred, @swcurran.

swcurran:
Agree on the new changes.  Now to get the CI to pass and then we can merge and approve...

PR #617, opened 2021-03-16
Coordinate mediation protocol update

dbluhm:
Move to Accepted status. Addressed unresolved questions and fleshed out examples. Renamed protocol folder to match new protocol name. Left redirect notice at old URL.

It is proposed that this supersedes #615.

Thanks for the prompting the clean up on this, @swcurran.


dbluhm:
Circle CI did not like my redirect notice. Am I missing a better way to handle something like a folder name change without breaking links out in the world? Can we instruct the CI to skip a file?

dbluhm:
As per discussion during WG call, terms will be removed from RFC pending a future more thorough discussion.

llorllale:
Agreed on the B call today:

* we will pull the `*term` fields from the RFC because they are not in use + additional discussion around how to properly present terms needs to be had

swcurran:
@dbluhm -- I'm prepping for this week's call.  Can you shoot me a diff of the old and new as if the file name had not changed?  Or perhaps revert the name and status change for now so that we can discuss the content changes, and then once that is merged, do a name and status change?

Also, make the `term` changes George references above?

On the other hand -- looks like perhaps you discussed this last week (sorry -- missed the call)

dbluhm:
This was discussed during the last call, apologies for the delay on the fixes as discussed. I'll split content changes and the folder rename so the changes are more obvious.

dbluhm:
Reverted the name change and redirect notice for now and removed terms as discussed.

swcurran (commented):
LGTM -- let's vote on this at the next Aries WG Meeting.

TelegramSam (approved):

PR #619, opened 2021-03-22
Goal codes in credential exchange protocols

TelegramSam:
I think we've made a mistake by not including optional goal codes in the credential exchange protocols. I know this is a late add, so we may need to skip it for AIP 2.0 purposes.


swcurran (approved):
Group likes it.

dhh1128:
Approved on community B call, 24 Mar 2021

PR #625, opened 2021-03-25
RFC0334: added clarifications to unresolved questions

Baha-sk:
including clarifications in the JWE format around key types and base64URL encoding and minor typo fixes

### DIFF
--- a/features/0334-jwe-envelope/README.md
+++ b/features/0334-jwe-envelope/README.md
@@ -14,7 +14,7 @@ Agents need to use a common set of algorithms when exchanging and persisting dat
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


Baha-sk:
@swcurran I meant to amend my previous pr with these updates, but no worries, a new PR makes changes to this RFC clearer, especially for answers around unresolved questions (which I believe is what we're aiming for in aries-framework-go and hoping the community agrees). 

I know there's a been a long discussion about KIDs in the past, but for the sake of moving forward with JWE and finalize implementation of this RFC, each Aries implementation can have their own definition of kid format (it could be a did key or any string value, we chose to use the JWK thumbprint in Aries-framework-go to be consistent with the public key the kid represents). 

tagging @kdenhartog @dhh1128 @llorllale @troyronda @andrewwhitehead  as well, I sincerely hope this pr doesn't translate into another argument debacle around kids

TelegramSam (approved):

PR #626, opened 2021-04-06
chore: Clarify the role of recipientKeys and removed the `signed` wording linked to attachments

### LINKED ISSUE #520: DIDComm RecipientKeys are they Verification or Encryption keys
As seen in the didComm example: https://github.com/hyperledger/aries-rfcs/blob/master/features/0067-didcomm-diddoc-conventions/README.md#service-conventions

`recipientKeys` are shown as verification keys. 

The DIDComm RFC does not mention what's the intended use of keys in `RecipientKeys`, are they used for encryption only, signature verification only or both?

The official definition of `recipientKeys` from [Aries-RFC 067](https://github.com/hyperledger/aries-rfcs/tree/master/features/0067-didcomm-diddoc-conventions) is the following:
>recipientKeys : This is an array of [did key references](https://w3c-ccg.github.io/did-spec/#public-keys) used to denote the default recipients of an endpoint. (note-1)
>
>> 1.  Notes
>>
>>The keys featured in this array must resolve to keys of the same type, for example a mix Ed25519VerificationKey2018 or RsaVerificationKey2018 in the same array is invalid.

I personally prefer to separate encryption and verification keys to increase security resilience, but the spec doesn't mention this anywhere. This issue is created to discuss a more detailed definition of the purpose of keys contained in `recipientKeys` (and probably `routingKeys` as well).

This issue is raised due to a related discussion in [did-core](https://github.com/w3c/did-core/issues/185#issuecomment-668786985)

Baha-sk:
Since we are currently using authenticated encryption (JWE envelopes) the
inviter/invitee relationship is trusted and doesn't require signing the attachment.

This makes it clear that `recipientKeys` and `routingKeys` are used for authenticated encryption (JWE) operations only.

Closes #520

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -173,7 +173,8 @@ The _requester_ may provision a new DID according to the DID method spec. For a 
   * It SHOULD include the `thid` property. In doing so, implementations MUST set its value to that of `@id` on the same `request` message. In other words, the values of `@id` and `~thread.thid` MUST be equal.
 * The `label` attribute provides a suggested label for the DID being exchanged. This allows the user to tell multiple exchange requests apart. This is not a trusted attribute.
 * The `did` attribute MUST be included. It indicates the DID being exchanged.
-* The `did_doc~attach` contains the DID Doc associated with the DID as a [signed attachment](../../concepts/0017-attachments/README.md). If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc~attach` attribute is optional.
+* The `did_doc~attach` contains the DID Doc associated with the DID.
+  * If the DID method for the presented DID is not a peer method and the DID Doc is resolvable on a ledger, the `did_doc~attach` attribute is optional.
 
 #### Correlating requests to invitations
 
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
This does not cover the original reason for signing the DID document attachment, which is to establish that the party sending the connection response is the same one that sent the invitation. That is probably covered by the fact that they can decrypt the connection request, however.

TelegramSam (approved):

Baha-sk:
> This does not cover the original reason for signing the DID document attachment, which is to establish that the party sending the connection response is the same one that sent the invitation. That is probably covered by the fact that they can decrypt the connection request, however.

Being able to sign/verify the doc attachment is one thing and encrypting/decrypting the request envelope is another.
Both signing and encryption keys should come from the DID doc, where:
- Signing key should be stored in the `authentication` verification relationship.
- Encryption key should/could (I have not explicitly mentioned this in the pr, but I believe it's intended for envelope encryptions) be stored in the `KeyAgreement` verification relationship.

Keeping the keys separate has the advantage of using independent keys, although you can convert an Ed25519 into an X25519 key and use that for KeyAgreement, you can use any unrelated X25519 or even another key type (eg: ECDSA P-256).

features/0023-did-exchange/README.md
```diff
@@ -316,15 +317,17 @@ The exchange response message is used to complete the exchange. This message is
 }
 ```
 
-The key used in the signed attachment must be verified against the invitation's `recipientKeys` for continuity.
+The key used in the signed attachment must be verified against the responder’s DID document’s authentication relationships for continuity.
```
swcurran, line 320: The attachment is no longer being signed, correct?  So this needs to be changed?

swcurran (changes requested):
Anything else need to be changed since the attachment is no longer going to be signed?

features/0023-did-exchange/README.md
Baha-sk: Text updated, let me know if it makes more sense now.

Baha-sk:
the last 2 commit updates above are commit comment update, you can ignore them

swcurran (approved):

PR #634, opened 2021-04-12
Align RFCs with DIDComm profiles

troyronda:
- Adds/adapts the profiles description originally from @dhh1128 in https://github.com/decentralized-identity/didcomm-messaging/pull/177 to RFC 44.
- Updates the media type examples in RFC 67, 434 and 587 to align with profiles.

### DIFF
--- a/features/0044-didcomm-file-and-mime-types/README.md
+++ b/features/0044-didcomm-file-and-mime-types/README.md
@@ -27,7 +27,12 @@ support a million other uses.
 
 We need to define how files and attachments can contain DIDComm messages, and what the
 semantics of processing such files will be.
+
 ## Tutorial
+
+### Media Types
+
+ Media types are based on the conventions of [RFC6838](https://tools.ietf.org/html/rfc6838). Similar to [RFC7515](https://tools.ietf.org/html/rfc7515#section-4.1.9), the `application/` prefix MAY be omitted and the recipient MUST treat media types not containing `/` as having the `application/` prefix present.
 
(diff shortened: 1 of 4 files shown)


swcurran (approved):
Nice.  The text looking good.  Would like someone else to make sure all the codes are right -- they look like they are.

troyronda:
> The text looking good

Yes - thanks @dhh1128 !

Baha-sk (approved):
This means `application/didcomm-enc-env` media type should not be referenced anywhere.

and that :

`didcomm/aip2;env=rfc587` is equivalent to `didcomm/v2`


LGTM

dhh1128:
>didcomm/aip2;env=rfc587 is equivalent to didcomm/v2

I don't agree with this summary. They are equivalent in the encryption envelope dimension, but are not equivalent in headers vs. decorators, in zero-round-trip semantics, in DID rotation, etc.

Baha-sk:
> > didcomm/aip2;env=rfc587 is equivalent to didcomm/v2
> 
> I don't agree with this summary. They are equivalent in the encryption envelope dimension, but are not equivalent in headers vs. decorators, in zero-round-trip semantics, in DID rotation, etc. Was that what you meant, @Baha-sk ?

Agree.. I meant they're equivalent in terms of envelope format/JWE packer selection only @dhh1128

dhh1128:
Ah, okay. Then I'm confident in our alignment again. Thanks.

dhh1128 (approved):
LGTM

PR #635, opened 2021-04-13
add `kid` and `skid` headers references section

Baha-sk:
To clarify how `kid` and `skid` headers are dereferenced in DIDComm V2 envelopes, this change adds a dedicated section about them.
It also includes a section about applications requiring to protect the `skid` protected header.


swcurran (approved):
Per discussion at Aries WG Call 2021.04.28, this can be merged.

PR #639, opened 2021-04-16
clarify URL shortening

b1conrad:
Signed-off-by: Bruce Conrad <bruce_conrad@byu.edu>

This change loosens the requirements for URL shortening to reflect actual usage.

1. The GUID used to track out-of-band messages may be embedded in the short URL path, rather than as a query string `id` parameter value.
2. The sender may respond to the short URL by redirecting to the long URL.


swcurran (commented):
Looks great to me -- much appreciated!  I wrote the original and really did not know the best practices in this area. 

I won't mark this as approved until those with more knowledge in this area review it -- e.g. @TelegramSam @dhh1128

dhh1128 (commented):
I am generally okay with this. However, I want to be sure we can recognize the GUID of the OOB message when it appears as something other than the value of an `id` parameter. Are we sure that is the case? (I don't believe message IDs are required to be GUIDs -- only recommended to be something unique -- so I don't believe we can detect which segment of a path is a message ID just by looking for something that matches a GUID regex.)

swcurran (approved):
Discussed on Aries WG Call 2021.04.28

swcurran:
@b1conrad -- can you please update the branch?  We can't merge until that's done.  Thanks!

TelegramSam:
> I am generally okay with this. However, I want to be sure we can recognize the GUID of the OOB message when it appears as something other than the value of an `id` parameter. Are we sure that is the case? (I don't believe message IDs are required to be GUIDs -- only recommended to be something unique -- so I don't believe we can detect which segment of a path is a message ID just by looking for something that matches a GUID regex.)

The `id` here is not the message ID, just a unique id for message retrieval. In this case, I'm not even sure we need to recognize the attribute, or that it needs to be a GUID.  Or am I missing something?

dhh1128:
Okay, I see now how this id is opaque. Concern withdrawn. Full speed ahead.

PR #643, opened 2021-04-26
Add ECDH-ES for Anoncrypt in Envelope v2

troyronda:

### DIFF
--- a/features/0587-encryption-envelope-v2/README.md
+++ b/features/0587-encryption-envelope-v2/README.md
@@ -19,7 +19,13 @@ The ciphersuites defined in this RFC are a subset of the definitions in [Aries R
 
 ## Encryption Algorithms
 
-[ECDH-1PU draft 03](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) defines the JWE structure. The following sections summarize the supported algorithms.
+DIDComm defines both the concept of authenticated sender encryption (aka `Authcrypt`) and anonymous sender encryption (aka `Anoncrypt`).
+In general, Aries RFCs and protocols use `Authcrypt` to exchange messages.
+In some limited scenarios (e.g., mediator and relays), an Aries RFC or protocol may define usage of `Anoncrypt`.
+
+[ECDH-1PU draft 03](https://tools.ietf.org/html/draft-madden-jose-ecdh-1pu-03) defines the JWE structure for `Authcrypt`.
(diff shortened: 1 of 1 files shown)


Baha-sk (approved):

andrewwhitehead:
LGTM

andrewwhitehead (approved):

troyronda:
Related: https://github.com/decentralized-identity/didcomm-messaging/pull/190

swcurran:
I'm pretty much a newbie on this trying to understand the ramifications, so my opinion counts for very little. I lean to what @dhh1128 has been saying and that we should stick to only using ECDH-1PU (authcrypt) for DIDComm. Both sides of the messaging should know with whom they are interacting.

llorllale (approved):

llorllale:
Anoncrypt particularly shines in routing use cases.

dhh1128:
I don't have serious heartburn about supporting ECDH-ES in DIDComm. My concerns were about implementing the Presentation Exchange protocol in anoncrypt mode rather than the existence of anonycrypt in general.

Of course, if we leave ECDH-ES out of DIDComm, we can prevent PEx from making what I consider a mistake...

swcurran:
Should there be guidance on when to use authcrypt vs. anoncrypt?  For example, authcrypt must be used for messages between sender and recipient, while anoncrypt may be used for messages to and from mediators and relays?

troyronda:
@swcurran I think guidance is helpful.

How's this:

> DIDComm defines both the concept of authenticated sender encryption (aka `Authcrypt`) and anonymous sender encryption (aka `Anoncrypt`).
In general, Aries RFCs and protocols use `Authcrypt` to exchange messages.
In some limited scenarios (e.g., mediator and relays), an Aries RFC or protocol may define usage of `Anoncrypt`.

swcurran:
I'd be happy with that.

troyronda:
@swcurran added the guidance text: https://github.com/hyperledger/aries-rfcs/compare/f39c12fb0b7dc43b2fa283718b9140c033bad44f..02c6b4efc6169ebd5e91dbd2085a8bef3c4bc031

swcurran (approved):

swcurran:
Sounds like we've arrived at a happy place.  I'll wait a bit more in case someone else wants to weigh in on the last changes and then will merge this.  Speak now...  :-)

Issue #632, opened 2021-04-09
Transitioning AIP 2.0 RFCs to the "ACCEPTED" status


swcurran:
A large number of AIP RFCs have had not had their status updated to ACCEPTED.  The goal of this issue is to track those transitions and propose a process that allows us to proceed with AIP 2.0 finalization without getting bogged down with the bureaucratic processes.

The process I propose is the following:

1. In the next Aries WG meeting, we walk through all of the "not yet Accepted" and take a "no discussion" vote on each RFC listed below.
2.  For each unanimous `yes` vote, the repo maintainers are given the authority to add and merge a PR to update the status of the RFC.
3. For each vote that has one or more `no` votes, a discussion will be held after going through the rest of the list, and if possible another vote is held. The discussion should be brief so as to determine if a longer discussion or an update to the RFC is needed before another  
4.  If there is a majority `yes` vote following the discussion, the repo maintainers are given the authority to add and merge a PR to update the status of the RFC.
5. For the remaining RFCs, the group will decide how to address them -- keep the RFC in AIP 2.0, request a volunteer to work the "difficult" issues, etc.

The goal will be to get through the full list.  Participants are asked to review at least the RFCs that they think may be insufficient as they currently stand.

Here is the list of RFCs to be reviewed:

- [x] Concept | [0519-goal-codes](https://github.com/hyperledger/aries-rfcs/tree/b982c24b9083dd5dddff6343dbf534cd1cfe36a6/concepts/0519-goal-codes) | :new:
- [x] Feature | [0015-acks](https://github.com/hyperledger/aries-rfcs/tree/b982c24b9083dd5dddff6343dbf534cd1cfe36a6/features/0015-acks) | AIP V1.0, Updated
- [x] Feature | [0023-did-exchange](https://github.com/hyperledger/aries-rfcs/tree/b982c24b9083dd5dddff6343dbf534cd1cfe36a6/features/0023-did-exchange) | :new:
- [x] Features | [0044-didcomm-file-and-mime-types](https://github.com/hyperledger/aries-rfcs/tree/b982c24b9083dd5dddff6343dbf534cd1cfe36a6/features/0044-didcomm-file-and-mime-types) | :new:
- [x] Feature | [0360-use-did-key](https://github.com/hyperledger/aries-rfcs/tree/b982c24b9083dd5dddff6343dbf534cd1cfe36a6/features/0360-use-did-key) | :new:
[...]

PR #644, opened 2021-04-28
simplify please ack decorator

TimoGlastra:
Simplify please ack decorator for inclusion in AIP 2.0 (#632). I simplified it to the most basic version where you can only indicate whether you want an ack on RECEIPT or OUTCOME. 

I'm not sure if I removed too much, but I think this basic variant will be quite easy to implement for agents, while still providing a lot of value.

@dhh1128 as the original author of the RFC, what do you think?


swcurran (approved):
LGTM

TimoGlastra:
From WG call:

- Make `on` an array
- Describe generic behaviour for `OUTCOME`
	- when interaction is done / outcome reached -> ack sent
	- Allow protocols to overwrite `OUTCOME` behaviour

dhh1128 (approved):
I am okay with the simplifications in this PR. I am happy that we retained a reference to the earlier version of the RFC so the advanced features can still be discovered for possible exploration in a future RFC. Thanks.

swcurran (changes requested):
Preventing merge until array update made.

swcurran (approved):

PR #645, opened 2021-04-28
fix incorrect thid wording in did exchange RFC

TimoGlastra:
As discussed on the WG call a few weeks ago. The did exchange protocol required the `thid` property to be equal to the message ID. This is an unnecessary restriction


swcurran (approved):
LGTM

PR #646, opened 2021-04-28
add 0646: W3C Credential Exchange using BBS+ Signatures

TimoGlastra:
Initial draft PR. Still needs some things resolved and would be best if we merge it after the issues discussed at IIW are resolved. This mainly involves updating the BBS+ LD Proofs spec and defining how private holder binding can be achieved.


swcurran:
@TimoGlastra -- to fix CI/CD add a "Since" data and change tags to be "feature".

swcurran (approved):
Per discussion at Aries WG Call - 2021.04.28

PR #650, opened 2021-05-04
Changed public DID language, removed GUID requirement in URL

TelegramSam:
As discussed in our WG meeting last week.


swcurran (approved):
Discussed on Aries WG Call 2021.05.05.

PR #661, opened 2021-05-12
Remove extra URI types for links

swcurran:


kdenhartog (approved):
It makes sense to add these in as people want to use them rather than guess which ones might be used in order to exemplify how the links property can support any URL

Issue #664, opened 2021-05-19
Update Example from RFC 0360


Matt-Spence:
The example on transforming an Ed25519 public key to a did:key reference seems to use the wrong multicodec identifier for an ed25519 public key. [The W3C spec](https://w3c-ccg.github.io/did-method-key/) says to prepend the raw key bytes with the multicodec identifier for the key type, which according to [this](https://github.com/multiformats/multicodec/blob/master/table.csv) (line 85) is 0xed for Ed25519-pub. The example says "Multicodec: Prefix with the algorithm type. In this case, ED01 as the key type is ed25519.". The RFC should be updated to use the correct multicodec identifier.

swcurran:
Resolved with #666

PR #665, opened 2021-05-19
Clarify text for did:key encoding - incorrect prefix per #664

swcurran:


swcurran:
Clarify text for did:key encoding - incorrect prefix per #664

Issue #671, opened 2021-05-20
Goal Codes in 0023 DID Exchange


domwoe:
Goal codes are currently not mentioned in RFC 0023, although they could be useful from my perspective in the scenario when a DID Exchange is not initiated by a particular out of band invitation, but by an ad hoc request to a public DID. The receiver could use this goal code to decide if the request should be accepted and which presentation request to send.

Should we allow  goal codes in the DID exchange request?

swcurran:
Goal codes are used in OOB which is one way to trigger DID Exchange 0023, so that path is covered. The implicit invitation route to start a DID Exchange is possible -- is that where you would like to see goal codes used?

domwoe:
My mental model is as follows:

If an entity creates a OOB invitation it wants to signal an intention to the receiver of the invitation. Therefore a goal code can be used and this is already in place.

However, if an entity has a public DID, the roles turn around. The entity sending the request to the public DIDs might want to signal an intention the receiver of the request and this is currently not covered.

swcurran:
Can you do a quick PR for that?  Really nice if we can get it into AIP 2.0, which we hope to close next Wednesday.

domwoe:
I'll try.

domwoe:
Included with #672

PR #672, opened 2021-05-20
add goal code to didx request

domwoe:
In response to #671 
I'm not confident to provide a [Message Catalogue Section](https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md#goal_code) as of now.

Signed-off-by: Woerner Dominic (RBCH/PJ-IOT) <dominic.woerner2@ch.bosch.com>

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -147,6 +147,8 @@ The _requester_ may provision a new DID according to the DID method spec. For a 
       "pthid": "<id of invitation>"
   },
   "label": "Bob",
+  "goal_code": "aries.rel.build",
+  "goal": "To create a relationship",
   "did": "B.did@B:A",
   "did_doc~attach": {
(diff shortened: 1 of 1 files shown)


swcurran:
@TelegramSam @dhh1128 @troyronda @andrewwhitehead @llorllale 

I think this is useful and should be added before we lock in AIP 2.0.  It adds optional fields, so it technically a minor update, although I think there can be flexibility on that given we are before locking in AIP 2.0.

What do you think?   

The only issue for me is what @domwoe mentioned above about not providing a Message Catalogue section as we don't really know what this will be used for -- just that it should be available.

Please respond earlier than later.

andrewwhitehead:
There's a couple typos around the word 'particular', but I think it's a good addition.

dhh1128:
+1 from me

dhh1128 (approved):
LGTM

TelegramSam (approved):

PR #679, opened 2021-06-11
fix: bbs rfc and osma link

TimoGlastra:


swcurran (approved):
Has OSMA become Aries-MAX yet?  It seems like the right name...

TimoGlastra:
Ah fixed the link but forgot the name. Should be MAX

PR #683, opened 2021-06-23
update proposed rfcs included in aip to accepted

TimoGlastra:
Saw that RFC 0510 and 0646 were still in proposed state, even though they are included in AIP2.0

Not sure what to do with the links in the AIP RFC, as it seems they all use the same commit. Can I just update the link to the newer commit for these two RFCs? No change has been made besides the `proposed` -> `accepted`


swcurran (commented):
Looks good, but we should discuss this merge on the Aries WG B Call today @TelegramSam 

Once the change is made, we can do a PR to update the commits for the AIP Links. Such changes are permitted when the change is just to clarify the linked RFCs in an AIP Version.

dhh1128 (approved):

TimoGlastra:
OOB is completely missing AIP2.0: https://github.com/hyperledger/aries-rfcs/blob/master/features/0434-outofband/README.md

swcurran:
@TimoGlastra -- could you please add OOB to the PR.  I think it is more than ready.  I just missed it...thanks for catching it.

TimoGlastra:
Done. Also updated the static peer dids to accepted as it's part of AIP2.0.

swcurran (approved):
LGTM -- @TelegramSam, please discuss on the Wednesday call and merge if acceptable.

dhh1128 (approved):

dhh1128:
Also LGTM.

PR #685, opened 2021-06-24
Pickup Protocol v2

TelegramSam:


features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+## Reference
+
+Each message sent should use the ~transport decorator as follows. This has been ommitted from the examples for brevity.
+
+```json=
+"~transport": {
+    turn_route": "thread"
+}
```
TimoGlastra, line 46: return_route thread requires the `return_route_thread` parameter to be present. 
```suggestion
"~transport": {
    "return_route": "thread",
    "return_route_thread": "<thread-uuid>"
}
```

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+    "newest_time": "2019-05-01 12:00:00Z",
+    "oldest_time": "2019-05-01 12:00:01Z",
+    "total_size": 8096,
+    "live_delivery": false
+}
+```
+
+`message_count` is the only required attribute. The others may be present if offered by the _message_holder_.
+
+`duration waited` is in seconds, and is the longest  delay of any message in the queue.
```
TimoGlastra, line 84:
```suggestion
`duration_waited` is in seconds, and is the longest  delay of any message in the queue.
```

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+`total_size` is in bytes.
+
+If a `recipient_key` was specified in the status-request message, the matching value should be specified in the `recipient_key` attribute of the status message.
```
TimoGlastra, line 88:
```suggestion
If a `recipient_key` was specified in the status-request message, the matching value MUST be specified in the `recipient_key` attribute of the status message.
```

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+}
+```
+
+## Prior art
+
+Version 1.0 of this protocol served as the main inspiration for this version. Version 1.0 suffered from not being very explicit, and an incomplete model of message delivery signaling.
+
+## Unresolved questions
+
+- An alternative to deriving a message ID is to wrap each message in a delivery wrapper. This would enable the mediator to include a mediator managed id and metadata along with the message itself, but carries the downside of
```
TimoGlastra, line 176: carries the downside of what....? :-)

features/0685-pickup-v2/README.md
TimoGlastra: Is there a specific reason you choose thread? I think this should be all, because the mediator can't know the thread?

TimoGlastra (commented):
Looks good. 

Curious to see the added complexity of implementing this. Don't think it will be that complex and the added explicitness is great I think.

features/0685-pickup-v2/README.md
TelegramSam: I have mixed feelings about this because of the changes in return_route coming in DIDComm V2. Specifically, using thread was designed to allow the remote agent to communicate with the mediator without invoking a flood of mediated messages. By going the explicit way we are going, including live mode, this flow isn't needed at all.

Given that thinking, I'm going to switch it to all.

features/0685-pickup-v2/README.md
TelegramSam: of suspense!

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+
+## Reference
+
+Each message sent should use the ~transport decorator as follows. This has been ommitted from the examples for brevity.
+
+```json=
+"~transport": {
+    turn_route": "all"
```
TimoGlastra, line 45:
```suggestion
    "return_route": "all"
```

TelegramSam:
Argh, I forgot momentarily that suggestions cause DCO failures.

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+    "newest_time": "2019-05-01 12:00:00Z",
+    "oldest_time": "2019-05-01 12:00:01Z",
+    "total_size": 8096,
+    "live_delivery": false
+}
+```
+
+`message_count` is the only required attribute. The others may be present if offered by the _message_holder_.
+
+`duration_waited` is in seconds, and is the longest  delay of any message in the queue.
```
dhh1128, line 84: Maybe rename to `longest_wait`? (just wordsmithing for clarity; ignore if not helpful)

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+`total_size` is in bytes.
```
dhh1128, line 86: I tend to like field names that make the units explicity. If you said "total_bytes", you wouldn't have to document that the field was in bytes...

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+Status details about pending messages
+
+```json=
+{
+    "@id": "123456781",
+    "@type": "https://didcomm.org/messagepickup/2.0/status",
+    "recipient_key": "<key for messages>",
+    "message_count": 7,
+    "duration_waited": 3600,
+    "newest_time": "2019-05-01 12:00:00Z",
```
dhh1128, line 75: Is this the time the mediator received the message? What if the mediator who's holding the message in turn used a pickup protocol to get the messages, and *knows* that the messages it's holding were actually held upstream for a week before it ever received them?

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+
+After receipt of this message, the mediator should deliver up to the limit indicated. 
+
+`recipient_key` is optional. When specified only return messages sent to that recipient key.
+
+If no messages are available to be sent, a status message MUST be sent immediately.
+
+Delivered messages MUST not be deleted until delivery is acknowledged by a `messages-received` message.
```
dhh1128, line 120: What if Alice has 1 agent that's been corrupted by malware, and 1 that is innocent? Could the corrupted agent use the pickup protocol to prevent the innocent agent from seeing any messages, by continually showing up to ask for messages and purging the queue (which the innocent agent might check far less aggressively)?

The issue I'm poking at here is whether it's okay for one agent to purge a queue when a message was sent to multiple recipients...

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+### Message Received
+After receiving messages, the Recipient sends an ack message like this:
+```json=
+{
+    "@type": "https://didcomm.org/messagepickup/2.0/messages-received",
+    "message_tag_list": ["123","456"]
+}
+```
+
+`message_tag_list` identifiers are tags of each message received, extracted from the encrypted form of each message.
```
dhh1128, line 132: I believe you mentioned using a hash here. I would like some comment about what a "tag" means, and how one is found. If it's not a hash, what is it, and how does the mediator know it?

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Upon receipt of this message, the `Mediator` knows that the messages have been received, and can remove them from the collection of queued messages with confidence. The mediator should send an updated `status` message reflecting the changes to the queue.
+
+## Live Mode
+Live mode is the practice of delivering newly arriving messages directly to a connected _Recipient_. It is disabled by default and only activated by the _Recipient_. Messages that arrive when Live Mode is off MUST be stored in the queue for retrieval as described above.
+
+Live mode must only be enabled when a persistent transport is used, such as websockets.
```
dhh1128, line 139: Can we say (I think this is implied below, but may need explicit mention) that if we're in live mode and the connection is broken, we automatically drop back to queued mode?

features/0685-pickup-v2/README.md
TelegramSam: the tag is a portion of a hash, but that detail isn't relevant to the implementation of the protocol.

TelegramSam:
@dhh1128 Updates made, Have a look at the new changes for feedback?

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+Messages are retrieved when the Recipient  sends a `delivery_request` message to the Mediator.
+
+Message delivery is confirmed with a `message_received` acknowledgement, at which point the message can be removed from the mediator's queue.
+
+When live delivery is enabled, messages that arrive when an existing connection exists are delivered over the connection immediately.
+
+## Reference
+
+Each message sent should use the ~transport decorator as follows. This has been omitted from the examples for brevity.
```
JamesKEbert, line 39: Suggestion to reference ~transport decorator RFC
```suggestion
Each message sent should use the ~transport decorator as follows, which has been adopted from [RFC 0092 transport return route](/features/0092-transport-return-route/README.md) protocol. This has been omitted from the examples for brevity.
```

features/0685-pickup-v2/README.md
JamesKEbert:
```suggestion
`longest_waited_seconds` is in seconds, and is the longest delay of any message in the queue.
```

features/0685-pickup-v2/README.md
JamesKEbert:
```suggestion
`total_bytes` represents the total size of all messages.
```

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+`message_count` is the only required attribute. The others may be present if offered by the _message_holder_.
+
+`duration_waited` is in seconds, and is the longest  delay of any message in the queue.
+
+`total_size` is in bytes.
+
+If a `recipient_key` was specified in the status-request message, the matching value MUST be specified in the `recipient_key` attribute of the status message.
```
JamesKEbert, line 86: Could be helpful to give additional details if the `recipient_key` was not specified--for instance should this not be returned if the `recipient_key` was not specified? Or would we expect an array of all keys for stored messages?

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+```json=
+{
+    "@type": "https://didcomm.org/messagepickup/2.0/delivery-request",
+    "limit": 1
+}
+```
+
+
+After receipt of this message, the mediator should deliver up to the limit indicated.
```
JamesKEbert, line 112: Is `limit` optional? I'd initially think it could be.

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+## Live Mode
+Live mode is the practice of delivering newly arriving messages directly to a connected _Recipient_. It is disabled by default and only activated by the _Recipient_. Messages that arrive when Live Mode is off MUST be stored in the queue for retrieval as described above. If live mode is active, and the connection is broken, a new inbound connection starts with live mode off.
+
+Live mode must only be enabled when a persistent transport is used, such as websockets.
+
+_Recipients_ have three modes of possible operation for message delivery with various abilities and level of development complexity:
+
+1. Never activate live mode. Poll for new messages with a `status_request` message, and retrieve them when available.
+2. Retrieve all messages from queue, and then activate Live Mode. This simplifies message processing logic in the _Recipient_.
+3. Activate Live Mode immediately upon connecting to the _Mediator_. Retrieve messages from the queue as possible. When receiving a message delivered live, the Queue may be queried for any pending messages delivered to the same key for processing.
```
JamesKEbert, line 147: Can you send a delivery request during live delivery? What is the behavior if the mediator has sent some messages to the recipient and then the recipient makes a delivery request--should the mediator track which messages have been sent but not received an ack for, or just respond with all messages and require the recipient to detect duplicate messages?

features/0685-pickup-v2/README.md
JamesKEbert: I'd agree with explicitly mentioning it so that it doesn't get overlooked

JamesKEbert (commented):
Overall I am in favor of this protocol change given the implicit behavior with mediation as of today. I've added a couple comments/questions

features/0685-pickup-v2/README.md
jakubkoci: That makes sense to me. When you omit `limit` you want to download all pending messages.

jakubkoci: In what order are the messages delivered? I assume it should be from oldest to newest, right? Actually, I'm thinking if it makes sense to have it delivered in the opposite order, not sure about that 🤔

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+```json=
+{
+    "@id": "123456781",
+    "@type": "https://didcomm.org/messagepickup/2.0/status-request",
+    "recipient_key": "<key for messages>"
+}
+```
+
+- `recipient_key` is optional. When specified only return status related to that recipient key. This allows the _Recipient_ to discover if any messages are in the queue that were sent to a specific key.
```
jakubkoci, line 60: It could be worthy referencing [0211-route-coordination](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md) where is the explanation what's the `recipient_key`, how it's created and updated. 

```suggestion
- `recipient_key` is optional. When specified only return status related to that recipient key. This allows the _Recipient_ to discover if any messages are in the queue that were sent to a specific key. You can find more details about `recipient_key` and how it's managed in [0211-route-coordination](https://github.com/hyperledger/aries-rfcs/blob/master/features/0211-route-coordination/README.md).
```

features/0685-pickup-v2/README.md
JamesKEbert: I'd agree to be able to download all pending messages. I'd assume oldest to newest, otherwise if you have a thread of messages queued they'll be delivered/processed out of order (which we have to be able to support, but I'd rather not complicate that intentionally here).

features/0685-pickup-v2/README.md
JamesKEbert: Agreed 👍

features/0685-pickup-v2/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+1. Never activate live mode. Poll for new messages with a `status_request` message, and retrieve them when available.
+2. Retrieve all messages from queue, and then activate Live Mode. This simplifies message processing logic in the _Recipient_.
+3. Activate Live Mode immediately upon connecting to the _Mediator_. Retrieve messages from the queue as possible. When receiving a message delivered live, the Queue may be queried for any pending messages delivered to the same key for processing.
+
+### Live Mode Change
+Live Mode is changed with a `live_delivery_change` message as follows:
+```json=
+{
+    "@type": "https://didcomm.org/messagepickup/2.0/live_delivery_change",
```
TimoGlastra, line 153:
```suggestion
    "@type": "https://didcomm.org/messagepickup/2.0/live-delivery-change",
```

TimoGlastra (approved):
LGTM!

PR #687, opened 2021-06-29
Add dateint representations

jsaur:
This PR is in response to a conversation on rocket chat starting with:
"Hi all, I had a question about the quintessential zero-knowledge proof example of Alice proving that she is over 18 without revealing her DOB - given that current Indy predicates only work on int32s, what is the best way to represent dates in a credential such that they work for predicate proofs?
I had one idea: at first I was trying to go with an approach of number of days since a certain epoch and being able to use common datetime functions to do the calculations, but it seemed rather complicated and there were a lot of subtle issues around timezones, leap seconds, which epoch to choose, etc. The approach I would like to propose is to apply simple modular arithmetic to map a date to an int of the form YYYYMMDD. This would allow easy conversion to int and back to date without any confusion that would result from using time libraries. The only gotchas I can see are that you couldn't represent dates before AD (which seems ok, or they could be negative which also seems ok), and that you'd need to have some error handling if somehow a credential was ever submitted with an invalid mapping eg 20201332 (13th month, 32nd day, etc) - although some error handling on credential issuing should address that. To construct the proof the verifier would just convert today's date minus 18 years to int form, and then check that the int-date DOB on the credential is < that.
What do people think? Ideally I'd like to use a standard date convention so if there's not already one, I could add a proposal to https://github.com/hyperledger/aries-rfcs/blob/master/concepts/0074-didcomm-best-practices/README.md#date-time-conventions (or create a new RFC).
(And I should specify that 32 bit unixtime wouldn't work, as we want to include dates outside of 1901-2038)"

Signed-off-by: Jacob Saur <jsaur@kiva.org>


TelegramSam:
@jsaur Can you make the Aries B call tomorrow for a discussion on this?

jsaur:
Sure, I should be able to make it

jsaur:
If it's helpful here's a handful of conversion functions in Typescript: https://github.com/kiva/protocol-common/blob/main/src/date.conversion.ts

dhh1128:
I think this is a very useful discussion. I'm glad you raised the PR. Thank you.

I have an uncertainty about the approach. It is around the importance of handling bad values wisely. The integer 20201301 (let's call it "A") is less than the integer 20210101 (let's call it "B"), but is invalid. A ZKP that doesn't reveal the integer value will validate that A < B. Because it is a ZKP, you won't have the option of testing to see whether the value in question is valid. So, do we want to ever say that <invalid value> is comparable to <valid value>?

The proposal to validate the data at issuance time is fine, if the issuer is trusted to be a provider of clean data, and is assumed not to make data mistakes. I don't know whether that's a reasonable assumption or not. So that's the essence of my uncertainty here.

The timezone issues are easily solved. All seconds-since-epoch values that I'm aware of define the epoch in terms of GMT/UTC -- so there's never any timezone ambiguity in them.

I believe the leap seconds issue is a red herring. Certainly, they do occur and they could be a problem -- but if the datatype we're representing is just a date, NOT a datetime, then the comparision should not be based on seconds anyway, but rather on something less granular. To understand what I mean, consider a scientific experiment where quantity X is measured at 3.4 units, and is compared to a standard of 3.37248567 units. Is X greater than Y?

The naive answer is yes -- but the science answer is "undefined". This is because X could only be measured with 2 digits of precision, and comparison to X requires 9 digits of precision. It is considered INCORRECT to compare these two units with anything other than the precision of the least precise of the two units.

Similarly, it is incorrect to compare a date value (which does NOT have seconds-level precision, but rather a unit of precision that is about 5 orders of magnitude less) to a seconds-since-epoch number that is far more precision.

For this reason, leap seconds should not be relevant to a comparison of "now" to a date; if you have to ask how many extra seconds might have elapsed since epoch, you're assuming too much precision.

(The way to fix this, BTW, is to encode in the credential the precision of the numeric value, and to normalize all inputs to only and exactly that value, and then to make predicates respect that precision and never work for any comparision more precise. But we don't have that solved yet.)

jsaur:
You bring up a good point around bad values, since this approach does allow for invalid dates to be represented and leaves it up to the issuer to create valid values. However I do think it's a perfectly reasonable assumption to make that a given issuer is trustworthy - since otherwise the verifier wouldn't be trusting that issuers credential values at all. Certainly there will be bad issuers out there, but then the verifier wouldn't accept credentials from them regardless of the date format.

The timezone issue was mostly around the potential confusion for converting into and out of epoch values. For example if the issuer accidentally created a Date object (in whichever language) for 00:00:00 in their local time, it could result in the day before or after when stored as a UTC seconds-since-epoch value.

I fully agree with your point about precision comparisons - that's why I wanted an approach that represented dates and not times, so there wouldn't be any ambiguity. I don't fully know how the leap seconds would work, but I didn't want a scenario where a seconds-since-epoch value for 00:00:00 on a certain date would somehow get misread as 23:59:59 for the date before due to a leap second.

I think a days-since-epoch approach would also work, but I wasn't sure how easy it would be to do these calculations. But one advantage of this approach vs days-since-epoch is that it's human readable.

dhh1128:
> However I do think it's a perfectly reasonable assumption to make that a given issuer is trustworthy

Okay, good point. +1

jsaur:
Made some updates based on suggestions from today's aries call

dhh1128 (approved):

concepts/0441-present-proof-best-practices/README.md
```diff
@@ -149,6 +149,18 @@ A presentation may include a timestamp outside of a the non-revocation interval
...
+
+#### Dates in Credentials
+
+In order for dates to be used in a predicate proof they MUST be expressed as an Int32. While unix timestamps could work for this, it has several drawbacks including: can't represent dates outside of the years 1901-2038, isn't human readable, and is overly percise in that birth time down to the second is generally not needed for an age check. To address these issues, dates attributes SHOULD be represented as integers in the form YYYYMMDD. This addresses the issues with unix timestamps (or any seconds-since-epoch system) while still allowing date values to be compared with < > operators. Note that this system won't work for any general date math (eg adding or subtracting days), but it will work for predicate proofs which just require comparisons. In order to make it clear that this format is being used, the attribute name SHOULD have the suffix `_dateint`. Since most datetime libraries don't include this format, [here](https://github.com/kiva/protocol-common/blob/main/src/date.conversion.ts) are some examples of helper functions written in typescript.
```
swcurran, line 158: Typo percise vs. precise 

Ironic, don't you think?

jsaur: Haha, good catch!

Issue #681, opened 2021-06-21
Enabling Multiple Presentations from a Presentation Request


swcurran:
From BC Gov and Kiva (@mattraffel , Jacob Saur)

We’ve seen multiple instances of the following kind of use case:

- A verifier requests a presentation (via RFC 0454 Present Proof v2) knowing that the prover may have multiple instances of a credential, and wanting to a presentations of the credentials in return, for example:
   - As part of a loan application, a request is made for a list of bank accounts held by the user.
   - As part of opening a business account, a request is made for the list of directors of the organization.

We propose adding the following items to the messages in the present-proof v2 messages:

- To the “propose-presentation” and “request-presentation” message add the item “present_multiple” set to “true” to indicate that the verifier will accept multiple presentations. If missing, default to “false”.
- To the “presentation” message add the item “last_presentation” set to “false” for all but the last presentation to be sent in response to the presentation request. If missing, default to “true”. Each presentation message sent would be part of the same thread.

A recommendation should be included that when using the “present_multiple” with value “true”, the presentation request should not include other than a simple presentation request of (usually) a single credential type. A combination of credential types, each with the prover holding multiple instances will be difficult for the wallet, worse for the user.

For the UX, we would expect the prover to provide a list of presentations that could be sent and allow the user to select the ones to be sent (or more perhaps, unselect the ones not to send). Once done, the “presentation” messages would be sent without further input from the controller.

There are multiple other ways to accomplish this that might be considered:

- In the “presentation” message, have all of the presentations to be sent as an array.
- Use goal-codes to convey the request to “present_multiple”.
- Allow the verifier to be more specific about the request for multiple, such as “send all” or “send up to 3”.
- Have the prover just send multiple presentations without indicating that more are to be sent.
- Use other data items in the messages, such as in the “presentation” message, the combination of items “presentation_number” and “total_presentations” so the verifier knows from the first response how many messages are coming.

As well, one could introduce the concept of “present_multiple” in the DIF Presentation Exchange protocol, so specifying it at the Aries protocol level. That would not work for Indy and would introduce the same dynamics as sending all the presentations in one message. Alternatively, the concept could be added at both the Aries Protocol and DIF PE levels.

rolsonquadras:
We are working on a similar problem statement where one presentation request results in multiple presentations.

Use Case: Verifier requires a Driver License (DL) and the latest Credit Report (CR) from the user to complete some transaction.

[...]

PR #689, opened 2021-06-30
New minor version of Present Proof that supports asking for and responding with multiple presentations for the same presentation request

swcurran:
Addresses #681


swcurran:
FYI -- @mattraffel @jsaur

TimoGlastra (approved):
Overall this looks good to me. I'm not 100% up to date on the use case, but wouldn't it be easier to require all proofs to be included in the same presentation message?

so `presentations~attach` will contain multiple entries for the same format.

TelegramSam (approved):

PR #692, opened 2021-07-14
DRAFT: Minor version update to Issue Credential V2 to enable issuing multiple VCs

swcurran:


swcurran:
Issues on this to be discussed before we can finalize:

1. Do we allow the array of attachments to be used to issue multiple different credentials in one message?  This is specifically precluded in the 2.0 version, and may not be valid for some VC formats.
2. With AnonCreds, is it a requirement that there be an `offer-credential` message per issuance?  The Offer is (according to the current RFC) to require the issuer to make a commitment to the credential to be issued.  I'm not sure if that is a business or cryptographic commitment, and in either case, whether it is necessary/recommended that it be used.
3. Is the repeated `request-credential` message required for each issuance for all/some VC formats?
4. Combining points 2 and 3, should we have first message after an `issue-credential` message with the `more_available` flag set to true being one or any of these, by VC format:
   1. offer-credential
   2. request-credential
   3. issue-credential
5. Should we have an explicit field in the `request-credential` message an indicator that the Holder will accept multiple credentials?

To decide on these, we need to starting with defining the per issuance sequence of messages that are necessary per VC format.

dhh1128:
This is an intriguing idea. Do we have any data about how compelling the use case is? Are we motivated to pursue it primarily because we're trying to decrease the chattiness of running the protocol multiple times? (If yes, I want to be sure we've analyzed correctly, not conflated chattiness with UX clumsiness. It would be easy to present multiple issuances as one interaction from a UX POV, without modifying the protocol at all...)

With the current data structures, I don't think there is a way to enforce the idea that multiple credentials should represent different claims (two different college degrees) versus identical claims but different formats (same degree, in cred format X v1 verses cred format X v2). And it also doesn't crisply prevent another misuse, which is batch issuance (different claims for different holders).

My gut is that we should change the message structure more deeply to accommodate different claimsets and to keep the batch abuse from happening. One way we could do that would be to start linking "claimsets" and attachments. A claimset would be the pieces of info we propose to send in one credential -- and we reference an attachment in a way that makes it clear which claimset it refers to. ALL of the claimsets in the protocol would have to be intended for the same holder, with the same terms of service and the same financial payment behaviors if they entailed a cost. That would prevent the batch abuse. But now you could explicitly understand which attachment addressed which subset of the union of all claims.

It's not clear how we deal with a partial acceptance; is this a take-it-all-or-leave-it-all offer and issuance? Or can the issuer offer 7 creds, and the holder accepts 4?

I'm not sure what consequences this has for a ZKP-oriented cred. If we're using link secrets, presumably we bind to the same one for every cred that's issued. If we're using Mattr's suggested replacement for link secrets, does the same binding apply to all -- and does that work, cryptographically?

swcurran:
@dhh1128 -- the primary motivation is to deal with UX clumsiness -- being able to tell the wallet ahead of time that there are multiple credentials of the same type to be issued -- if you want them, and letting the user decide what to do.

With the current data structures, I agree you can't force the credentials to be the same content/different format -- but that's the stated direction of the RFC. It literally says that you have to do that to meet the spec.  We discussed this on the Aries WG call, and while we think considering allowing the use of multiple attachments as a way to do multiple issuing is interesting, we don't think we're ready for single protocol, multiple issuances just using arrays of attachments in the existing message sequence.

I don't recall any past discussions about abuse in the form of different claims for different holders. Do we really think that this applies here differently from the existing protocol? 

Regards the all or none -- I added the note about the Holder using a Problem Report to say "No mas" and to stop the flow. Up to the Holder agent to figure out how to use that.

Nathan feels fairly sure that in Indy we have to have a request for each issuance, but not an offer. As such, we think the right thing is to have a "request-credential" / "issue-credential" loop.  We could make the loop VC format specific so that it could be just "issue-credential" repeated.

dhh1128:
Based on the comment stream above, I have two strong opinions:

1. This is not a minor change. It deserves a major version number increment. It is fundamentally changing the logic of the interaction, and the mental model for what it means to ask for credential(s) and receive it-or-them. In the revision, we should adjust the data structures as much as we need, to elegantly represent what we want. We should not seek to minimize changes to the message structures.

    The major reason for me taking this position is that I don't think people should be required to support a mental model and a style of data processing that expects a multi issuance -- just so they can get single issuance support.

    So in terms of concrete feedback on this PR, I'd say that I believe this should be 3.0 of the protocol, not 2.1. And as I said above, I think we should adjust the data structures more aggressively, so it's clear which attachment has which type of cred.

2. As a general rule and as a community, we are associating a protocol with a UX experience too strongly. There is nothing that says a UX workflow needs to map 1:1 onto a protocol instance. We can have one UX that encompasses multiple protocol instances (e.g., issue cred and then turn around and use it to prove something; pay for a cred and issue it; etc). We *can* adjust protocols for UX reasons -- but we shouldn't feel compelled to do so. We could also write a UX, based on the current protocol, that simply expects to represent to a user the unfolding of multiple credential issuances in parallel.

    I'm not recommending that we tell people they must implement a fancy, non-1:1 mapping for this particular problem. It may well be that it's smart to create a new protocol (or a significantly updated one) so that UIs in mobile apps only have to think about one protocol being in progress when they pop up a notification message or a question. But I am nervous about enthroning that assumption too strongly in our thinking, *in general*. We all need to be very clear that when we are writing a protocol, we are NOT writing UI. They are separable concerns, and they need to stay that way.

smithbk:
@dhh1128 Daniel, you said "It would be easy to present multiple issuances as one interaction from a UX POV, without modifying the protocol at all.".  Exactly how are you proposing this be done?

dhh1128:
>Exactly how are you proposing this be done?

I'm not *proposing*, exactly. As I said above, maybe in this case it's worth revising the protocol. Rather, I was just observing that there doesn't have to be a 1:1 correspondence between the arrival of credential offers and the UI that prompts about such an arrival.

The way it could be done is more or less how it could be done in natural human interactions. Imagine you are your wife's agent and you are going to check the mail. She gets three letters from the university where she studied. Letter 1 says, "We've got a digital version of your Bachelor's Degree in Mathematics that we'd like to give you." Letter 2 says, "We've got a digital version of your Master's Degree in Discrete Mathematics that we'd like to give you." Letter 3 says, "We've got a digital version of your PhD in Cryptography that we'd like to give you."

Now, you can walk over to the table where your brilliant wife is having dinner, and you can lay out those letters one at a time in serial:

```
"You have an offer of a digital bachelor's degree. What would you like to do about it?"
--> "Great. Let's ask them to proceed."
--> Do the work as her agent to make that happen. Then come back to the table.
"You have an offer of a digital master's degree. What would you like to do about it?"
--> "Great. Let's ask them to proceed."
--> Do the work as her agent to make that happen. Then come back to the table.
"You have an offer of a digital PhD. What would you like to do about it?"
--> "Great. Let's ask them to proceed."
--> Do the work as her agent to make that happen. Then come back to the table.
```

But I think most human agents would take a different tack:

```
"You got 3 offers for digital degrees. Which ones would you like to accept?"
-->"All 3."
-->Agent goes away and files 3 sets of paperwork, pays fees 3 times, etc.
```

In the latter case, 3 separate issuance protocols are running, *and they don't have to be changed in any way.* But the agent chooses to combine the UX of all 3 into a single approval ritual with the user. Once the approval is given, the 3 issuance protocols continue to unfold independently. They can succeed or err out independently. They each have a status.

What I am worried about is that our community seems to be equating user experience with protocol design. As a general rule, this equivalence may be a useful shortcut, but there should be NOTHING in protocol design that requires it. And anytime we ask the community to change a protocol just because we want a different UX, we should thoughtfully consider whether we're putting the architectural pressure in the right place.

dhh1128:
PS. This general aggregation-of-UX-for-multiple-independent-phenomena feature is familiar in email and chat apps today, in a non SSI context. You get an individual notifications when individual mails arrive. But if you're like me and you don't answer every message in realtime, it's common to see aggregated stacks in the notifications area on the phone. I may have a 100 unread messages, but they are stacked by conversation, and possibly further stacked by a time grouping or similar, so I don't have to deal with them individually. I can dismiss or mark as read or whatever, on many notifications at once. Or I can explode the aggregate view into individual items and act on them one at a time.

I'm suggesting that mobile SSI wallets could do that same thing with credential offers. If I accept 10 credential offers from the same issuer all at once, the wallet does what's needed to run 10 instances of the credential issuance protocol under the hood without bothering me 10 times for approval.

Again, I'm not arguing that this MUST be done, or even that it SHOULD be done in this case. I'm arguing that nothing is stopping UI engineers from doing it right now. We don't have to change protocols to get better UI. It's always an option to just change UI to accomplish that goal.

smithbk:
I think the human analogy breaks down a bit in that we get postal mail in bulk, once per day.  But if it were delivered in real time as in our case, it would be different.

When I as my wife's agent get the 1st letter, I don't know if I should wait for others to arrive before I tell her about the 1st letter.   And if I decide to wait, for how long?  If I wait too long, my wife might get mad :-)

So we're just proposing that the sender of the letters be able to tell us in advance how many letters they are sending ... and it is expected that they will be sent in rapid succession, so I can wait and give my wife all of her mail at once.

swcurran:
I disagree that as we have framed it @dhh1128 , that this is a major change.  For what has been proposed (optional fields triggering message handling behaviour), the changes meet the semver minor update requirements.  If you are going to propose other ways to address this, then we can evaluate them against the semver rules.

While a motivation for this is UX, the hints provided also affect message processing. Rather than adding the optional fields to let the issuer state their intentions, I interpret your suggestion as saying that there be tribal knowledge with something like -- after receiving an issue, a holder agent may try to infer the intention of the issuer by adding an "wait" event to a queue so that processing on the message can be restarted to see if other "offers" have come in by then.  If so, check to see if the type of the offer is the same as the last, and if so, bundle them together into a notification for the user, otherwise process them separately.  Also, if another offer has come in, wait again, as we don't know if this is the last.  Note that this also implies that the issuer is not to serially issue credentials, but rather send a set of offers in parallel (separate protocols instances) without waiting for the corresponding requests in case the holder decides to batch them together.  And so on...  [Yes there are other ways to do this, but I would argue all require complex handling to infer the intention of the Issuer.]

I don't see how that will ever be implemented, vs. what becomes possible and really simple with the addition of two optional fields in the existing protocol that constitute a minor upgrade.

dhh1128:
> So we're just proposing that the sender of the letters be able to tell us in advance how many letters they are sending ... and it is expected that they will be sent in rapid succession, so I can wait and give my wife all of her mail at once.

That could be done by having a single new field, optional, named "set_of_n" with an integer value. If you want to tell someone that this is one of a set of 3 credential offers, you do `"set_of_n": 3`. Now the agent knows to wait for 3 to arrive. It's a UI hint. The rest of the protocol is unmodified. That would be worthy of a semver minor update.

dhh1128:
>I disagree that as we have framed it @dhh1128 , that this is a major change. For what has been proposed (optional fields triggering message handling behaviour), the changes meet the semver minor update requirements. If you are going to propose other ways to address this, then we can evaluate them against the semver rules.

It is true that the semver rules as currently written say that adding optional fields = a minor upgrade. So if we're going to argue the letter of the law, I concede the point. Hoisted on my own pitard, since I wrote the rules. :-) But this is totally out of harmony with the spirit of the law. The optional fields you have proposed change the state machine from a sequence to a loop -- so every implementation of this protocol now needs to be upgraded not just to recognize additional information, but to support an entirely new paradigm for the workflow. And these optional fields change the meaning of error handling semantics in a substantial way. What happens if there's an error in the middle of a loop? Did the protocol succeed or fail? Is it possible to fail on issuance 2 out of 4, but continue on to issuances 3 and 4 anyway? They also change the meaning of timeouts: do you timeout per iteration through the issuance loop, or for the protocol as a whole? What does retry mean? If you negotiated terms and conditions (separate protocol) or payment (separate protocol) before this issuance began, how does your loop change the meaning of those actions?

This is what I meant when I said that minor changes to fields is not a good way forward. The change looks minor on paper, but it is not minor at all in terms of the mental model of what's happening. The difference is obvious if you compare your proposal to the one I suggested (`set_of_n`), which communicates some extra info that could optionally be used by agents to alter a UI (even to achieve exactly the UI you're after), but that doesn't change the meaning of the existing protocol or its errors, timeouts, retries, or preceding context in any way.

Better to design a protocol that actually communicates about all of these issues, and that describes itself as a multi credential issuance protocol, than to add a couple optional fields that radically change the paradigm a coder must assume when implementing.

smithbk:
Thinking more about this, rather than solving this separately for verification and issuance (and potentially others in the future), what about solving it for all protocols at once without changing any existing protocol.

I think we could do this by introducing a new protocol, call it 'group', which allows the initiator of the group interaction
to tell in advance how many sub-protocols will be included, a high level description of what the grouping is.  It could include messages similar to:

1) group/1.0/start-request - to request starting a group interaction and (optionally?) include a total number of sub-protocols and a description

2) group/1.0/start-response - for the receiver to say whether or not to start the group of sub-protocols

The sub-protocols would then be linked to this parent protocol via the thread.pthid field.  Based on the total field, the receiver would be able to tell when the group is complete.  Of course we could allow an error or abort message to prematurely terminate the group of sub-protocols prematurely.

To support bulk delivery of multiple sub-protocol messages for efficiency, we could also support a 3rd message type ...

3) group/1.0/deliver -  which would contain an array of sub-protocol messages of any type

We may also want/need for the start-request message to include a list of allowable sub-protocol types that it can initiate, so that the initiator can't bait-and-switch by starting other protocols after the receiver has potentially already approved them locally.

One other comment ... I think we need to make sure that any enhancement can tolerate out-of-order message delivery and that handling this properly puts the least burden on the agents to handle properly as possible

swcurran:
This was discussed on the 20210714 and 20210721 calls.

We talked about the proposal from @smithbk about the `group` protocol and decide that while more general, it was likely more complex than we need and we probably need some more experience and use cases for the general feature.  As such, we're more or less "sticking with" the original proposal.

Updated the PR to change the indicators of multiple/more credentials to be counts instead of just true/false indicators. Also removed the suggested of using the attachment to the `issue-credential` as a way to issue multiple different credentials of the same type -- leaving in the original language that this is not permitted. Thus, the only way to use a single instance of a protocol to issue multiple credentials is to use do a `request-credential` / `issue-credential` loop.   Clarified how to use `problem-report` and `~please-ack` in conjunction with multiple credentials.

@dhh1128 I think we (and certainly I) disagree that this is a major change.  There is wording in the RFC to say that the issuer MUST not say that there are more credentials to be issued if the Holder uses the "2.0" message type. That will prevent confusion unless both parties explicitly state that they support the 2.1 or later version of the protocols that allows for looping through issuances.   I think the concerns you raise in your most recent message are not much different whether the issuances are done in a single looping protocol instance or in a loop of many protocol instances.  It's potentially complicated either way and I don't see how this proposal makes it worse.

swcurran:
Per the discussion at the Aries WG Meeting (2022.01.02) -- updated the RFC to highlight where minor version differences between agents may impact processing. Notably when an agent using v2.1 gets a credential-request message from an agent using the v2.0 protocol, they v2.1 agent should NOT use multiple issuance features.

TelegramSam:
Discussed WG 20220309.
This flow only allows the user to see the preview of the first credential. 
Additionally, this would affect json-ld credentials differently, as the associated request actually contains attribute values (where AnonCreds does not contain the attribute values).

Side issue: sometimes the more_availble is a boolean, sometimes a number?

swcurran:
Updated the "more_available" references to always be integers.

TelegramSam (approved):
Discussed WG 20220608.

PR #694, opened 2021-07-21
RFC-06930 - Cross-Platform Credential Representation

horationunez:
Aries Agent developers currently build end user products without a standard method of rendering credentials.
This RFC proposes how the Aries community can reuse available open technologies to build such a rendering method.

Key results include:
- Feasibility of cross platform rendering.
- Enable branding of credentials.


features/0693-credential-representation/README.md
```diff
@@ -0,0 +1,59 @@
...
+The solution we propose is to adopt SVG as the default format to describe how to represent SSI credentials, and to 
+introduce a convention to ensure that credentials values could be embedded in the final user interface. 
+The following images illustrates how this can work:
+
+![SVG + Credential Values](https://i.imgur.com/3ssaQUB.png "SVG + Credential Values")
+
+### SVG + Credential Values
+
+We propose a notation of the form `{{credential.values.[AttributeName]}}` and `{{credential.names.[AttributeName]}}`.
+This way both values and attributes names can be used in branding activities.
```
TimoGlastra, line 47: How would this work with JSON-LD credentials? Does it only support top level values of the `credentialSubject` object (what if it is an array)? Or is this specifically targeted at indy credentials?

horationunez: Yes, we aim to support JSON-LD as well. For non top-level values the path will be is to offer a simple expression language based on F#/Haskell pattern-matching and algebraic types.

swcurran: I was going to ask the same question :-).  Might look at what DIF PE did with JSONPath and then define what that would look like for Indy (which may just work, since it is simple).

gorgsenegger:
As much as I like the idea of a more customisable presentation, how can this be done in a safe way? If I was a malicious issuer, I might style my credentials to look like credentials from a different issuer. A verifier might be tempted to accept the credential based on its "styled looks".

swcurran:
That's a reasonable comment, @gorgsenegger. That said, the intention of digital credentials is to move away from human inspection and on to cryptographic verification.  What you as a verifier don't want to be doing is trusting anything based on the look of something, but on the cryptographic contents and it's tie back to its source.

Verifiers have to be careful in allowing fallback scenarios where humans have to compare data to determine the validity of data.  Presumably, they will see that the data has been verified (e.g. it did come from an issuer that signed it, presented by the holder that may have signed it), and so they really only have to inspect the issuer is one that is trusted.  Doing that by inspection is a really Bad Idea.  If that is going to be done -- a technique like this might be the way to go: https://lifehash.info/.

horationunez:
@gorgsenegger the proposal is aware that forgery/fraud exist outside of the implementation of this RFC. Nothing prevents a developer to this right now, even outside of Hyperledger (consider just cloning the whole Apple Card interface to claim a certain membership).

TelegramSam:
Are any changes or updates needed before we merge this RFC?

horationunez:
@TelegramSam I need to update the branch at this point, and possible update the number.

Not sure about this error of the CI: https://app.circleci.com/pipelines/github/hyperledger/aries-rfcs/1482/workflows/9499c79a-eae6-4bd2-8bf8-5cae5d7c5833/jobs/1487

Does it ring a bell?

TelegramSam (approved):
Discussed WG 20220112

PR #699, opened 2021-10-08
docs: rfc-0699: push notifications apns

berendsliedrecht:
- Rfc that enables agents to exhange a push notification configuration 
- It is possible with the discover features protocols, the roles specifically, to check which notification services another agent can provide.

Interested to hear any feedback!

Signed-off-by: Berend Sliedrecht <berend@animo.id>


TimoGlastra:
PR with client side implementation in AFJ: https://github.com/hyperledger/aries-framework-javascript-ext/pull/41

features/0699-push-notifications/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Sending Push Notifications
+
+When an agent wants to send a push notification to another agent, the payload of the push notifications MUST include the `@type` property to indicate the message is sent by the notification-sender. Guidelines on notification messages are not defined.
+
+```jsonc
+{
+  "@type": "https://didcomm.org/push-notifications",
+  ...
+}
```
TimoGlastra, line 155: Should we include a message_tag (or message_tag_list) in the push notification to indicate which messages it is about.

TelegramSam:
Discussed 20211215. Would prefer a different protocol for each push notification method to simplify discovery and impl of each protocol. This addresses both Android and iOS, and they should be split.

blu3beri:
@TelegramSam It took a while, but I have resolved the feedback from the conversation of a while back. I will soon raise a PR with almost the same RFC, but for Android.

TimoGlastra:
Discussed during 20220518 WG call. Some points to address (for both RFCs):
- make fcm-android just fcm
- in addition to the message_tag for push notification data also add a message_id (as used in the pickup protocol v2)

PR #700, opened 2021-10-12
feat: rfc-0700: oob through redirect

sudeshrshetty:
- RFC about how to send out-of-band invitation through redirect URLs for
intended targets
- Also covers redirect back from recipient to requestor

Interested to hear any feedback!

Signed-off-by: sudesh.shetty <sudesh.shetty@securekey.com>


swcurran:
Probably need to talk through this more, but I found the overuse of the term "redirect" a bit confusing. I think it is used for both URL shortening (Sample 1 and 2) and the decorator that can be used to redirect in the browser after processing.  Is that the intention, or did I miss it?

The "Sample 1" and "Sample 2" just seem to be straight OOB handling. In "Sample 1", the query parameter here is `_invitation`, but per OOB, should be `oob` (per [here](https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband#standard-out-of-band-message-encoding)). The "Sample 2" is straight out of OOB [here](https://github.com/hyperledger/aries-rfcs/tree/main/features/0434-outofband#url-shortening)  I don't think this information should be repeated here, but should reference OOB.

What I think is novel in this is the redirect after the completion of the invitation-initiated actions -- the decorator that indicates where the browser should go next.  Is that right?

sudeshrshetty:
@swcurran yes, the `_invitation` & `_invitationurl` are for sending oob invitation & oob invitation URL to target domain as query parameters. 

`~web-redirect` decorator is for indicating browser or mobile app where to go next.

Initially, I was tempted to use `oob` and `oobid` for sending out-of-band invitation, but it is not clear what are `<domain> ` and ` <path>` of those URLs, are they inviter domains or invitee domains?

For example: in case of `https://example.com/ssi?id=5f0e3ffb-3f92-4648-9868-0d6f8889e6f3`, `example.com` is inviter.

let me know if you have any suggestions regarding this, I am open to change these query parameters. We can use `oob` or `oobid` for these scenarios, but `<domain>` and `<path>` has to be intended targets (invitees) of the invitation.

TelegramSam:
I could also use a walkthrough. Could be in person, could also be in a video recording?

swcurran:
> @swcurran yes, the _invitation & _invitationurl are for sending oob invitation & oob invitation URL to target domain as query parameters.

That is also the purpose of them in OOB -- the domain and path are for the inviter to manage.  As such I recommend that they be removed from this PR (perhaps replaced with references to OOB), and if appropriate, add a PR to clarify the OOB RFC and we can get that updated as part of AIP 2.0 (clarifications that don't change the functionality are permitted to AIPs).

> let me know if you have any suggestions regarding this, I am open to change these query parameters. We can use oob or oobid for these scenarios, but <domain> and <path> has to be intended targets (invitees) of the invitation.

Hmm...now I'm confused.  How does the inviter specify the URLs that are the intended invitees?  All invitees would (presumably -- although not required) get the same <domain> and <path> that are controlled by the inviter, and it is the `@id` that can be unique to the invitee.

> ~web-redirect decorator is for indicating browser or mobile app where to go next.

Definitely like to see how that works.  I think I get it -- this is a web wallet that is processing the invitation, so it is integrated with the browser and can therefore control the browser and send it to a given URL. However, the "inviter" already has control of the browser on the web page and so can take action to redirect the browser after the response from the invitee is received.  We're doing that with mobile wallets -- could it not work the same for browser wallets?  Or is that connection with the browser lost when the web wallet comes into play?

Definitely interested in seeing and discussing this.

sudeshrshetty:
@swcurran @TelegramSam as per our discussion during today's [Aries WG call](https://wiki.hyperledger.org/display/ARIES/2021-11-10+Aries+Working+Group+Call),  I have updated flow diagram with steps.

Please have a look and let me know if anything else needs to be changed

swcurran (approved):
LGTM

PR #702, opened 2021-10-20
update please ack to accepted

TimoGlastra:
AIP 2.0 doesn't include support for please ack decorator, even though it is used in revocation notification and other protocols (issue credential and present proof I believe). We decided to add it to AIP 2.0 but I think we forgot. The state was also still proposed, so I updated it to "Active"


TelegramSam:
If I remember right, we discussed the minor inclusion of the ack decorator, but not necessarily the larger form.

swcurran:
@TelegramSam -- this is the simplified version.  @TimoGlastra did this PR https://github.com/hyperledger/aries-rfcs/pull/644/files to remove all but what you see there.  This version should be included in AIP 2.0.

TelegramSam (approved):
Discussed WG 20211215.

PR #703, opened 2021-10-20
update all references from master to main

TimoGlastra:
The default branch was recently renamed from master to main, but all references were still using master.


swcurran:
Given that there has been no objection to this change, I'll merge it.  These are all clarifying changes, so we can use these as candidates to update the AIP 1.0 and 2.0 lists -- providing there have not be other changes to the RFCs that go beyond clarifications.

PR #706, opened 2021-11-18
Update OOB spec with clarifications

amanji:
* Updates instructions on base64 encoding to include an additional URL encoding step of message before constructing full OOB URLs.
* Adds clarifications to URL shortening instructions for when/how senders should respond with invitations in JSON format.

### DIFF
--- a/features/0434-outofband/README.md
+++ b/features/0434-outofband/README.md
@@ -360,13 +360,13 @@ https://<domain>/<path>?oob=<outofbandMessage>
 
 > To do: We need to rationalize this approach `https://` approach with the use of a special protocol (e.g. `didcomm://`) that will enable handling of the URL on mobile devices to automatically invoke an installed app on both Android and iOS. A user must be able to process the out-of-band message on the device of the agent (e.g. when the mobile device can't scan the QR code because it is on a web page on device).
 
-The `<outofbandMessage>` is an agent plaintext message (not a DIDComm message) that has been base64 url encoded.
+The `<outofbandMessage>` is an agent plaintext message (not a DIDComm message) that has been base64 url encoded and further url (percent) encoded for special characters that may be generated in the base64 string.
 
 ```javascript
-outofband_message = b64urlencode(<outofbandMessage>)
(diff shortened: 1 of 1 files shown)


amanji:
@TelegramSam @swcurran Could we review this at the next WG meeting?

swcurran:
@TelegramSam how about we have an Aries RFC PR review session two weeks from now, including this one.  In the meantime, comments can/should be made. We want to go through the PRs (ideally -- folks review ahead of time), and we merge/not and also decide if any that are merged should impact AIP 2.0.  If they are just clarifications then we can update the link in AIP 2.0 to point to the clarification.

features/0434-outofband/README.md
```diff
@@ -360,13 +362,13 @@ https://<domain>/<path>?oob=<outofbandMessage>
 
 > To do: We need to rationalize this approach `https://` approach with the use of a special protocol (e.g. `didcomm://`) that will enable handling of the URL on mobile devices to automatically invoke an installed app on both Android and iOS. A user must be able to process the out-of-band message on the device of the agent (e.g. when the mobile device can't scan the QR code because it is on a web page on device).
 
-The `<outofbandMessage>` is an agent plaintext message (not a DIDComm message) that has been base64 url encoded.
+The `<outofbandMessage>` is an agent plaintext message (not a DIDComm message) that has been Base64Url encoded and further Uri (percent) encoded for special characters that may be generated in the base64 string.
```
TelegramSam, line 365: The goal of Base64URL encoding is specifically to avoid characters ( '+', '/' and '=') that need percent-encoding. Why the extra step here, exactly?

swcurran: I think it's meant as a clarification about what Base64URL encoding it vs. a second step.  Maybe:

` Base64URL encoded such that the resulting string can be safely used in a URL.`

swcurran:
Updates made as per requested at the Aries Working Group call.  Now ready to merge.

swcurran (approved):

Issue #717, opened 2022-02-06
RFC 0023: Can we accept inline peer did without a signature?


jakubkoci:
DID Exchange [request message attributes](https://github.com/hyperledger/aries-rfcs/blob/main/features/0023-did-exchange/README.md#request-message-attributes) contain the following specification for DID Document:

> * The `did_doc~attach` (optional), contains the DIDDoc associated with the `did`, if required.
 >   * If the `did` is resolvable (either an inline `peer:did` or a publicly resolvable DID), the `did_doc~attach` attribute should not be included.
>    * If the DID is a `did:peer` DID, the DIDDoc must be as outlined in [RFC 0627 Static Peer DIDs](../0627-static-peer-dids/README.md).

Also, as I understood from the rest of the DID Exchange specification, when the `did_doc~attach` is present, it should contain a signature which we need to validate. 

If the `did` is resolvable public DID. We can still trust that the sender (owner of that public DID) controls private keys because it's written to a ledger therefore we don't need to validate signatures. However, if the `did` is inline `peer:did`, there are no signatures present. Can we still accept such `peer:did`?

TelegramSam:
Signatures were never needed in the Requests, and were added by mistake by... me. Does this address your confusion?

TimoGlastra:
This is still an issue, and is surfaced again by the move to did:peer:2.

In this case there's no diddoc to put the signed attachment on (same with public dids).

For the request I think it's fine to nit require a signature, as long as the didcomm message used to send the request is authenticated using one of the keys from the diddocument.

However the response signature it's required when you rotate the key used from the invitation in the response. We should have an alternative. I think signing the did from the invitation (as base 64) should suffice in this case.

Can we make a modification to the protocol to allow:
1) the request message not to be signed. For did:peer:1 dids it's still required to not break interop, but for all resolvable dids, the signed attachment SHOULD NOT be present
2) do not require the response message to be signed. For did:peer:1 dids it's still required to not break interop, and if the did from the invitation is rotated, it must also be signed by the invitation did. for all resolvable dids, the signed attachment SHOULD NOT be present. If a resolvable did is used, and the key/did from the invitation is rotated in the response, a new property should be added to the response (e.g. "invitation_did~attach") that contains the did from the invitation that is being rotated base64 encoded, and signs it with a key from the invitation did. This authenticated the did rotation from invitation to response (and is important to keep for multi-use invitations with public dids that rotate to peer dids)

[...]

PR #725, opened 2022-03-17
Remove Signatures from Example Requests

TelegramSam:
These were never appropriate, but added by mistake to the request. Sigs still required in Responses.
Related to #717

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -154,14 +154,7 @@ The _requester_ may provision a new DID according to the DID method spec. For a 
       "@id": "d2ab6f2b-5646-4de3-8c02-762f553ab804",
       "mime-type": "application/json",
       "data": {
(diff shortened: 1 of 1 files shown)


TimoGlastra:
Awesome thanks. Maybe we can add a few words on why the signature is still required for the response? I think it would also be valuable to mention with which key(s) you need to sign the attachment? It seems this is only needed if the keys used for the invitation are different than the keys used in the response right?

TelegramSam:
@TimoGlastra this will be discussed in the Aries WG call this week.

TimoGlastra:
Hey @TelegramSam, what did we land on in the end with signing the response message if a public did or peer did method 2 did is used? Should we just sign the resolved did document and attach that?

TelegramSam:
Discussed 20241009. Points handled by DID Exchanged 1.1, and moving away from DID Exchange anyway in DIDComm v2.

PR #726, opened 2022-03-21
fix: force the connection into a complete state

KimEbert42:
When connecting, a race connection can be created. Some invitees will process messages out of order, and a connection may not be in a state that is ready to receive additional messages. The work around is to sleep for an estimated number of seconds. This may cause connectivity issues between invitee and the inviter, and a poor experience in the community.

The recommendation is to change the language so that the invitee MUST send a message to complete the connection, so the inviter may know when it can send additional messages.

Signed-off-by: Kim Ebert <kim@indicio.tech>


TimoGlastra (commented):
I agree we _should_ have made this change from the start, but I'm not sure we should make this change now.

The connection protocol is about to be deprecated and replaced by the did exchange protocol which does require completion. Making this required now will only increase complexity as you can't assume implementations will update a deprecated protocol.

I think a 'workaround' most implementations use is to just interpret the state of response as active.

reflectivedevelopment:
Old protocols seem to have a way of living a lot longer than we ever intend them to live. I think we should still make the change to encourage proper behavior.

reflectivedevelopment:
I will update wording here to clarify the agent may see other agents not doing this behavior.

b1conrad:
Totally understand the hope that implementers will move away from deprecated protocols.

That being said, I need to second the suggestion to make the change in the spec. We (ACA-Pico) will likely continue to use the connections protocol for some time. Eventually, we will move to did-exchange, but I cannot predict when that will be.

TelegramSam:
Discussed 20241009. Closed as connections is being actively phased out.

PR #733, opened 2022-04-15
docs: remove invalid note for pickup protocol v2

JamesKEbert:
Removes an invalid note regarding message delivery in pickup protocol v2. This note originated from a first iteration of this RFC in which messages would be sent directly down one at a time instead of being bundled in attachments in a single delivery message.


TimoGlastra (approved):
LGTM

PR #737, opened 2022-05-19
docs(0699): mention  of pickup v2

blu3beri:
Fix as discussed in the previous WG call.

Signed-off-by: blu3beri <berend@animo.id>


TimoGlastra (approved):
LGTM

TimoGlastra:
This has been open for quite some time with no other feedback so I'm inclined to merge it. @blu3beri as we discussed it would probably be good to take versioning serious so this should probably be version 1.1. Could you make the change?

TimoGlastra:
Also incorporated this change into #745

PR #747, opened 2022-07-26
Credential Supplements

TelegramSam:


TelegramSam:
An alternative is to use a generic 'extras' structure instead of a structure for each purpose, but this leads to custom attributes that are more painful in some strongly typed languages.

I used the generic `~attach` method to make the transition to DIDComm v2 attachments easier.

TelegramSam:
Discussed WG 20220727.
Other types of 'supplements' (thanks Lance!) that might be useful: Proof of non-revocation, device binding attachments. That's 4.

Warren had the suggestion to generalize in a schema predictable way. by using key value pairs under the type of supplements.

TelegramSam:
Updated Issuance, but not Presentation, to illustrate the differences in approaches. Ready for more discussion on Wed.

features/0453-issue-credential-v2/README.md
```diff
@@ -10,6 +10,10 @@
 
 ## Version Change Log
 
+### 2.2 - Addition of Supplements
+
+An optional mechanism for providing credential supplements during issuance. Supplements are also used in credential presentation.
```
swcurran, line 15: @TelegramSam -- sorry for the delay in going through this, especially because I want to get this PR merged so that I can reference it in my own RFC :-).

I think this RFC needs more content here about what a supplement is for -- clarity. An example of how to use both of the types of supplements would be good -- especially a hashlink example, making clear how the key/value pair(s?) are used. An issuer-credential example would be good as well.

features/0453-issue-credential-v2/README.md
```diff
@@ -303,19 +307,34 @@ Message Format:
...
         {
-            "attach_id" : "<attach@id value>",
+            "attach_id" : "<attachment identifier>",
             "format" : "<format-and-version>",
         }
     ],
     "credentials~attach": [
         {
-            "@id": "<attachment-id>",
+            "@id": "<attachment identifier>",
```
swcurran, line 316: Are you taking into account the 3 types of attachments? For example, if the attachment type is "link", the inclusion of a hash is automatic.  I'm wondering if we should minor-version update the attachment RFC to include an optional hash when base64url encoding is used.

I would think that we might also want to talk about the option of signing the attachment, as is possible with the attachment RFC.  That would ensure that the supplement is signed in a way that both the holder and subsequent verifiers can verify.

swcurran:
I'd like to use this approach for OCA, so would like to move this forward sooner than later.  For OCA, the issuer would include an "official" type of perhaps "credential-oca", and the attachment would be most likely an "link" attachment, with a signature on it. That way the OCA file can accompany a proof to the verifier using the same technique.  If either the holder or verifier did not understand OCA -- they would just ignore the supplement.

swcurran:
Discussed on the [Aries Working Group Call 2022.10.19]( https://wiki.hyperledger.org/display/ARIES/2022-10-19+Aries+Working+Group+Call)-- ready to merge, approved, no one was against merging the PR.  Some (OK me) think that clarifications such as examples would be helpful, but that can be added after merging.

Last call for objections for merging this.

swcurran (approved):
LGTM

PR #748, opened 2022-08-03
Proposal of RFC for n-wise DID exchanging

mikelytaev:


genaris:
What an amazing RFC! I think group chats over DIDComm is something definitely needed and this approach looks promising.

The only thing that I see that could be missing from existing group chats would be the possibility of having multiple administrators. This spec explicitly states that there could be only a single owner. What are the technical challenges of allowing multiple owners or adding an administrator role?

Another question that arises when thinking about these group relationships is what happens if the owner leaves the n-wise: other systems automatically set an existing participant as owner/admin. In this case the n-wise relationship will become 'orphan'. I guess this could be solved by remaining parties that could create another group (however who will be the leader of such new group?).

mikelytaev:
Thank you for your appreciation and interest in our work!
While creating this version of the protocol, we discussed the problem of users hierarchy and considered various options. In particular, we considered the idea to adopt the users hierarchy  from Telegram or Whatsapp chats. However, any complex hierarchy of chats defined in the rfc simultaneously makes the chat highly specialized. The ideal option would be to develop a pluggable hierarchy so that the chat creator can determine user categories depending on the specifics of the group. But this is a complex topic that requires further research and possibly a separate rfc. So in the first version of the protocol, we decided to define the simplest hierarchy possible so that it could be easily implemented. We plan to observe and analyze the usage of this version of the protocol and take this into account in the next versions.
I agree with the comment about orphan. But the chat can also become orphan if the owner stops using it or loses his wallet. So we decided not to consider this case in this version of the protocol for simplicity reasons.

talltree:
So this might be a very naive question, but what's the relationship between this RFC and the discussions about group messaging going on in the DIF DIDComm Working Group? DIDComm V2 is in the can at this point, but group messaging is a key potential feature of V3.

mikelytaev:
Thank you for your comment! We are closely following the progress of DIDComm development. Judging by the available publications, the current work on group messaging in DID Comm WG has not yet reached the possibility of its implementation.

In our solution, the task of consensus between the participants is clearly highlighted and several options for its solution are proposed. In addition, we have a complete implementation based on Sirius-SDK-java.

We believe that the aries rfc platform is well suited for testing such complex protocols as group communication. We hope that the proposed ideas will find its application in strict DIF standards in the future.

dhh1128:
@mikelytaev : There is someone who wants to write a paper about group messaging atop DIDComm and is planning to convene meetings about it at the next Rebooting Web of Trust conference in September. His name is Karim Stekelenburg. I don't have his email, just his handle on Discord. However, I will point him at this RFC so he is aware. Perhaps you can collaborate.

The DIDComm Users Group at DIF is actively discussing this topic, BTW (see [here](https://github.com/decentralized-identity/didcomm-usergroup/blob/main/meetings/2022-04-06.md#6-bridging-rich-chat-with-didcomm), for example). You are welcome to come and present, and/or ask questions [there](https://github.com/decentralized-identity/didcomm-usergroup/blob/main/schedule.md).

genaris:
@karimStekelenburg

swcurran:
Discussed on the [Aries Working Group Call 2022.10.19]( https://wiki.hyperledger.org/display/ARIES/2022-10-19+Aries+Working+Group+Call)-- ready to merge, needs approval, no one was against merging the PR.  This is still an experimental protocol, as noted in the header, and so it will likely require some uptake by others to move this forward beyond this point.

Approval needed and last call for objections to merge this PR.

swcurran:
@mikelytaev -- just noticed that the DCO check is failing (DCO - Developer Certificate of Origin - https://github.com/apps/dco).  Can you please update the bad commit?  See the "Details" link on the failed check for how to do that.  Worst case if you are not a Git/GitHub guru is to make a copy of the folder for this PR close this PR and open a new one that handled DCO properly.

Thanks!

mikelytaev:
> @mikelytaev -- just noticed that the DCO check is failing (DCO - Developer Certificate of Origin - https://github.com/apps/dco). Can you please update the bad commit? See the "Details" link on the failed check for how to do that. Worst case if you are not a Git/GitHub guru is to make a copy of the folder for this PR close this PR and open a new one that handled DCO properly.
> 
> Thanks!

fixed

swcurran (approved):
Approved per discussion at Aries WG Call on 2022-10-19.

PR #751, opened 2022-08-18
Clarifications to RFC 0592 AnonCreds Attachments about unrevealed attributes and string integer encoding

swcurran:
As discussed in the [Aries WG Meeting, August 17, 2022](https://wiki.hyperledger.org/display/ARIES/2022-08-17+Aries+Working+Group+Call), this PR adds clarifications about the processing of unrevealed attributes in a proof request (verified, and the verifier business logic must decide if not revealing all claims is sufficient in a given use case), and the handling of strings of integers when encoding claims (convert them to integers).

### DIFF
--- a/features/0592-indy-attachments/README.md
+++ b/features/0592-indy-attachments/README.md
@@ -367,7 +367,7 @@ A gist of test value pairs can be found [here](https://gist.github.com/swcurran/
 
 #### Notes on Encoding Claims
 
-- In converting any string integer to an integer, leading 0s in the string are (by definition) not part of the integer. They leading 0's remain in the (untouched) `raw` value.
+- In converting any string integer to an integer, leading 0s in the string are (by definition) not part of the integer. The leading 0's remain in the (untouched) `raw` value.
 - The use of AnonCreds predicates, such as proving "older than 21" based on a date of birth claim without sharing the date of birth, is based on an expression involving the `encoded` value. Thus, only `raw` integers or string integers can be used in AnonCreds predicates.
 
 ## Implementations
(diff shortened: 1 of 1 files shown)


features/0592-indy-attachments/README.md
```diff
@@ -346,6 +365,11 @@ An example implementation in Python can be found [here](https://github.com/hyper
 
 A gist of test value pairs can be found [here](https://gist.github.com/swcurran/78e5a9e8d11236f003f6a6263c6619a6).
 
+#### Notes on Encoding Claims
+
+- In converting any string integer to an integer, leading 0s in the string are (by definition) not part of the integer. They leading 0's remain in the (untouched) `raw` value.
```
TimoGlastra, line 370:
```suggestion
- In converting any string integer to an integer, leading 0s in the string are (by definition) not part of the integer. The leading 0's remain in the (untouched) `raw` value.
```

TimoGlastra: I'm not sure I fully understand, but does that mean the value of `012` will be encoded as `12`? If we have the creditcard CVC code it will be impossible for it to start with a `0`. IMO it should be encoded as as the string value (so `012`). Or is that what this suggests?

swcurran: The "raw" value will remain "012".  The encode value will be the integer 12, and will NOT be the hash of "12".

As long as you use the raw value, which you should, you are good.  And of course, you should make sure that the encoded  value is the expected integer 12.

Does that make sense?

swcurran:
Dang...the "sign and commit" button didn't put my signing on the commit.  

I don't know how to update the PR to fix the DCO problem, so I'll close and open a new PR.

Sigh...

swcurran:
Nevermind...the DCO fix I did worked -- yay!!!

FYI -- I just followed the instructions on the "Details" link of the DCO  check.

FYI 2 -- had I just added the DCO signing, I could have done the merging of the change into the PR without having to fix it.

TelegramSam (approved):
Discussed WG 20220907.

PR #752, opened 2022-09-06
Clarify the handling of minor versions of a protocol

swcurran:

This is in response to an interop issue between Aries Framework JavaScript and Aries Cloud Agent Python, where ACA-Py is not properly handling the messages sent.  The Aca-Py issue is [here](https://github.com/hyperledger/aries-cloudagent-python/issues/1917).


genaris:
In last WG call I was surprised that the minimum minor version an agent can support for a given protocol could be different from 0, as I had understood (in concordance with what you said in the ACA-Py issue referenced) that minor versions of a protocol could only add some optional features in such a way that there would always be a way to negotiate communication for two agents that support the same major version of it (e.g. an agent that supports OOB 1.1 and another that only supports OOB 1.0).

Reading a bit more the same RFC, I found a few sentences in **Recipient Rules** that I think are the reason of my confusion:

- "For major version >= 1, recipients should also accept messages that differ only in that the message's minor version is earlier than their own preference. "
- "Message types that differ in only in minor version are guaranteed to be compatible for the feature set of the earlier version. "

So this gives me the idea that if I support OOB 1.4, it should be enough to send that single version when disclosing my features, at it implicitly means that I accept OOB 1.0, 1.1, 1.2, 1.3 and 1.4 messages (provided the protocol version history has followed the convention stated in this RFC). If the other agent implemented OOB 1.7, it will simply expect from me to send 1.4 messages and should be fine with that (and I'll accept its 1.7 messages as well, ignoring the features I don't support).

What happens if an agent is willing to do an interaction using a specific version of the protocol? For instance, suppose that OOB 1.4 adds an optional field called "imageUrl" and that particular field is very important for it. In such case, if it interacts with another agent that only supports 1.3,  it would like to abandon the flow. Is that possible or it's something forbidden? As the first sentence I quoted uses the term 'should', it makes me think that such behaviour is not allowed.

swcurran:
Discussed on the [Aries Working Group Call 2022.10.19]( https://wiki.hyperledger.org/display/ARIES/2022-10-19+Aries+Working+Group+Call)-- ready to merge, needs an approval, no one against merging.

Can we please get a editor/maintainer to approve this PR?

TelegramSam (approved):
Finally gave this a read. Much improved!

swcurran:
@TelegramSam -- can we have a brief discussion about this one on the Aries Working Group call tomorrow? There is one more topic to talk about with it.

PR #755, opened 2022-09-29
OCA for Aries and OCA for Aries Style Guide RFCs

swcurran:


features/0755-oca-for-aries/README.md
```diff
@@ -0,0 +1,402 @@
...
+  credential data.
+- header_background_color - an RGB color code for the credential display header
+  background color.
+- footer_background_color - an RGB color code for the credential display footer
+  background color.
+- header_font_color - an RGB color code for the credential display header font
+  color.
+- footer_font_color - an RGB color code for the credential display footer font
+  color.
+- primary_attribute - the name of the primary capture base attribute to display
```
cvarjao, line 179: what is the "primary attribute" and "secondary_attribute"? is that the "Issuer Name" and "Credential Name", respectively?

features/0755-oca-for-aries/README.md
```diff
@@ -0,0 +1,402 @@
...
+additional "extension" overlay, the [credential_branding overlay](#), that
+provides the issuer with a way to provide a series of data elements about the
+credential branding that they would like to see applied. The credential_branding
+overlay provides a list of name/value pairs, and the holders (and verifiers) can
+use the values when displaying the credential of that type from that issuer. The
+holder/verifier controls the details of how the credentials are displayed, but
+the issuer does have some input into the branding. The following are the branding
+data elements the issuer can set:
+
+- logo_uri - a URI pointing to a logo to display with the credential.
```
cvarjao, line 165: Can we assume that the `logo` and `background_image` is an URL? embedded base64 provided as [Data URLs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs) (e.g.: `data:image/png;base64,...`)

cvarjao:
I think we need to call out allotted image sizes (logo, backgrounds) and how they are scaled up/down

features/0755-oca-for-aries/README.md
swcurran: The idea is to allow the issuer to suggest to the wallet (display software) the "best" attributes from the credential to display on top of the background image.  In other words, if the wallet (or whatever) displays the "card", it could follow the issuers suggestion and put one or both of the primary and secondary values (perhaps with their multilingual labels).  They might put those instead of "Issue Date" (which is NOT an attribute in the credential -- it's metadata collected by the wallet), or in addition to "Issue Date".  Without some guidance from the issuer, the wallet would have no information to decide which attributes might be helpful to show.

Does that sound useful?  I confess that it was an idea of mine that I thought might be useful.

swcurran:
Updates made to the RFC per the comments received.  I still want to add the Credential Branding Guide to this RFC, and to deprecate/archive the existing overlays RFC as part of this PR.

knguyenBC:
In regards to the primary and secondary attribute, there's a few usability concerns for in-person interactions. In user testing, it's been observed that people instinctively show the digital credential rather than go through the scanning process and there's concerns that adding a primary and secondary attribute to the front facing credential design will encourage this behaviour. 

That being said, we've also noted that people prefer to have a recognizable attribute to their credential. For example their student ID should have their name.

Then there's the concern with real estate. The more we add to the credential itself, the more busy it'll look and the focus shouldn't be about the credential. 

If the goal is to allow issuers differentiate between credentials of the same type and to help users recognize their credentials, I wonder if there's a different solution for that.

swcurran:
Significant update to the RFC. AFAIK, other than applying feedback, the only remaining requirement for completing this RFC is the conversion of the Style Guide to Markdown format.

I would definitely appreciate a serious review of this PR by those involved in OCA for Aries.  @cvarjao @foxbike Tim B.

swcurran:
Added the OCA for Aries Style Guide. Changed the "branding" overlay to be a Meta overlay that has the language code "branding".  Updated the instructions for creating and maintaining an OCA Source file, and generating an OCA Bundle from the source file.

pknowl:
@swcurran - From a data harmonisation perspective, adding anything other than ISO 639 language codes [https://www.iso.org/iso-639-language-codes.html] as values for the "language" attribute won't fly. However, adding branding assets into a language-agnostic schema-level overlay (as opposed to an existing language-specific schema-, attribute-, entry code-, or unit-level overlay) makes sense. Furthermore, as OCA is such an extensible and scalable architecture, there is scope to add several new overlay types without having to shoehorn requirements into existing ones. 

The current OCA Specification (v1.0) shows a delineation of "Semantic", "Inputs", "Transformation", and "Presentation" overlay categories. However, the next version of the spec will likely see that delineation changed to "Collection" (i.e., overlays for capturing metadata to enable data context to be preserved at the point of collection), "Validation" (i.e., overlays for capturing metadata to allow validation checks between format definitions and data inputs), "Transformation" (i.e., overlays for capturing metadata to enable transformation mapping from source to target schemas), and "Presentation" (i.e., overlays for capturing metadata to allow the strong binding of presentation assets to transient data objects (e.g., forms, credentials, etc.)). 

Over the past two months, that reasoning has come about through valuable discussions in the OCA community. It aligns the conceptual modelling of the data management and technology leads. 

For some neat noodling on Presentation-type overlays, Andrew Slack from SICPA put forward three new overlay types for review in the following RWoT paper. The naming aside, the methodology is in keeping with pure OCA methodology. In other words, his suggestions refrain from shoehorning required capture requirements into existing overlay types that don't necessarily fit those requirements. 

https://github.com/WebOfTrustInfo/rwot11-the-hague/blob/master/final-documents/data-exchange-agreements-with-oca.pdf 

Head straight to Appendix B of the paper.
 
In short, we have more freedom than we had initially thought possible. OCA's main task is to ensure data and metadata integrity [https://humancolossus.foundation/blog/accuratedata1]. 

The architecture is robust yet flexible enough to undertake this ambitious mission elegantly.

swcurran:
Regarding this:

> @swcurran - From a data harmonisation perspective, adding anything other than ISO 639 language codes [see [link](https://www.iso.org/iso-639-language-codes.html)] as values for the "language" attribute won't fly. However, adding branding assets into a language-agnostic schema-level overlay (as opposed to an existing language-specific schema-, attribute-, entry code-, or unit-level overlay) makes sense. Furthermore, as OCA is such an extensible and scalable architecture, there is scope to add several new overlay types without having to shoehorn requirements into existing ones.

Agree that it is a hack. Is there going to an Overlay added to the core spec to do this?  Or should we go back to using a "branding" overlay name?

From a practical aspect, we want the OCA Parser tool to support as much of this as possible. For now, we can simply rename the overlay after generation, but it would be nice to have a way to support this WITHOUT changing the tool.  For such a minor adjustment, it would be painful to have to fork the OCA Parser.

swcurran:
Regarding this from @pknowl :

> The current OCA Technical Specification [see [link](https://oca.colossi.network/specification/)] shows a delineation of "Semantic", "Inputs", "Transformation", and "Presentation" overlay categories. However, the next version of the spec will likely see that delineation changed to "Collection" (i.e., overlays for capturing metadata to enable data context to be preserved at the point of collection), "Validation" (i.e., overlays for capturing metadata to allow validation checks between format definitions and data inputs), "Transformation" (i.e., overlays for capturing metadata to enable transformation mapping from source to target schemas and datasets), and "Presentation" (i.e., overlays for capturing metadata to allow the strong binding of presentation assets to transient data objects (e.g., forms, credentials, etc.)).

This is not really overly relevant to this RFC, but my $0.02CDN is that many of the (current) overlays are applicable across categories, so I'm not sure it is very helpful to categorization the overlays. For example, in putting together this RFC, we selected the Overlays that were most helpful to an entity receiving data and an OCA Bundle. The fact that some of those are in the "collection" category and we are using the OCA primarily for presentation is irrelevant. As long as the issuer is providing data matching their overlays, we're happy.

tbloomfi:
For the meta layer, we might want to consider the following:

- offer_expires - this would allow the wallet to display an offer expiry message on the offer screen. The attribute could also be used to clean up older offers. The controller would enforce the expiry but this could be good for usability. Didcomm v1 has an optional decorator expires_time (in v2 this is part of the message header) that could also be used, but the meaning is not really the same. The V2 spec does include a comment that protocols could "nuance" this header in their respective spec.

- credential_privacy_policy_url - This may also be part of the machine-readable governance or perhaps a specific consent receipt issued as a supplement. Verifiers would also need to share privacy policies. This could be a fallback.

swcurran:
@tbloomfi -- a group of us talked about your suggestions and have this response:

- `offer_expires` -- we think that really belongs in the issue-credential protocol. This could be handled by adding support for [RFC 0032 Message Timing](https://github.com/hyperledger/aries-rfcs/tree/main/features/0032-message-timing) decorator. Or, it could specifically added to the `credential-offer` message (although that seems unnecessary).  Either way, we don't think it belongs in the OCA Bundle.

- `credential_privacy_policy_url ` -- definitely one that could be added.  Would it make sense to expand the context to `terms_and_conditions_url`, which would encompass the Privacy Policy and any other items the users wants to share?

A caveat on this.  As noted in the propose RFC, the use of links in the OCA Bundle could be used as a tracking mechanism.  Please see that part of the RFC.  Here's a [link in the pull request branch](https://github.com/swcurran/aries-rfcs/tree/oca4aries/features/0755-oca-for-aries#warning-external-attachments).  I will expand that section a bit to include links.

TelegramSam:
I think the concept of the card view is good, but struggle with the header. That header design presumes some things about the overall app layout that may not be true for all current designs, or future designs either. I DO think the concept card layout is likely to be much more tolerant of layout changes etc, and so inflicts fewer opinions on the overall app layout.

pknowl:
Regarding `terms_and_conditions_url` idea, I would treat it as a separate OCA bundle. Have a look at the [Data Agreement Specification](https://docs.igrant.io/docs/data-agreements/). We could build any data agreements using OCA methodology and then strongly bind that new bundle to the original one.

tbloomfi:
@swcurran expiry timer seems like a better option. Also terms_ and_conditions_url works. It does not need to be an explicit privacy policy.

A question around problem reports has also come up while working with an issuer that has done a great job documenting corner cases - for example, what happens if the credential offer is accepted after the data has changed or the holder is no longer eligible to receive the credential? 

The proposed approach is to use a problem report. Seems simple enough until the translation of the message is considered. In addition, the issuer may want to use error codes which would need to be mapped to user-friendly messages in the wallet. Is there a use case here for OCA to handle error codes and the display of localized error messages?

For two languages the issuer could just include both languages in the message, but not a very clean solution.

pknowl:
In the [Decentralised Semantics WG](https://humancolossus.foundation/decentralised-semantics-wg) call on [January 26th](https://wiki.colossi.network/en/Community/Working-Groups/DSWG/Meeting-Notes/2023-01-26), one of the agenda topics was "What is considered "Metadata" in OCA?"

The key takeaways were ...
(i.) “Metadata” (as defined [here](https://en.wikipedia.org/wiki/Metadata)) is important for cataloguing, but, for the most part, doesn't require new overlay types to capture the information. 

(ii.) With OCA, we can differentiate between _“data”_ and _“metadata”_ as follows:
- _“data”_ is inputted data into a semantic structure logged at a given moment in time (i.e., event-triggered);
- _“metadata”_ is supporting information captured by the semantic structure that brings definitional or contextual meaning to “data”.

What does that mean in practice?
As with any free-form inputted data, Information describing a parent schema (e.g., descriptive or legal information about the schema) can be flagged as sensitive. In OCA terminology, this information is captured in associated _metadata schemas_ and, as such, is treated as _“data”_ that describes the parent schema. As with any other OCA bundle, you would also define _“metadata”_ for those _metadata schemas_. In other words, the information describing a parent schema can also be translated into multiple languages, nuanced formatting, etc.

With that in mind, Human Colossus is currently working on a semantic engine project with a research institute in Canada working through their cataloguing requirements. This will inevitably result in us defining a couple of "metadata schemas" using the categories described in the [wikipedia definition](https://en.wikipedia.org/wiki/Metadata) [e.g., Administrative metadata schema, Legal metadata schema, etc.]. Once we've got something solid, I'll ensure that we include tabular definitions of those _metadata schemas_ as appendices in the next version of the official [OCA Technical Specification](https://oca.colossi.network/specification/).

I wanted to inform the OCA for Aries group as I'm sure this information will be useful for credential objects also.

dbluhm (approved):

PR #769, opened 2023-02-16
fix: update versions in indexgenerate

anwalker293:
Primarily updating version of Node/Node dependencies in indexgenerate flow


anwalker293:
@TelegramSam

PR #776, opened 2023-03-17
RFC 0771: 'cred' to 'credential' in identifiers

genaris:
Fix a few typos in identifiers (cred-filter -> credential-filter and cred-request -> credential-request) to be consistent with the rest of the document.

### DIFF
--- a/features/0771-anoncreds-attachments/README.md
+++ b/features/0771-anoncreds-attachments/README.md
@@ -57,7 +57,7 @@ A complete [`propose-credential` message from the Issue Credential protocol 2.0]
   "formats": [
     {
       "attach_id": "<attach@id value>",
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

TelegramSam:
Does this change match the historical, currently in-use reality?

genaris:
> Does this change match the historical, currently in-use reality?

AFAIK this is not yet used in production, and the first and currently only implementation is the one for AFJ (where I noticed this inconsistency). That one of course matches this change.

swcurran (approved):

swcurran:
@genaris Please update the branch so we can merge this.  Thanks!

genaris:
> @genaris Please update the branch so we can merge this. Thanks!

Done!

PR #781, opened 2023-04-15
using BTreeMap

Parikalp-Bhardwaj:
`Use a HashMap when`:

You want to associate arbitrary keys with an arbitrary value.

You want a cache.

You want a map, with no extra functionality.

`Use a BTreeMap when`:

You're interested in what the smallest or largest key-value pair is.

You want to find the largest or smallest key that is smaller or larger than something.

You want to be able to get all of the entries in order on-demand.

You want a map sorted by its keys

### DIFF
--- a/.circleci/requirements.txt
+++ b/.circleci/requirements.txt
@@ -1,2 +1,2 @@
-pytest>=4.0
+pytest>=7.3.1
 requests>=2.7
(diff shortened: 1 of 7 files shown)


swcurran:
This PR cannot be merged while failing the checks — notably DCO (DCO - Developer Certificate of Origin - https://github.com/apps/dco).  While an interesting change, AFAIK, it doesn’t change the behaviour.

Closing.  Please reopen if you think I’m missing the point.

Parikalp-Bhardwaj:
@swcurran can you check this

PR #784, opened 2023-06-01
[RFC0721] Add generic revocation format for anoncreds

genaris:
Adding a dedicated revocation format entry for AnonCreds in Revocation Notification V2 protocol.

I'm not sure if it's actually needed, as we can probably use the existing 'indy-anoncreds' if we consider it appropriate for both indy and the standardized anoncreds specification. At least from the formatting (with the :: as separator) it seems it suits pretty well if all AnonCreds object IDs are URIs.

Not sure also if it would be needed to version it (like `anoncreds-v1`) in case AnonCreds V2 data model require a different approach.

### DIFF
--- a/concepts/0008-message-id-and-threading/README.md
+++ b/concepts/0008-message-id-and-threading/README.md
@@ -71,6 +71,10 @@ Message threading will be implemented as a [decorator](../0011-decorators/README
 The `~thread` decorator is generally required on any type of response, since
 this is what connects it with the original request.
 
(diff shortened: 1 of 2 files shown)


TimoGlastra (approved):

swcurran:

@genaris -- can you please update your branch and we'll get this merged?

genaris:
> Discussed on the Aries Working Group call on 2023.06.28 and there was agreement to merge.
> 
> @genaris -- can you please update your branch and we'll get this merged?

Done!

PR #785, opened 2023-06-05
Revise the well known goal codes aries.rel and aries.rel.build

swcurran:
Updates the definition of where to use "aries.rel" and "aries.rel.build to remove the "can't be used to establish a DIDComm connection". We see no reason for that limitation. 

We have come across the need for a goal code that we can use when doing nothing more than making a DIDComm connection between agents controlled by humans. Scenarios:

- Two Mobile Wallet apps, with an invitation from one for the other in order to establish a connection to be used for a variety of purposes - messaging, issuance, and verification.
- An enterprise agent sends an invitation to a Mobile Agent (via a QR code) to establish a connection for a variety of purposes - messaging, issuance, and verification.

### DIFF
--- a/concepts/0519-goal-codes/README.md
+++ b/concepts/0519-goal-codes/README.md
@@ -126,17 +126,35 @@ aries | Hyperledger Aries Community | TBD
 The following goal codes are defined here because they already have demonstrated utility, based on early SSI work in Aries and elsewhere.
 
 ##### `aries.vc`
+
 Participate in some form of VC-based interaction. 
+
 ##### `aries.vc.issue`
+
 Issue a verifiable credential. 
+
 ##### `aries.vc.verify`
+
 Verify or validate VC-based assertions.
+
 ##### `aries.vc.revoke`
+
 Revoke a VC.
+
 ##### `aries.rel`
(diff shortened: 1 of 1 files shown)


swcurran:
FYI -- @cvarjao

dhh1128 (approved):
I am okay with this. However, I would like to explain the history as I remember it. IIRC, we wrote this caveat because people were declaring a dislike for DIDComm's pattern of establishing a "connection" before running protocols. They felt that this was "heavy." They had also implemented UI that asked users before establishing such a connection. @TelegramSam and I felt that this was an antipattern, and that DIDComm "connections" should be thought of more like SSL tunnels. Your browser doesn't ask you in advance if you want an SSL tunnel; it just builds one, automatically, and THEN asks you if you want to login or do other stuff. So we felt like making a DIDComm connection a goal was teaching the wrong idea -- that this was a heavy operation requiring user consent/intention, where we wanted it to feel lightweight and automatic (a precursor to interesting goals).

If actual experience is, now, that people look at building a DIDComm connection as a meaningful goal unto itself -- and not one that users should have to consent to, normally, but one that users have to help with when scanning a QR code -- then I'll go along with community wisdom.

PR #789, opened 2023-06-20
Add clarification to thread RFC about handling an empty thread decorator

swcurran:

See issue [#875 in Aries VCX](https://github.com/hyperledger/aries-vcx/issues/875).  This PR clarifies that while it is not recommended, an Aries agent MAY put an empty `~thread: {}` item in the first message of a protocol instance, and a recipient Aries agent MUST be able to handle that.

### DIFF
--- a/concepts/0008-message-id-and-threading/README.md
+++ b/concepts/0008-message-id-and-threading/README.md
@@ -71,6 +71,10 @@ Message threading will be implemented as a [decorator](../0011-decorators/README
 The `~thread` decorator is generally required on any type of response, since
 this is what connects it with the original request.
 
+While not recommended, the initial message of a new protocol instance MAY have an
+empty (`{}`) `~thread` item. Aries agents receiving a message with an empty
+`~thread` item MUST gracefully handle such a message.
+
 #### Thread object
 
 A thread object has the following fields discussed below:


usingtechnology:
That should help clear up the ambiguity.
The code in `aries-cloudagent-python` is very defensive and assumes `~thread` may not exist and `~thread{}` may exist without a `thid`.  All applications should be coded with the same defensiveness.

usingtechnology (approved):
LGTM.

gmulhearn-anonyome (approved):
LGTM. But curious; why go this route rather than a route of "~thread MUST have a `thid`"?

swcurran:
My thought is that making the `thid` required is a change in the protocol (and hence a minor version change), and simply being defensive in the implementation is sufficient.  And, even if we make the change now, some implementations may not do it, and so it is best to be defensive anyway.

gmulhearn-anonyome:
thanks, sounds good

usingtechnology:
Not that ACA-Py's implementation should dictate the RFC, but whoever implemented it in ACA-Py understood it as expecting to encounter a thread without a thid. And unfortunately, in ACA-Py the thread logic is not 100% isolated. Ensuring that all threads had a thid with a value would be a fairly significant exercise, certainly at the regression testing level.

TimoGlastra:
I think it's valid to have a pthid, without a thid, so making this required would also be problematic in such cases

swcurran:

PR #792, opened 2023-07-08
Goal Code for a flow that simulates a connectionless present proof using a connection

swcurran:
Per the addition to the goal codes list -- useful when you want to do a connection-less present proof using a QR code, but the OOB message is too large for the QR code, and URL shortening is not an option.  With this, establish the connection, do the present proof protocol, and delete the connection.

@jleach @cvarjao 

BTW -- not sure where is the best place to create a repository of goal codes to use.  We should discuss that as part of the review of this PR.


TelegramSam (approved):
Discussed WG 20240313. approved for merge

PR #794, opened 2023-08-29
DID Rotate Protocol

TelegramSam:


features/0794-did-rotate/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+This message is still sent to the prior DID to acknowledge the receipt of the rotation. Following messages will be sent to the new DID.
+
+`to_did`: The new DID to be used to identify the **rotating_party**
+
+```json
+{
+    "@id": "123456780",
+    "@type": "https://didcomm.org/did-rotate/1.0/ack",
+    "to_did": "did:example:newdid",
```
genaris, line 71: Is it needed to repeat the `to_did` field to confirm the reception of it, or should this message follow the convention including status and thread id? For example:

```
"status": "OK",
  "~thread": {
    "thid": "rotate-did-message-id",
  }
```

TelegramSam: Good topic for discussion in the Aries WG call. I included it as a bit of redundant info, but maybe it shouldn't be.

TelegramSam:
note about reasonable time for observing party

TelegramSam:
notes about stored messages and state, and what should be updated.

TelegramSam:
discussed WG 20230830. ok to merge after discussed changes.

genaris:
I'm sorry if you've discussed this today and I wasn't paying attention but I noticed from a [Discord message](https://discord.com/channels/905194001349627914/905206466410057728/1146310708456345610) from @swcurran something that seems to me a very good idea: to be able to set `to_did` field to a null value in order to terminate the connection (pretty much the same as supported natively DIDComm V2).

I think this may add a little bit of complexity, but certainly would give to this protocol more value and solve an interesting use-case in an elegant way.

swcurran:
Thanks for raising that @genaris — I forgot to raise it. I think it is an excellent idea. The Discord use case was for ending a mediator connection, but there are many other use cases, I’m sure.

features/0794-did-rotate/README.md
TelegramSam: fixed in next commit

features/0794-did-rotate/README.md
```diff
@@ -19,6 +19,10 @@ This protocol is only applicable to DIDComm v1 - in DIDComm v2 use the more effi
 This mechanism allows a party in a relationship to change the DID in use for the relationship. This may be used to switch DID methods,
```
swcurran, line 19: Tweak — “their DID in use” instead of “the DID in use"

features/0794-did-rotate/README.md
```diff
@@ -19,6 +19,10 @@ This protocol is only applicable to DIDComm v1 - in DIDComm v2 use the more effi
...
 but also to switch to a new DID within the same DID method. Inspired by (but different from) the DID rotation feature of the DIDComm Messaging (DIDComm v2) spec.
```
swcurran, line 20: “new DID within the same DID method” to “new DID within the same DID method, such as to update the mediator relationship in the DIDDoc.”

I think it would be helpful to include that example.

features/0794-did-rotate/README.md
```diff
@@ -19,6 +19,10 @@ This protocol is only applicable to DIDComm v1 - in DIDComm v2 use the more effi
...
 
+## Implications for Software Implementations
+
+This protocol enables the identifiers used in a relationship to change. Implementations will need to consider how data related to the relationship is managed. If the relationship DIDs are used as identifiers, those identifiers may need to be updated during the rotation to maintain data integrity.
```
swcurran, line 24: I think this comment is pretty subtle, and useful only for those “deep in the know”.   I’m not sure this is any better, so take this with a grain of salt...

> Implementations will need to consider how data (public keys, DIDs and the ID for the relationship) related to the relationship is managed. If the relationship DIDs are used as identifiers, those identifiers may need to be updated during the rotation to maintain data integrity. For example, both parties might have to retain and be able to use as identifiers for the relationship the existing DID and the rotated to DID, and their related keys for a period of time until the rotation is complete.

features/0794-did-rotate/README.md
```diff
@@ -62,13 +66,17 @@ Message Type URI: https://didcomm.org/did-rotate/1.0/ack
 
 This message is still sent to the prior DID to acknowledge the receipt of the rotation. Following messages will be sent to the new DID.
 
-`to_did`: The new DID to be used to identify the **rotating_party**
+In order to correctly process out of order messages, the The **observing_party** may choose to receive messages from the old did for a reasonable period. This allows messages sent before rotation but received after rotation in the case of out of order message delivery.
```
swcurran, line 69: Consistency — “old did” —> “old DID"

features/0794-did-rotate/README.md
```diff
@@ -62,13 +66,17 @@ Message Type URI: https://didcomm.org/did-rotate/1.0/ack
 
 This message is still sent to the prior DID to acknowledge the receipt of the rotation. Following messages will be sent to the new DID.
 
-`to_did`: The new DID to be used to identify the **rotating_party**
```
swcurran: For this ACK and Problem Report section it is typical to talk about the two messages being “adopted” by the ACK and Report Problem messages, to include a link and to ensure that the usage is consistent.  I didn’t check on the consistency of use, but I’m guessing it was done :-).

features/0794-did-rotate/README.md
```diff
@@ -0,0 +1,137 @@
...
+    "@id": "123456780",
+    "@type": "https://didcomm.org/did-rotate/1.0/rotate",
+    "to_did": "did:example:newdid",
+
+}
+```
+
+The **rotating_party** is expected to receive messages on both the existing and new DIDs and their associated keys for a reasonable period that MUST extend at least until the following `ack` message has been received.
+
+This message MUST be sent using authcrypt or as a signed message in order to establish the provenance of the new DID. Proper provenance prevents injection attacks that seek to take over a relationship. Any rotate message received without being authcrypted or signed MUST be discarded and not processed.
```
swcurran, line 59: I would add something like so that readers unfamiliar with the DIDComm message handling will not go down a rabbit hole of trying to figure out if a message is AuthCrypt or not: “In Aries implementations, messages sent in the context of an existing relationship are by default sent using authcrypt.”

swcurran:
Looks good to me.  A few suggestions, but very minor — I won’t be upset if they aren’t made. Let me know when you have looked at them to your satisfaction, and I’ll approve.

Nice work!

features/0794-did-rotate/README.md
```diff
@@ -0,0 +1,137 @@
...
+```json
+{
+    "@id": "123456780",
+    "@type": "https://didcomm.org/did-rotate/1.0/hangup"
+}
+```
+
+## Prior art
+
+This protocol is inspired buy the rotation feature of DIDComm Messaging (DIDComm v2). The implementation differs in important ways.
```
dbluhm, line 125:
```suggestion
This protocol is inspired by the rotation feature of DIDComm Messaging (DIDComm v2). The implementation differs in important ways.
```

features/0794-did-rotate/README.md
TelegramSam: I've gone with :
>This mechanism allows a party in a relationship to change the DID they use to identify themselves in that relationship.

features/0794-did-rotate/README.md
TelegramSam: added

dbluhm:
In a quick attempt to feel out any unexpected gotchas, I'm going to take a quick pass at an implementation of the proposed protocol. I'll report back with my findings.

dbluhm (approved):

features/0794-did-rotate/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Summary
+
+This protocol signals the change of DID in use between parties.
+
+This protocol is only applicable to DIDComm v1 - in DIDComm v2 use the more efficient DID Rotation header.
+
+## Motivation
+
+This mechanism allows a party in a relationship to change the DID they use to identify themselves in that relationship. This may be used to switch DID methods,
+but also to switch to a new DID within the same DID method. For non-updatable DID methods, this allows updating DID Doc attributes such as service endpoints. Inspired by (but different from) the DID rotation feature of the DIDComm Messaging (DIDComm v2) spec.
```
genaris, line 20: I think updating DID Doc attributes for non-resolvable DIDs would need to also attach the DID Document, like we do in DID Exchange protocol.

Do you think we can add an attachment for those cases to allow a broader range of use cases of this protocol, or limit it to those DIDs that are publicly resolvable or embed the DID Document on its identifier?

swcurran: Given the timing of this and the availability/desirability of `did:peer:4`, that we should not add this. We don’t want implementations using either unqualified DIDs or `did:peer:1`, which are the only options AFAIK that would need this.  We could add a note about this, but I don’t think we should add the attachment option.

TelegramSam: I'm not sure I understand the question. Let's discuss in the Aries WG.

features/0794-did-rotate/README.md
swcurran: Super minor, and not necessary, but I realize there are other motivators for adding/using this protocol, and perhaps a list is needed:

- change mediators
- change the key in an non-updatable DID method.
- hanging up
- Switch from a deprecated DID Method, such as unqualified DIDs or `did:peer` numalogos 1 and 2/3 to `did:peer:4`

swcurran (approved):

features/0794-did-rotate/README.md
TelegramSam: per discussion in WG 20230920: Would prefer not to add support in favor of did:peer:4, but willing to add if this presents a significant hurdle for the community using did:peer:1.

PR #795, opened 2023-09-13
Add DID rotation signature to 23

### LINKED ISSUE #717: RFC 0023: Can we accept inline peer did without a signature?

TelegramSam:
Fixes #717

### DIFF
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -350,7 +350,7 @@ The invitation's `recipientKeys` should be dedicated to envelopes authenticated 
 * The `~thread` decorator MUST be included. It contains a `thid` reference to the `@id` of the request message.
 * The `did` attribute MUST be included. It denotes the DID in use by the responder. Note that this MAY NOT be the same DID used in the invitation.
 * The `did_doc~attach` optional, contains the DID Doc associated with the `did`, if required.
-  * If the `did` is resolvable (either an inline `peer:did` or a publicly resolvable DID), the `did_doc~attach` attribute should not be included.
+  * If the `did` is resolvable (either an inline `did:peer` or a publicly resolvable DID), the `did_doc~attach` attribute should not be included.
   * If the DID is a `did:peer` identifier, the DIDDoc must be as outlined in [RFC 0627 Static Peer DIDs](../0627-static-peer-dids/README.md).
 * The `did_rotate~attach` attribute is optional, but SHOULD be included if the `did` attribute is resolvable and the `did_doc~attach` is not included. The value is the Base64url encoded DID, and signed with the key used in the invitation.
 
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -16,6 +16,11 @@ This RFC describes the protocol to exchange DIDs between agents when establishin
 
 Aries agent developers want to create agents that are able to establish relationships with each other and exchange secure information using keys and endpoints in DID Documents. For this to happen there must be a clear protocol to exchange DIDs.
 
+## Version Change Log
+
+### 1.1 -  Signed Rotations without DID Documents
+Added the optional `did_rotate~attach` attachment for provenance of rotation without an attached DID Document. 
+
 ## Tutorial
 
 We will explain how DIDs are exchanged, with the roles, states, and messages required.
--- a/features/0023-did-exchange/README.md
+++ b/features/0023-did-exchange/README.md
@@ -325,6 +325,19 @@ The exchange response message is used to complete the exchange. This message is 
             "signature": "3dZWsuru7QAVFUCtTd0s7uc1peYEijx4eyt5... (bytes omitted)"
             }
       }
+   },
+   "did_rotate~attach": {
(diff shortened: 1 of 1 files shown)


features/0023-did-exchange/README.md
```diff
@@ -325,6 +325,19 @@ The exchange response message is used to complete the exchange. This message is
             "signature": "3dZWsuru7QAVFUCtTd0s7uc1peYEijx4eyt5... (bytes omitted)"
             }
       }
+   },
+   "did_rotate~attach": {
```
swcurran, line 329: Two questions to consider — more to know why you did it this way then to propose changes.

- Should this be a change to the protocol version to 1.1?
   - It is a new item, so I wanted to check to see if you think there should be a version bump.
- Why the “_” and “-“ in the same item?
  - I see it in the did_doc-attach, so assume that is the reason, but thought I would ask :-).

TelegramSam: second item: it's a tilde, not a dash.
first item - maybe/probably? A topic for tomorrow.

features/0023-did-exchange/README.md
```diff
@@ -339,6 +352,7 @@ The invitation's `recipientKeys` should be dedicated to envelopes authenticated
 * The `did_doc~attach` optional, contains the DID Doc associated with the `did`, if required.
   * If the `did` is resolvable (either an inline `peer:did` or a publicly resolvable DID), the `did_doc~attach` attribute should not be included.
   * If the DID is a `did:peer` identifier, the DIDDoc must be as outlined in [RFC 0627 Static Peer DIDs](../0627-static-peer-dids/README.md).
+* The `did_rotate~attach` attribute is optional, but SHOULD be included if the `did` attribute is resolvable and the `did_doc~attach` is not included. The value is the Base64url encoded DID, and signed with the key used in the invitation.
```
TimoGlastra, line 356: I think we should include that if a did is used in the response message that is different from the invitation, the `did_rotate~attach` MUST be present.

TelegramSam: Great wording, though this implies a stronger MUST requirement than we usually allow for a minor version change.

features/0023-did-exchange/README.md
swcurran: Doh!  Sorry about that — it flashed through my head that the attach was a decorator, but didn’t stick, and not when I was looking at the item name :-).

That makes it easier for this to be a no version change.  The guidance is to put a decorator on it with this content vs. the must.  Of course, any implementation can do what they think is appropriate — either including the `~attach` decorator or not, ignoring or processing the decorator on receipt, and rejecting or not the connection if NOT included.

features/0023-did-exchange/README.md
swcurran: The MUST is in the eye of the recipient. If they choose to reject a connection because its not included, then so be it.

TelegramSam: need more language - feels wrong (warren) to not have a MUST included.

genaris (commented):
What about the `request` message? Is it some clarification needed in regards of the signature for the DID Doc attachment?

From #717:
> Signatures were never needed in the Requests, and were added by mistake by... me. Does this address your confusion? (@TelegramSam  @ https://github.com/hyperledger/aries-rfcs/issues/717#issuecomment-1069768955)

> - the request message not to be signed. For did:peer:1 dids it's still required to not break interop, but for all resolvable dids, the signed attachment SHOULD NOT be present (@TimoGlastra  @ https://github.com/hyperledger/aries-rfcs/issues/717#issuecomment-1684297399)

TelegramSam:
> What about the `request` message? Is it some clarification needed in regards of the signature for the DID Doc attachment?

The signature in the response is used to prove ownership over the key used in the invitation. In the request message, the only key they could prove ownership of is the one for the DID presented in the message, and that is already proven by sending an authcrypted message, per standard practice.

Am I correct here?

TelegramSam:
Discussed WG 20231004. Needs to bump the version to 1.1

swcurran (approved):
Discussed and agreed to be merged at Aries Working Group Call 2023.12.06.

PR #801, opened 2023-11-21
Update the 'unresolved questions' section of the 0317-please-ack RFC

alexsdsr:
This PR adds one more 'unresolved question' to the 0317-please-ack RFC.


swcurran:
Sorry for the delay. We did talk about this on the Aries Working Group call and I’ll get back to you with feedback from the group.

swcurran:
The Aries Community is removing please_ack from AIP 2.0 and retiring the please_ack protocol in https://github.com/hyperledger/aries-rfcs/pull/814 that should be merged Real Soon Now. In that PR, links are provided to the PRs done here (but not merged) to provide extra details as to why the protocol is being retired — because a “general purpose” please_ack feature is not viable and if wanted, MUST be part of each protocol.

Thanks again for the work, @alexsdsr. Even though it didn’t get merged we learned a lot from this and moved DIDComm forward. The biggest learning is to make sure that we don’t “throw in” cross-cutting or optional features that alter the protocols state machines. For example — as recently come up — throwing in an optional, extra “problem-report” in response to a flawed final message of a protocol breaks the state machine. Either the extra message MUST be made every time, or not ever made — not sometimes...

PR #804, opened 2023-12-05
Proposes the DIDComm RPC Protocol for App Attestation and more

swcurran:
The DIDComm RPC (DRPC) protocol allows a remote request-response interaction, a Remote Procedure Call (RPC), to be carried out through aDIDComm channel. The requesting Aries agent sends a DIDComm message to request the remote service be invoked, and gets back the response in another DIDComm message. The protocol enables any request to be conveyed, and the subsequent response to be returned, with flexibility in the formats (JSON, Base64 encoded) sent and
received. Out of scope of this protocol is how the requesting agent discovers the services available from the responding agent, and how the two agents know about the semantics of the requests and responses. By using DIDComm between the requesting and responding agents, the security and privacy benefits of DIDComm are accomplished, and the generic payloads of the messages allows for flexibility in how and where the protocol can be used.

The initial proposed use of this protocol is for App Attestation.

### DIFF
--- a/features/0804-didcomm-rpc/README.md
+++ b/features/0804-didcomm-rpc/README.md
@@ -1,0 +1,381 @@
+# 0335: DIDComm Remote Procedure Call DRPC
+
+- Authors: [Stephen Curran](swcurran@cloudcompass.ca) (BC Gov), [Clecio Varjao](clecio.varjao@gov.bc.ca) (BC Gov)
+- Status: [PROPOSED](/README.md#proposed)
+- Since: 2023-11-29
+- Status Note: An evolution of the HTTP over DIDComm protocol to enable an Agent to request an arbitrary service from a connected Agent and get a response.
+- Supersedes:
+- Start Date: 2023-11-29
+- Tags: [feature](/tags.md#feature), [protocol](/tags.md#protocol)
+
+## Summary
+
+The DIDComm RPC (DRPC) protocol allows a remote request-response interaction, a
+Remote Procedure Call (RPC), to be carried out through a
+DIDComm channel. The requesting Aries agent sends a DIDComm message to request the remote
+service be invoked, and gets back the response in another DIDComm message. The
+protocol enables any request to be conveyed, and the subsequent response to be
+returned, with flexibility in the formats (JSON, Base64 encoded) sent and
+received. Out of scope of this protocol is how the requesting agent
+discovers the services available from the responding agent, and how the two
+agents know about the semantics of the requests and responses. By using DIDComm
+between the requesting and responding agents, the security and privacy benefits
+of DIDComm are accomplished, and the generic payloads of the messages allows for
+flexibility in how and where the protocol can be used.
+
+## Motivation
+
+There are several use cases that are driving the need for this protocol:
+
+### App Attestation
+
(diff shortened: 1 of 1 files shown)


features/0804-didcomm-rpc/README.md
```diff
@@ -0,0 +1,381 @@
...
+  "body.base64": "<Base64 Encoded JSON>"
+}
+```
+
+The items in the message are as follows:
+
+- `@type` -- required, must be as above
+- `@id` -- required, must be as defined in [RFC 0005 DIDComm]
+- `request_type` -- **required**, an identifier indicating the type of the request. The value could be a URI, such as an HTTP URL, that the `sender` would like `responder` to resolve, or the value could be just a string understood by the `responder`.
+- `headers` -- optional, a JSON array typically used for conveying HTTP headers to be used in the processing of an HTTP URL `request_type`. As with all parts of this protocol, the meaning is up to the participants of an instance of the protocol.
```
TimoGlastra, line 183: I think RPC has no need to be related to HTTP style headers. Besides content type, do you have specific needs for this? We can solve that part with attachments, and we could then just make this part of the body. So if you need headers, we make it part of the body?

TimoGlastra (commented):
This might be a good fit for the didcomm.org website?

tbloomfi:
This comment is specific to the use case of app attestation and not the actual protocol. There are some emerging OpenID (driven by EUDI and the EU ARF) standards in this space that might be useful to review. The standard includes both the issuance of attestation credentials and a device binding (authentication) for high assurance use cases:

- Italian EUDI wallet design: https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/wallet-instance-attestation.html
- Draft IETF standard as part of OAUTH 2.0: https://www.ietf.org/archive/id/draft-looker-oauth-attestation-based-client-auth-00.html

swcurran:
Suggestions from the discussion at the Aries Working Group call 20231216:


- A number of people thought that we should do an App Attestation protocol vs. using the generic DRPC. Not against DRPC, but that app attestatation is important enough to warrent a protocol.
   - Related to that is looking at the App Attestation HTTP interface that OpenID has defined in @tbloomfi ’s comment above.
- Several people thought that we should look at [gRPC] and [JSON-RPC] to see if we can apply anything from there.
- Some (such as @TimoGlastra in his comments above) liked the idea of using Attachments vs. the current header-has-MIME-type and `body.base64`, since attachments give the implementers flexibilty in formatting the data.
- Also the request above to drop `header` altogether and having any user of the protocol put whatever extra info (headers, etc.) into the body (or attachment if we go that way).

[gRPC]: https://grpc.io/
[JSON-RPC]: https://www.jsonrpc.org/specification

We’ll consider the suggestions to this protocol and provide an update with discussion of the changes.

As to the App Attestation protocol — we’ll look at that, but would like more community feedback on the current implementation (based on `basic-message` — and even less suitable protocol) and experience in using it.

Thanks for the feedback!

TelegramSam:
WG 20240124 implementation under way in ACA-Py.

TelegramSam (approved):
Discussed WG 20240313. merge approved.

PR #811, opened 2024-01-31
Update RFC statuses in accordance with 01/31/24 WG meeting

dbluhm:


TelegramSam (approved):
Checked against my notes, and it's all there. Nice work @dbluhm!

PR #818, opened 2024-03-13
Removes references to ~please_ack in other protocols

swcurran:
`~please_ack` was [recently retired](https://github.com/hyperledger/aries-rfcs/tree/main/features/0317-please-ack). This PR removes the remaining references to `~please_ack` in other protocols. Most relevant to the community is the removal from the issue-credential and present-proof protocols -- which need to be update in AIP 2.0 to include recent clarifications (including this one).

This is not controversial as we have already decided to retire `~please_ack` -- this is just post retirement cleanup.

I think this PR could just be merged with approvals from a @TelegramSam, @TimoGlastra, @genaris but could also wait for an Aries Working Group meeting.


TelegramSam (commented):
Discussed WG 20240320. requested changelog entries on affected RFCs

genaris (approved):
Just to comment that from Credo perspective I think there is no problem, since in the protocols where we have references to it we are not explicitly using it:

- In Issue Credential and Present Proof V1/V2, we set it at issuance/presentation step. However when accepting a credential or a presentation (something triggered usually by the controller), we send the ack regardless of the original `please_ack` decorator
- Same story Revocation Notification V1/V2: we can set it when sending it, but when receiving we are not relying on it to send an ack or not (it is up to agent controller to do so)

So I think we can keep setting it for legacy reasons, but avoid using it when implementing new protocols (e.g. Introduce).

TimoGlastra (approved):

swcurran:
Updated the RFCs to have change log entries for the change. Made the change logs consistently formatted across the updated RFCs.

PR #819, opened 2024-03-13
Update the Aries RFC Index and remove the indexgenerate GHA since it is not helpful

swcurran:
Updates the Aries RFC index.md file since some RFC statuses changed.

Also removes the indexgenerate GHA since does not work and is not really worth fixing.  I'm tired of getting failed run emails for the task :-) .


TelegramSam (approved):
Discussed WG 20240320. approved for merge.

PR #821, opened 2024-03-20
Update DID Exchange examples to use version 1.1 from 1.0

swcurran:
Makes consist the version numbering in the example. The protocol is at version 1.1, so the examples should reflect that.


TelegramSam (approved):
Discussed WG 20240320. approved for merge.

PR #823, opened 2024-03-27
First draft of the Aries RFCs website -- generation and publishing

swcurran:
A PR that might trigger the publication of a GH Pages site - depending on the success of the included GHAction. I tested it as much as I could.

In theory, this should create a gh-pages branch from which we can publish the site. The gh-pages static content will be based on the docs generated by the GHAction. From there we can publish it.

As noted on the Aries Working Group call (20240327) this is the start of the work on this, there are some todos that will be added as issues to be worked on.

This should be a "do no harm" PR -- it shouldn't affect anything other than adding a gh-pages branch in the repo.

### DIFF
--- a/.github/workflows/publish-site.yml
+++ b/.github/workflows/publish-site.yml
@@ -1,0 +1,39 @@
+name: publish-docs 
+
+on:
+  push:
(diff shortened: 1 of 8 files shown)


TelegramSam (approved):
great initial work and foundation for rendered RFC docs.

PR #828, opened 2024-04-22
Credential offer proposal

PatStLouis:
I'd like to suggest a re-alignment with the vc-api approach for the credential offer attachment. I'm not convinced the suggested change provides sufficient improvements and the rational aligns with the direction of the vc-api. The Credential Offer Exceptions section is in line with the recent discussions on the vc-api calls.

This change seems to get rid of a dedicated options field and put all options at the root layer of the request.

The JSON-LD attachment dates from 2021 and the vc-api specification has changed since then.

I have a few questions:

- Why include the data model version in the options? Could it not be derived from the context? In which scenario would an array be beneficial?
- Why include a `binding_required` bool? Could it not be assumed as required if there is a `binding_method` present? Is there a scenario where the binding isn't required but a method is provided?

@TimoGlastra

### DIFF
--- a/features/0809-w3c-data-integrity-credential-attachment/README.md
+++ b/features/0809-w3c-data-integrity-credential-attachment/README.md
@@ -29,7 +29,6 @@ Format identifier: `didcomm/w3c-di-vc-offer@v0.1`
 ```json
 {
   "options": {
-    "data_model_version": "v2",
     "binding_method": {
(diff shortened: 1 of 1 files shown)


TelegramSam:
Discussed 2024. Hard to merge given the changes, but could be useful in a note on the RFC.

PR #834, opened 2024-05-15
Update 496 OOB transition with dates and title

TelegramSam:
Updated to bring it back into action.


JamesKEbert (approved):
LGTM
Note--in the Between Step Triggers section, it notes that the RFC should move to the status `Adopted` upon reaching Step 2.

swcurran (approved):

PR #840, opened 2024-06-04
fix: RFC JSON formatting

JamesKEbert:
A couple of RFCs' JSON examples don't display properly in deployed site, so I ran a formatter over it to resolve the issues (the ```json= is the primary problem). The issue can be observed here: https://hyperledger.github.io/aries-rfcs/latest/features/0685-pickup-v2/#reference
Also updates RFC 0496 to be in ADOPTED category in mkdocs/index (#834)


swcurran (approved):

swcurran:
I manually verified that the DCO is correct, despite the hung check.  @ryjones — should we plan on going ahead with merges while the DCO check is stuck based on manual checking?  What is the impact if we mistakenly think the DCO is correct and it is not?

ryjones:
@swcurran [you should add this file](https://github.com/hyperledger/besu/pull/7179) in all of your repos

swcurran:
@JamesKEbert could you push some other commit to trigger the running of the DCO check? It should go then...

JamesKEbert:
Done :+1:

swcurran (approved):

PR #851, opened 2024-08-25
Adding unrevoke messages to revocation-notification protocol

thiagoromanos:
Adding a `unrevoke` message to the revocation-notification protocol (v1 and v2), so the holder would be aware when one of his revoked credentials have been unrevoked by the issuer.


swcurran:
@TelegramSam — please add this as a topic at the next Aries Working Group meeting.

@thiagoromanos — what about adding an optional flag “unrevoke” to the revoke message and set it to `true` when passed for an unrevocation?  Not sure it is any better — it a minor change either way with an addition to the protocol.

I suppose that with that we could get subtle and say that if a “revoke” message is sent with an `”unrevoke”: false` then the issuer is letting the holder know that the credential _could_ be unrevoked in the future — and that it should not necessarily delete the credential.  Too subtle?

Let’s discuss.

swcurran (approved):
Reviewed at the 2024.09.25 Aries Working Group meeting and approved.

